PR #31, opened 2023-07-11
Add rancher cluster wrapper

alexander-demicev:
Related issue https://github.com/rancher-sandbox/rancher-turtles/issues/30

This is the first PR for introducing a wrapper around the rancher cluster object to avoid fighting with dependencies. We can use `mapstucture` library that will convert from wrapper to unstructured and back based on JSON tags. The wrapper can be later extended with the status field for example.

I will introduce the actual usage of it in a follow-up PR.

Fixes #

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/internal/ranchercluster/ranchercluster.go
+++ b/internal/ranchercluster/ranchercluster.go
@@ -1,0 +1,122 @@
+package ranchercluster
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"time"
+
+	"github.com/mitchellh/mapstructure"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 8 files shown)


internal/ranchercluster/ranchercluster.go
```diff
@@ -0,0 +1,122 @@
...
+	stringToTimeHook = func(f reflect.Type, t reflect.Type, data interface{}) (interface{}, error) {
+		if f.Kind() == reflect.String && t == reflect.TypeOf(metav1.Time{}) {
+			time, err := time.Parse(time.RFC3339, data.(string))
+			return metav1.Time{Time: time}, err
+		}
+		return data, nil
+	}
+)
+
+type RancherCluster struct {
```
richardcase, line 29: Be good to have some godoc comments on the exported types/functions. wdyt?

richardcase (approved):
LGTM - just one comments about godoc

Issue #46, opened 2023-07-21
Implement initial e2e tests


richardcase:
### What would you like to be added (User Story)?

As a rancher turtles dev
I want to have e2e tests
So that i have confidence that there are no regressions

### Detailed Description

We need this as a base for specifically testing various scenarios such as #4

### Label(s) to be applied

/kind feature

Danil-Grigorev:
Initial implementation was done, closing this.

PR #49, opened 2023-07-24
Linting and installing charts on PR

Danil-Grigorev:
#46

This github workflow is linting and installing our chart in a test cluster, validating the correctness of the provided templates.

Uses https://github.com/marketplace/actions/helm-chart-testing

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


.github/workflows/test_chart.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Create kind cluster
+        uses: helm/kind-action@v1.7.0
+        with:
+          cluster_name: kind
+
+      - name: Add local docker image
+        run: kind load docker-image ${{ env.IMAGE }}:${{ env.TAG }}
+
+      - name: Run chart-testing (install)
+        run: helm install test charts/rancher-turtles/ --post-renderer ./config/kustomize.sh --wait
```
alexander-demicev, line 46: what if include manifests that are outside customized like here https://github.com/kubernetes-sigs/cluster-api-operator/tree/main/hack/charts/cluster-api-operator/templates? How the post-renderer script will behave?

Danil-Grigorev: It will work in tandem

Danil-Grigorev:
> Thanks for working on this! I like the idea of testing helm chart but what is the reason for moving chart out of `hack/chart`? We don't provide a functional chart in the repo, the chart can be compiled with `make release-chart` and consumed in the workflow.

The desired packaging tool for the repo is helm, it seems weird to have first class feature to live in a hack folder.

.github/workflows/test_chart.yaml
alexander-demicev: so validation won't fail if we add manifests that are not part of kustomize built ones?

.github/workflows/test_chart.yaml
```diff
@@ -0,0 +1,46 @@
...
+        with:
+          version: v3.12.1
+
+      - name: Set up chart-testing
+        uses: helm/chart-testing-action@v2.4.0
+
+      - name: setupGo
+        uses: actions/setup-go@v4
+        with:
+          go-version: '=1.20.4'
```
furkatgofurov7, line 29: let's use `1.20.6` that is the version we are trying to [use](https://github.com/rancher-sandbox/rancher-turtles/pull/51/files#diff-d9a956120ac84289d2650137f64bd8f0893331de05e44cc41899dd984c9e0d48R22) in GH workflows

.github/workflows/test_chart.yaml
furkatgofurov7: why not to place `config/kustomize.sh` under `config/chart` which is specific to chart folder?

Danil-Grigorev: Discussed it with Alex, will work upon existing implementation, where helm templating comes second after kustomize. This implementation does it in the opposite order - first templating helm resources, then applies kustomize from config/default on templated resources. Provides a benefit of doing something like setting default namespace on all resources, including templated ones, but current implementation of release process will have to be updated to reflect that. Otherwise we may end up with two different versions of helm chart - packaged and release, and installed locally from folder.

alexander-demicev (approved):

mjura (approved):
LGTM

furkatgofurov7 (approved):

PR #50, opened 2023-07-25
✨ Support deleting the cluster from Rancher

### LINKED ISSUE #7: Support deleting the cluster from Rancher
**Story**

As a Rancher Operator,
I want to delete a CAPI cluster from Rancher,
So that it is no longer listed in Rancher

**Description**

We need to allow deleting a CAPI imported cluster from Rancher.  The underlying CAPI cluster should **not be deleted**.

furkatgofurov7:
kind/feature

This PR adds a support for deleting the CAPI cluster from Rancher, so that it is no longer listed in Rancher by introducing a 
`clusterImportedAnnotation`.  The cluster creation and deletion then as following:

- renames existing `importAnnotationName` to `importAnnotation` to align with new annotation
- introduces `util` folder where we have `annotations` and `predicates` packages 
- `reconcileDelete` made to receive two inputs: CAPI and Rancher cluster, and  if the Rancher Cluster has `ClusterImportedAnnotation` annotation, then we annotate the CAPI cluster to say don't auto-import again.
- introduces unit tests

Fixes #7 

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/internal/controllers/import_controller.go
+++ b/internal/controllers/import_controller.go
@@ -33,10 +33,12 @@ import (
 	"sigs.k8s.io/cluster-api/util/predicates"
 
 	"github.com/rancher-sandbox/rancher-turtles/internal/rancher"
+	turtlesannotations "github.com/rancher-sandbox/rancher-turtles/util/annotations"
+	turtlespredicates "github.com/rancher-sandbox/rancher-turtles/util/predicates"
 )
 
 const (
-	importAnnotationName         = "rancher-auto-import"
+	importAnnotation             = "rancher-auto-import"
 	defaultRequeueDuration       = 1 * time.Minute
 	clusterRegistrationTokenName = "default-token"
 )
@@ -63,7 +65,8 @@ func (r *CAPIImportReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Ma
 
 	// TODO: we want the control plane initialized but removing for the time being
 	// capiPredicates := predicates.All(log, predicates.ClusterControlPlaneInitialized(log), predicates.ResourceHasFilterLabel(log, r.WatchFilterValue))
-	capiPredicates := predicates.All(log, predicates.ResourceHasFilterLabel(log, r.WatchFilterValue))
+	capiPredicates := predicates.All(log, predicates.ResourceHasFilterLabel(log, r.WatchFilterValue),
+		turtlespredicates.ClusterWithoutImportedAnnotation(log))
 
 	c, err := ctrl.NewControllerManagedBy(mgr).
 		For(&clusterv1.Cluster{}).
@@ -147,8 +150,10 @@ func (r *CAPIImportReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		}
 	}
 
-	if !capiCluster.ObjectMeta.DeletionTimestamp.IsZero() {
-		return r.reconcileDelete(ctx, capiCluster)
+	if rancherCluster != nil {
+		if !rancherCluster.ObjectMeta.DeletionTimestamp.IsZero() {
+			return r.reconcileDelete(ctx, capiCluster, rancherCluster)
+		}
 	}
 
 	return r.reconcileNormal(ctx, capiCluster, rancherClusterHandler, rancherCluster)
@@ -175,6 +180,14 @@ func (r *CAPIImportReconciler) reconcileNormal(ctx context.Context, capiCluster 
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      rancherClusterNameFromCAPICluster(capiCluster.Name),
 				Namespace: capiCluster.Namespace,
+				OwnerReferences: []metav1.OwnerReference{
+					{
+						APIVersion: clusterv1.GroupVersion.String(),
+						Kind:       "Cluster",
+						Name:       capiCluster.Name,
+						UID:        capiCluster.UID,
+					},
+				},
 			},
 		}); err != nil {
 			return ctrl.Result{}, fmt.Errorf("error creating rancher cluster: %w", err)
@@ -257,11 +270,26 @@ func (r *CAPIImportReconciler) shouldAutoImport(ctx context.Context, capiCluster
 	return autoImport, nil
 }
 
-func (r *CAPIImportReconciler) reconcileDelete(ctx context.Context, _ *clusterv1.Cluster) (ctrl.Result, error) {
+func (r *CAPIImportReconciler) reconcileDelete(ctx context.Context, capiCluster *clusterv1.Cluster,
+	rancherCluster *rancher.Cluster,
+) (ctrl.Result, error) {
 	log := log.FromContext(ctx)
 	log.Info("Reconciling CAPI Cluster Delete")
 
-	// TODO: add implementation
+	// If the Rancher Cluster has ClusterImportedAnnotation annotation that said we imported it,
+	// then annotate the CAPI cluster so that we don't auto-import again.
+	if turtlesannotations.HasClusterImportAnnotation(rancherCluster) {
+		log.Info(fmt.Sprintf("rancher cluster %s has %s annotation, so annotating CAPI cluster %s",
+			rancherCluster.Name, turtlesannotations.ClusterImportedAnnotation, capiCluster.Name))
+
+		annotations := capiCluster.GetAnnotations()
+		if annotations == nil {
+			annotations = map[string]string{}
+		}
+
+		annotations[turtlesannotations.ClusterImportedAnnotation] = "true"
+		capiCluster.SetAnnotations(annotations)
+	}
 
 	return ctrl.Result{}, nil
 }
@@ -295,7 +323,7 @@ func (r *CAPIImportReconciler) getClusterRegistrationManifest(ctx context.Contex
 }
 
 func shouldImport(obj metav1.Object) (hasLabel bool, labelValue bool) {
-	labelVal, ok := obj.GetAnnotations()[importAnnotationName]
+	labelVal, ok := obj.GetAnnotations()[importAnnotation]
 	if !ok {
 		return false, false
 	}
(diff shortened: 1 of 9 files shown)


internal/controllers/import_controller_test.go
```diff
@@ -276,3 +277,49 @@ var _ = Describe("reconcile CAPI Cluster", func() {
...
+	})
+
+	It("should delete CAPI cluster with imported annotation", func() {
+		if capiCluster.Annotations == nil {
+			capiCluster.Annotations = map[string]string{}
+		}
+		capiCluster.Annotations = map[string]string{
+			clusterImportedAnnotation: "true",
+		}
+		Expect(capiCluster.GetAnnotations()[clusterImportedAnnotation]).To(Equal("true"))
```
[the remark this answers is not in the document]
alexander-demicev: yes, `GetAnnotations()` can still return nil

internal/controllers/import_controller.go
```diff
@@ -246,13 +247,21 @@ func (r *CAPIImportReconciler) shouldAutoImport(ctx context.Context, capiCluster
...
 
-func (r *CAPIImportReconciler) reconcileDelete(ctx context.Context, _ *clusterv1.Cluster) (ctrl.Result, error) {
+func (r *CAPIImportReconciler) reconcileDelete(ctx context.Context, capiCluster *clusterv1.Cluster) (ctrl.Result, error) {
 	log := log.FromContext(ctx)
 	log.Info("Reconciling CAPI Cluster Delete")
 
-	// TODO: add implementation
+	if hasAnnotation(capiCluster, clusterImportedAnnotation) {
+		if err := r.Client.Delete(ctx, capiCluster); err != nil && !apierrors.IsNotFound(err) {
+			return ctrl.Result{}, fmt.Errorf("error deleting %s cluster with imported annotation", capiCluster.Name)
```
alexander-demicev, line 256: seems like `err` value is ignored in the error message.

internal/controllers/import_controller_test.go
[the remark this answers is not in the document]
Danil-Grigorev: I meant to set the map with a single annotation in it. This is what this test case is testing.

internal/controllers/import_controller.go
```diff
@@ -350,6 +371,14 @@ func createObject(ctx context.Context, c client.Client, obj client.Object, log *
 		return nil
 	}
 
+	annotations := obj.GetAnnotations()
+	if annotations == nil {
+		annotations = map[string]string{}
+	}
+
+	annotations[clusterImportedAnnotation] = "true"
```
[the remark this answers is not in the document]
alexander-demicev: `rancher-auto-import` is managed by users, this way they mark clusters that need to be imported, and `imported` will be managed by the controller indicating processed objects. Question to @richardcase, why did we decide to use annotation for marking cluster to import and not labels? If I'm not mistaken annotations are not intended to be used by humans.

alexander-demicev (changes requested):

internal/controllers/import_controller.go
furkatgofurov7: Thanks, added it to error message

internal/controllers/import_controller_test.go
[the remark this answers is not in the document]
furkatgofurov7: reworked it in last commit, PTAL

alexander-demicev (approved):
thanks!

Danil-Grigorev:
Shouldn't we delete Rancher cluster? CAPI will manage deletion itself.

alexander-demicev:
and if we plan to do something custom on CAPI cluster deletion, we might need a finalizer to ensure that CAPI doesn't delete the cluster before we finish cleaning up.

furkatgofurov7:
> Shouldn't we delete Rancher cluster? CAPI will manage deletion itself.

what do you mean by CAPI will manage deletion itself? we already have a rancher cluster delete in https://github.com/rancher-sandbox/rancher-turtles/blob/main/internal/rancher/cluster.go#L127-L139

alexander-demicev:
@furkatgofurov7 If a user manually deletes CAPI cluster, CAPI controller will handle all deletion but what is rancher-turtles expected to do?

mjura (approved):
LGTM

Danil-Grigorev (changes requested):
This does not seem to be fully clear to me. Upon further inspection of the side-to-side installation type with a manual (user driven) cluster import, I’m still under impression that what we need to do for proper deletion handling is:
- Prevent CAPI cluster deletion on generated RANCHER cluster deletion, as per user managed deployment type. This is no-op by now, our repository already acts accordingly, no changes in this PR needed until we support Rancher managed scenario.
- Propagate removal request to Rancher cluster, which was initially issued by our code, if the CAPI cluster is currently being removed or is not already present. This applies to both management type options.

/hold

Danil-Grigorev:
Based on the discussion in the meeting, the deletion could be handled like this:

1. Set the ownership on the Rancher cluster from CAPI cluster (CAPI owns Rancher) when we create it..
2. Set the annotation on Rancher cluster deletion to the CAPI cluster manifest.
3. If the annotation is present, don’t re-import the CAPI cluster next time. If the annotation is not present, cluster could be imported.

Use cases:
1. User manually deletes CAPI cluster. k/k garbage collection will remove the rancher cluster based on resource ownership.
2. User removes the rancher cluster from UI. Annotation is being set on CAPI cluster, preventing automatic re-import. If the annotation is removed, cluster will be reimported. 
 
Potential things to add:

Alex: set the finalizer on CAPI cluster we create, to prevent automatic deletion. The finalizer will be unset by the operator during processing the deletion logic for CAPI cluster.

furkatgofurov7:
> Based on the discussion in the meeting, the deletion could be handled like this:
> 
> 1. Set the ownership on the Rancher cluster from CAPI cluster (CAPI owns Rancher) when we create it..
> 2. Set the annotation on Rancher cluster deletion to the CAPI cluster manifest.
> 3. If the annotation is present, don’t re-import the CAPI cluster next time. If the annotation is not present, cluster could be imported.
> 
> Use cases:
> 
> 1. User manually deletes CAPI cluster. k/k garbage collection will remove the rancher cluster based on resource ownership.
> 2. User removes the rancher cluster from UI. Annotation is being set on CAPI cluster, preventing automatic re-import. If the annotation is removed, cluster will be reimported. CAPI cluster and underlying infrastructure is still present. (Externally Managed mode)
> 
> Potential things to add:
> 
> Alex: set the finalizer on CAPI cluster we create, to prevent automatic deletion. The finalizer will be unset by the operator during processing the deletion logic for CAPI cluster.

@Danil-Grigorev @alexander-demicev @richardcase  the current state of the PR should follow the above notes we have agreed offline, PTAL, thanks!

Danil-Grigorev (dismissed):
LGTM

internal/controllers/import_controller.go
```diff
@@ -257,11 +269,17 @@ func (r *CAPIImportReconciler) shouldAutoImport(ctx context.Context, capiCluster
...
 	log := log.FromContext(ctx)
 	log.Info("Reconciling CAPI Cluster Delete")
 
-	// TODO: add implementation
+	annotations := capiCluster.GetAnnotations()
+	if annotations == nil {
+		annotations = map[string]string{}
+	}
+
+	annotations[clusterImportedAnnotation] = "true"
```
alexander-demicev, line 281: Should we annotate the rancher cluster instead?

alexander-demicev (approved):

internal/controllers/import_controller.go
furkatgofurov7: based on the discussions, we agreed to annotate capi cluster, no?

>2. Set the annotation on Rancher cluster deletion to the CAPI cluster manifest.

richardcase: Yes we agreed to annotate the CAPI cluster as the Rancher Cluster will have been deleted.

richardcase: `reconcileDelete` will only be called if the CAPI cluster has a non-zero deletion timestamp. So what will happen here is that we will annotate the capi cluster and then it will be deleted.

As we are also watching Rancher Clusters we could change object where we are [checking the deletion timestamp ](https://github.com/rancher-sandbox/rancher-turtles/blob/2ccb252757b2febd270ae1bf3b7383565c45a3a3/internal/controllers/import_controller.go#L154) to be the Rancher Cluster instead and then call **reconcileDelete** passing in the CAPI cluster & Rancher Cluster. If the Rancher Cluster has our annotation that said we imported it, then we can then annotate the CAPI cluster to say don't auto-import again.

internal/controllers/import_controller.go
```diff
@@ -475,3 +493,53 @@ func (r *CAPIImportReconciler) createObject(ctx context.Context, c client.Client
 
 	return nil
 }
+
+// ClusterWithoutImportedAnnotation returns a predicate that returns true only if the provided resource does not contain
+// "clusterImportedAnnotation" annotation. When annotation is present on the resource, controller will skip reconciliation.
+func ClusterWithoutImportedAnnotation(logger logr.Logger) predicate.Funcs {
```
richardcase, line 499: Shall we move this to a separate predicates package?

mjura (approved):
LGTM

internal/controllers/import_controller.go
alexander-demicev: this can be done in a follow up PR

alexander-demicev (dismissed):

alexander-demicev (approved):

internal/controllers/import_controller.go
furkatgofurov7: Done

internal/controllers/import_controller.go
furkatgofurov7: @richardcase thanks for review, I changed `reconcileDelete` logic based on your suggestions, PTAL

PR #58, opened 2023-07-31
✨ Use helm chart releaser and generate notes on each release

### LINKED ISSUE #23: Automate extension build
#### Story

The extension should be built automatically via GitHub workflows.

#### Description

Existing Rancher extensions [implement a series of GitHub Actions](https://github.com/kubewarden/ui/actions/workflows/build-extension.yml) that automatically build the extension package. Based off these workflows, automate building and packaging the CAPI extension.

Danil-Grigorev:
kind/feature

This PR is implementing support for publishing charts with [chart-releaser](https://github.com/helm/chart-releaser) tool. Additionally adds auto release note generation for each release based on CAPI [release/note](https://github.com/kubernetes-sigs/cluster-api/blob/main/hack/tools/release/notes.go) tool.

Here is an example release created by this workflow: https://github.com/Danil-Grigorev/rancher-turtles/releases/tag/rancher-turtles-1.2.20, might be good to have labeling on PRs similar to CAPI operator or CAPI. 

Additionally it removes `rancher-turtles-components.yaml` file created with each release, as there is no need for that in rancher-turtles opposed to CAPI operator.

Fixes #23

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ SHELL = /usr/bin/env bash -o pipefail
 #
 GO_VERSION ?= 1.20.4
 GO_CONTAINER_IMAGE ?= docker.io/library/golang:$(GO_VERSION)
+REPO ?= rancher-sandbox/rancher-turtles
 
 # Use GOPROXY environment variable if set
 GOPROXY := $(shell go env GOPROXY)
@@ -83,6 +84,9 @@ GOLANGCI_LINT_VER := v1.53.3
 GOLANGCI_LINT_BIN := golangci-lint
 GOLANGCI_LINT := $(abspath $(TOOLS_BIN_DIR)/$(GOLANGCI_LINT_BIN))
 
+NOTES_BIN := notes
+NOTES := $(abspath $(TOOLS_BIN_DIR)/$(NOTES_BIN))
+
 # Registry / images
 TAG ?= dev
 ARCH ?= $(shell go env GOARCH)
@@ -95,6 +99,7 @@ MANIFEST_IMG ?= $(CONTROLLER_IMG)-$(ARCH)
 
 # Relase
 RELEASE_TAG ?= $(shell git describe --abbrev=0 2>/dev/null)
+PREVIOUS_TAG ?= $(shell git describe --abbrev=0 --exclude $(RELEASE_TAG) 2>/dev/null)
 HELM_CHART_TAG := $(shell echo $(RELEASE_TAG) | cut -c 2-)
 RELEASE_ALIAS_TAG ?= $(PULL_BASE_REF)
 RELEASE_DIR ?= out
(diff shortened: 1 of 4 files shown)


hack/make-release-notes.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -e
+set -o errexit
+set -o pipefail
+
+TOOLS_FOLDER=${1:-../hack/tools/bin}
+
+mkdir $TOOLS_FOLDER/cluster-api
+cd $TOOLS_FOLDER/cluster-api
```
alexander-demicev, line 10: I think this script has to be adjusted for rancher turles

Makefile
```diff
@@ -356,6 +364,9 @@ $(GOLANGCI_LINT): # Download and install golangci-lint
 		-b $(TOOLS_BIN_DIR) \
 		$(GOLANGCI_LINT_VER)
 
+$(NOTES): # Download and install note generator from cluster-api commit
```
furkatgofurov7, line 367: we need to remove capi notes, perhaps?

furkatgofurov7 (changes requested):

Makefile
Danil-Grigorev: None of the created binaries are getting removed by any of the makefile tasks, why should it be?

hack/make-release-notes.sh
Danil-Grigorev: I'm not sure I understand. This script was created for rancher turtles specifically, what I missing here to be adjusted?

Danil-Grigorev:
One gap I'm seeing now, is that the image field is not configurable by helm in any way. So this helm chart is not tailored to specific architecture as it should probably be based on the docker images built, it just gets the amd64 image due to CI environment. Not sure though if it was addressed in the initial implementation of the make release. But we could populate values.yaml later, and add templating for the created resource, similar to https://github.com/kubernetes-sigs/cluster-api-operator/pull/199. And we could remove kustomize by incremental refactoring I guess.

alexander-demicev (approved):
In the future, we should talk with CAPI folks about converting the release note tool to a github action, with some changes it can be reused by CAPI providers.

hack/make-release-notes.sh
alexander-demicev: Ok, I see there is no way to import the release note tool here

Makefile
alexander-demicev: @furkatgofurov7 There is no nice way to import CAPI release tool here, all methods are private. The only option is to build it from CAPI repo. We can talk with CAPI folks about converting it to a github action.

furkatgofurov7 (approved):

PR #62, opened 2023-08-02
Use label instead of annotation for importing clusters

alexander-demicev:
Use label instead of annotation for importing clusters, I also changed label key to `"cluster-api.cattle.io/rancher-auto-import"` so it's more similar to other rancher labels/annotations.

Fixes https://github.com/rancher-sandbox/rancher-turtles/issues/59

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/internal/controllers/import_controller.go
+++ b/internal/controllers/import_controller.go
@@ -38,7 +38,7 @@ import (
 )
 
 const (
-	importAnnotation             = "rancher-auto-import"
+	importLabelName              = "cluster-api.cattle.io/rancher-auto-import"
 	defaultRequeueDuration       = 1 * time.Minute
 	clusterRegistrationTokenName = "default-token"
 )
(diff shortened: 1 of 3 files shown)


internal/controllers/import_controller.go
```diff
@@ -39,6 +39,7 @@ import (
 
 const (
 	importAnnotation             = "rancher-auto-import"
```
Danil-Grigorev, line 41: I think this should be removed to satisfy the fact that the labels should be used instead of annotation

mjura: I am not sure, Labels usually are used by Kubernetes control plane services, annotation are much more dedicated for third party operators.

mjura: I have checked and Rancher is mostly using annotations, which is right approach

internal/controllers/import_controller.go
```diff
@@ -39,6 +39,7 @@ import (
...
+	importLabelName              = "cluster-api.cattle.io/rancher-auto-import"
```
mjura, line 41: Why we don't use annotations ?

mjura (commented):
I am not sure about switching to labels, annotations are dedicated concept for third party operators

internal/controllers/import_controller.go
alexander-demicev: annotations are usually not meant for user interactions, they can even have non human readable format. Labels are designed for users, in addition to that we can do filtering with labelselector if needed, which we can't with annotations.

mjura: ok, after discussion, it seems that it was already discussed with the team

furkatgofurov7 (approved):

Danil-Grigorev (approved):

PR #70, opened 2023-08-11
Introduce new predicates

alexander-demicev:

Add new predicates to filter out clusters that don't have a control plane ready or were not marked for import. I didn't remove all similar logic from reconciler because I've seen in the past predicates failing and unexpected behavior happening because of that. 

Fixes #

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/internal/controllers/import_controller.go
+++ b/internal/controllers/import_controller.go
@@ -23,7 +23,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"strconv"
 	"strings"
 	"time"
 
@@ -52,6 +51,7 @@ import (
 	utilyaml "sigs.k8s.io/cluster-api/util/yaml"
 
 	"github.com/rancher-sandbox/rancher-turtles/internal/rancher"
+	"github.com/rancher-sandbox/rancher-turtles/util"
 	turtlesannotations "github.com/rancher-sandbox/rancher-turtles/util/annotations"
 	turtelesnaming "github.com/rancher-sandbox/rancher-turtles/util/naming"
 	turtlespredicates "github.com/rancher-sandbox/rancher-turtles/util/predicates"
@@ -83,10 +83,12 @@ func (r *CAPIImportReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Ma
 		r.remoteClientGetter = remote.NewClusterClient
 	}
 
-	// TODO: we want the control plane initialized but removing for the time being
-	// capiPredicates := predicates.All(log, predicates.ClusterControlPlaneInitialized(log), predicates.ResourceHasFilterLabel(log, r.WatchFilterValue))
-	capiPredicates := predicates.All(log, predicates.ResourceHasFilterLabel(log, r.WatchFilterValue),
-		turtlespredicates.ClusterWithoutImportedAnnotation(log))
+	capiPredicates := predicates.All(log,
+		predicates.ResourceHasFilterLabel(log, r.WatchFilterValue),
+		turtlespredicates.ClusterWithoutImportedAnnotation(log),
+		turtlespredicates.ClusterWithReadyControlPlane(log),
+		turtlespredicates.ClusterOrNamespaceWithImportLabel(ctx, log, r.Client, importLabelName),
+	)
 
 	c, err := ctrl.NewControllerManagedBy(mgr).
 		For(&clusterv1.Cluster{}).
@@ -216,7 +218,7 @@ func (r *CAPIImportReconciler) reconcileNormal(ctx context.Context, capiCluster 
 	log := log.FromContext(ctx)
 
 	if rancherCluster == nil {
-		shouldImport, err := r.shouldAutoImport(ctx, capiCluster)
+		shouldImport, err := util.ShouldAutoImport(ctx, log, r.Client, capiCluster, importLabelName)
 		if err != nil {
 			return ctrl.Result{}, err
 		}
@@ -290,38 +292,6 @@ func (r *CAPIImportReconciler) reconcileNormal(ctx context.Context, capiCluster 
 	return ctrl.Result{}, nil
 }
 
-func (r *CAPIImportReconciler) shouldAutoImport(ctx context.Context, capiCluster *clusterv1.Cluster) (bool, error) {
-	log := log.FromContext(ctx)
-	log.V(2).Info("should we auto import the capi cluster", "name", capiCluster.Name, "namespace", capiCluster.Namespace)
-
-	// Check CAPI cluster for label first
-	hasLabel, autoImport := shouldImport(capiCluster)
-	if hasLabel && autoImport {
-		log.V(2).Info("Cluster contains import annotation")
-
-		return true, nil
-	}
-
-	if hasLabel && !autoImport {
-		log.V(2).Info("Cluster contains annotation to not import")
-
-		return false, nil
-	}
-
-	// Check namespace wide
-	ns := &corev1.Namespace{}
-	key := client.ObjectKey{Name: capiCluster.Namespace}
-
-	if err := r.Client.Get(ctx, key, ns); err != nil {
-		log.Error(err, "getting namespace")
-		return false, err
-	}
-
-	_, autoImport = shouldImport(ns)
-
-	return autoImport, nil
-}
-
 func (r *CAPIImportReconciler) reconcileDelete(ctx context.Context, capiCluster *clusterv1.Cluster) (ctrl.Result, error) {
 	log := log.FromContext(ctx)
 	log.Info("Reconciling rancher cluster deletion")
@@ -370,20 +340,6 @@ func (r *CAPIImportReconciler) getClusterRegistrationManifest(ctx context.Contex
 	return manifestData, nil
 }
 
-func shouldImport(obj metav1.Object) (hasLabel bool, labelValue bool) {
-	labelVal, ok := obj.GetLabels()[importLabelName]
-	if !ok {
-		return false, false
-	}
-
-	autoImport, err := strconv.ParseBool(labelVal)
-	if err != nil {
-		return true, false
-	}
-
-	return true, autoImport
-}
-
 func (r *CAPIImportReconciler) rancherClusterToCapiCluster(ctx context.Context, clusterPredicate predicate.Funcs) handler.MapFunc {
 	log := log.FromContext(ctx)
 
@@ -417,7 +373,7 @@ func (r *CAPIImportReconciler) namespaceToCapiClusters(ctx context.Context, clus
 			return nil
 		}
 
-		_, autoImport := shouldImport(ns)
+		_, autoImport := util.ShouldImport(ns, importLabelName)
 		if !autoImport {
 			log.V(2).Info("Namespace doesn't have import annotation label with a true value, skipping")
 			return nil
--- a/util/predicates/cluster_predicates.go
+++ b/util/predicates/cluster_predicates.go
@@ -17,6 +17,7 @@ limitations under the License.
 package predicates
 
 import (
+	"context"
 	"strings"
 
 	"github.com/go-logr/logr"
@@ -24,6 +25,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+
+	"github.com/rancher-sandbox/rancher-turtles/util"
 	"github.com/rancher-sandbox/rancher-turtles/util/annotations"
 )
 
@@ -46,6 +50,7 @@ func ClusterWithoutImportedAnnotation(logger logr.Logger) predicate.Funcs {
 	}
 }
 
+// processIfClusterNotImported returns true if the provided object is a cluster and does not have the imported annotation.
 func processIfClusterNotImported(logger logr.Logger, obj client.Object) bool {
 	kind := strings.ToLower(obj.GetObjectKind().GroupVersionKind().Kind)
 	log := logger.WithValues("namespace", obj.GetNamespace(), kind, obj.GetName())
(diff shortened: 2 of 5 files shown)


util/predicates/cluster_predicates.go
```diff
@@ -43,3 +48,82 @@ func processIfClusterNotImported(logger logr.Logger, obj client.Object) bool {
...
+	}
+}
+
+func processIfClusterReadyControlPlane(logger logr.Logger, obj client.Object) bool {
+	kind := strings.ToLower(obj.GetObjectKind().GroupVersionKind().Kind)
+	log := logger.WithValues("namespace", obj.GetNamespace(), kind, obj.GetName())
+
+	cluster, ok := obj.(*clusterv1.Cluster)
+	if !ok {
+		panic(fmt.Sprintf("Expected a Cluster but got a %T", obj)) // it should never happen
```
Danil-Grigorev, line 77: Don’t panic here. Return false, this just means that some other object came from cache, which could potentially happen.

internal/controllers/import_controller.go
```diff
@@ -273,38 +273,6 @@ func (r *CAPIImportReconciler) reconcileNormal(ctx context.Context, capiCluster
...
 }
 
-func (r *CAPIImportReconciler) shouldAutoImport(ctx context.Context, capiCluster *clusterv1.Cluster) (bool, error) {
-	log := log.FromContext(ctx)
-	log.V(2).Info("should we auto import the capi cluster", "name", capiCluster.Name, "namespace", capiCluster.Namespace)
-
-	// Check CAPI cluster for label first
-	hasLabel, autoImport := shouldImport(capiCluster)
-	if hasLabel && autoImport {
-		log.V(2).Info("Cluster contains import annotation")
```
Danil-Grigorev, line 283: It is no longer an annotation, but a label

internal/controllers/import_controller.go
```diff
@@ -66,10 +64,12 @@ func (r *CAPIImportReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Ma
...
 
-	// TODO: we want the control plane initialized but removing for the time being
-	// capiPredicates := predicates.All(log, predicates.ClusterControlPlaneInitialized(log), predicates.ResourceHasFilterLabel(log, r.WatchFilterValue))
-	capiPredicates := predicates.All(log, predicates.ResourceHasFilterLabel(log, r.WatchFilterValue),
-		turtlespredicates.ClusterWithoutImportedAnnotation(log))
+	capiPredicates := predicates.All(log,
+		predicates.ResourceHasFilterLabel(log, r.WatchFilterValue),
+		turtlespredicates.ClusterWithoutImportedAnnotation(log),
+		turtlespredicates.ClusterWithReadyControlPlane(log),
+		turtlespredicates.ClusterOrNamespaceWithImportLabel(ctx, log, r.Client, importLabelName),
```
Danil-Grigorev, line 71: This will no longer work. Namespace is a different object, and predicates.All is an equivalent of && here. Each of the previous predicates are checking if the object is a Cluster object.

Danil-Grigorev (changes requested):

util/predicates/cluster_predicates.go
richardcase: I agree, maybe log an error and return false?

util/predicates/cluster_predicates.go
```diff
@@ -43,3 +48,82 @@ func processIfClusterNotImported(logger logr.Logger, obj client.Object) bool {
...
+	}
+}
+
+func processIfClusterOrNamespaceWithImportLabel(ctx context.Context, logger logr.Logger, cl client.Client, obj client.Object, label string) bool {
+	kind := strings.ToLower(obj.GetObjectKind().GroupVersionKind().Kind)
+	log := logger.WithValues("namespace", obj.GetNamespace(), kind, obj.GetName())
+
+	cluster, ok := obj.(*clusterv1.Cluster)
+	if !ok {
+		panic(fmt.Sprintf("Expected a Cluster but got a %T", obj)) // it should never happen
```
richardcase, line 119: Same comment as @Danil-Grigorev above with removing the use of the panic

internal/controllers/import_controller.go
richardcase: This also checks if its a Cluster object .

util/predicates/cluster_predicates_test.go
```diff
@@ -1,79 +1,148 @@
...
+	It("should return true when namespace has import label", func() {
+		namespace.Name = "test-ns-1"
+		Expect(cl.Create(ctx, namespace)).To(Succeed())
+
+		capiCluster.Namespace = namespace.Name
+		result := ClusterOrNamespaceWithImportLabel(ctx, logger, cl, importLabel).UpdateFunc(event.UpdateEvent{ObjectNew: capiCluster})
+		Expect(result).To(BeTrue())
+	})
+
+	It("should return true if client fails to get namespace", func() {
```
richardcase, line 131: Are you thinking of transient errors here?

util/util.go
```diff
@@ -0,0 +1,60 @@
+package util
```
richardcase, line 1: Could we change the package name to something different? `util` is very generic.

util/predicates/cluster_predicates_test.go
```diff
@@ -1,79 +1,148 @@
...
-				},
-			}
-			capiCluster = &clusterv1.Cluster{
-				ObjectMeta: metav1.ObjectMeta{
-					Name:        "test-cluster",
-					Namespace:   "test-ns",
-					Annotations: map[string]string{},
-				},
+			capiCluster.Annotations = map[string]string{
+				"some-other-annoations": "true",
```
furkatgofurov7, line 50:
```suggestion
				"some-random-annotation": "true",
```

util/predicates/cluster_predicates_test.go
```diff
@@ -1,79 +1,148 @@
...
+		}
+	})
+
+	It("should return true when cluster has ready control plane", func() {
+		capiCluster.Status.ControlPlaneReady = true
+		result := ClusterWithReadyControlPlane(logger).UpdateFunc(event.UpdateEvent{ObjectNew: capiCluster})
+		Expect(result).To(BeTrue())
+	})
+
+	It("should return false when cluster has no ready control plane", func() {
```
furkatgofurov7, line 82:
```suggestion
	It("should return false when cluster does not have ready control plane", func() {
```

util/util.go
furkatgofurov7: I think it was me who introduced a `util` when moving predicates to its own package in https://github.com/rancher-sandbox/rancher-turtles/pull/50/commits/6c01783876743422e93be6e8e4cf675a6f4ee708 following on how CAPI does it: https://github.com/kubernetes-sigs/cluster-api/tree/main/util

internal/controllers/import_controller.go
furkatgofurov7: +1

internal/controllers/import_controller.go
furkatgofurov7: It is a removed code, so should not matter?

internal/controllers/import_controller.go
alexander-demicev: it works as expected, the predicate checks if object is a cluster and gets namespace if label is not found

util/predicates/cluster_predicates_test.go
alexander-demicev: yes

alexander-demicev:
@furkatgofurov7 @Danil-Grigorev @richardcase all fixed, please take a look again

alexander-demicev:
@Danil-Grigorev let's merge e2e suite first :D

furkatgofurov7:
> @Danil-Grigorev let's merge e2e suite first :D

@alexander-demicev can we update the PR, e2e should be merged now

richardcase:
Apart from the action failure this looks good to me 👍

internal/controllers/import_controller.go
```diff
@@ -83,10 +83,12 @@ func (r *CAPIImportReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Ma
 		r.remoteClientGetter = remote.NewClusterClient
 	}
 
-	// TODO: we want the control plane initialized but removing for the time being
-	// capiPredicates := predicates.All(log, predicates.ClusterControlPlaneInitialized(log), predicates.ResourceHasFilterLabel(log, r.WatchFilterValue))
```
Danil-Grigorev, line 87: This comment had a better handling of predicates. Specifically it used ResourceHasFilterLabel, which used label selector. This functionality is imported from controller-runtime and does not require to be tested, so this decreases the amount of code to review. Additionally this gives confidence in what is being added with the change. Why it is not being used here?

Danil-Grigorev: This is the predicate covering 50% of what is being done within the custom logic we have - https://github.com/kubernetes-sigs/controller-runtime/blob/1449c365465511e047333f9bc504e47cb0fae1ae/pkg/predicate/predicate.go#L352

alexander-demicev: That's not going to work in this case. We have to check the cluster label anyway in the predicate because we use the namespace as a fallback option: if the cluster doesn't have a label, look at the namespace. If we remove the label check for the cluster in the predicate, then when the namespace is not labeled, it's always going to return false.

alexander-demicev:
I managed to run e2e locally and it passed

richardcase (approved):

internal/controllers/import_controller.go
Danil-Grigorev: Yeah, maybe not in the current form. This requires a separate set of predicates on each resource to function properly, and ideally a mechanism to pass these predicates to event mapper.

Danil-Grigorev (approved):

util/predicates/cluster_predicates.go
```diff
@@ -59,3 +63,84 @@ func processIfClusterNotImported(logger logr.Logger, obj client.Object) bool {
...
+	log := logger.WithValues("namespace", obj.GetNamespace(), kind, obj.GetName())
+
+	cluster, ok := obj.(*clusterv1.Cluster)
+	if !ok {
+		log.V(4).Info("Expected a Cluster but got a different object, will not attempt to map resource", "object", obj)
+		return false
+	}
+
+	if !cluster.Status.ControlPlaneReady {
+		log.V(4).Info("Cluster does not have an ready control plane, will not attempt to map resource")
```
furkatgofurov7, line 97:
```suggestion
		log.V(4).Info("Cluster does not have a ready control plane, will not attempt to map resource")
```

util/predicates/cluster_predicates.go
```diff
@@ -59,3 +63,84 @@ func processIfClusterNotImported(logger logr.Logger, obj client.Object) bool {
...
+		return false
+	}
+
+	log.V(6).Info("Cluster has an ready control plane, will attempt to map resource")
```
furkatgofurov7, line 101:
```suggestion
	log.V(6).Info("Cluster has a ready control plane, will attempt to map resource")
```

util/predicates/cluster_predicates.go
```diff
@@ -59,3 +63,84 @@ func processIfClusterNotImported(logger logr.Logger, obj client.Object) bool {
...
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return processIfClusterReadyControlPlane(logger.WithValues("predicate", "ClusterWithReadyControlPlane", "eventType", "delete"), e.Object)
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return processIfClusterReadyControlPlane(logger.WithValues("predicate", "ClusterWithReadyControlPlane", "eventType", "generic"), e.Object)
+		},
+	}
+}
+
+func processIfClusterReadyControlPlane(logger logr.Logger, obj client.Object) bool {
```
furkatgofurov7, line 86: can we add a godoc as we have for other funcs?

util/predicates/cluster_predicates.go
```diff
@@ -59,3 +63,84 @@ func processIfClusterNotImported(logger logr.Logger, obj client.Object) bool {
...
+
+	cluster, ok := obj.(*clusterv1.Cluster)
+	if !ok {
+		log.V(4).Info("Expected a Cluster but got a different object, will not attempt to map resource", "object", obj)
+		return false
+	}
+
+	shouldImport, err := util.ShouldAutoImport(ctx, log, cl, cluster, label)
+	if err != nil {
+		log.Error(err, "Client failed to get namespace, this error is ignored returning true")
```
furkatgofurov7, line 141: this is not always the case, `ShouldAutoImport` can return true also when there is an import annotation set on the cluster object. Reading this log does not help much tbh, I would suggest rather having something like:

```suggestion
		log.Error(err, "namespace or cluster has already import annotation set, ignoring it")
```

util/predicates/cluster_predicates_test.go
```diff
@@ -17,79 +17,148 @@ limitations under the License.
...
-						"some-other-annoations": "true",
-					},
-				},
-			}
-			capiCluster = &clusterv1.Cluster{
-				ObjectMeta: metav1.ObjectMeta{
-					Name:      "test-cluster",
-					Namespace: "test-ns",
-				},
-			}
```
furkatgofurov7, line 67: yeah this is not needed, but I would rather leave it for readiness of the code

util/predicates/cluster_predicates_test.go
```diff
@@ -17,79 +17,148 @@ limitations under the License.
...
-				ObjectMeta: metav1.ObjectMeta{
-					Annotations: map[string]string{"some-random-annoation": "true"},
-				},
-			}
-			capiCluster = &clusterv1.Cluster{
-				ObjectMeta: metav1.ObjectMeta{
-					Name:        "test-cluster",
-					Namespace:   "test-ns",
-					Annotations: map[string]string{},
-				},
```
furkatgofurov7, line 84: same as above

furkatgofurov7 (commented):
Few nits, overall lgtm:

alexander-demicev:
@furkatgofurov7 Thanks for the review, I addressed all comments except for the last ones about tests. I like using before each blocks to set up objects for tests and then only assign parameters that are different inside it blocks. I'm not strongly against repeating object setup but it just looks prettier to me with before each.

furkatgofurov7 (approved):

PR #71, opened 2023-08-11
🌱 Use utilyaml to collect agent manifests in remote cluster

Danil-Grigorev:
kind/cleanup

Using CAPI utilyaml to collect and apply raw manifests from import endpoint.
- Improve logging
- Reduce number of remote cluster requests from 2 to 1 (Create)

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/internal/controllers/import_controller.go
+++ b/internal/controllers/import_controller.go
@@ -3,7 +3,6 @@ package controllers
 import (
 	"bufio"
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -32,6 +31,7 @@ import (
 	"sigs.k8s.io/cluster-api/controllers/external"
 	"sigs.k8s.io/cluster-api/controllers/remote"
 	"sigs.k8s.io/cluster-api/util/predicates"
+	utilyaml "sigs.k8s.io/cluster-api/util/yaml"
 
 	"github.com/rancher-sandbox/rancher-turtles/internal/rancher"
 	turtlesannotations "github.com/rancher-sandbox/rancher-turtles/util/annotations"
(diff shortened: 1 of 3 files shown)


internal/controllers/import_controller.go
```diff
@@ -472,69 +472,35 @@ func (r *CAPIImportReconciler) createImportManifest(ctx context.Context, remoteC
...
 	gvk := obj.GetObjectKind().GroupVersionKind()
 
-	u := &unstructured.Unstructured{}
-	u.SetGroupVersionKind(gvk)
-	err := c.Get(ctx, client.ObjectKey{
-		Namespace: obj.GetNamespace(),
-		Name:      obj.GetName(),
-	}, u)
+	err := c.Create(ctx, obj)
+	if client.IgnoreAlreadyExists(err) != nil {
```
richardcase, line 494: Whats the difference between this check and the `apierrors.IsAlreadyExists` below?

Danil-Grigorev: The first one filters out the AlreadyExists error, and returns all other error types or nil. The other one specifically checks for the error being of type AlreadyExist. I hoped this will be more readable, but may be it isn’t. Can change to err != nil for the second one, it is equivalent.

furkatgofurov7: we could move maybe this:

`if apierrors.IsAlreadyExists(err) {}` up to check for api errors and later check for err != nil since it will be not api errors anymore?

Danil-Grigorev: I’ve updated the implementation. Though this change brings no functional difference, just the code style.

furkatgofurov7 (approved):

alexander-demicev (approved):

Issue #19, opened 2023-06-20
Ensure operator can run side-by-side with Rancher


richardcase:
**Story**

As a Rancher Operator,
I want to be able to install Rancehr Turtles in the same cluster as Rancher Manager.
So that i can provision clusters with CAPI and the existing Rancher provisioning mechanisms

**Description**

This means that v2prov should continue to work and we shouldn't be required to disable any of the features in Rancher to use the CAPI extension.

This requires some small changes to Rancher (issue already created).

PR #82, opened 2023-08-16
Support out-of-cluster installation

salasberryfin:

This is intended to allow a Rancher operator to choose between deploying rancher-turtles in the same or a different cluster to Rancher Manager. The current implementation of rancher-turtles defaults to an in-cluster configuration with both Rancher Manager and rancher-turtles supposed to be installed together in the same cluster.

The initial idea is to create the client based on whether a flag with a path to a kubeconfig file is passed. When this parameter is used, rancher-turtles will load the REST Config from this file and create a client for an out-of-cluster installation. If it is not, it defaults to the current in-cluster behavior.

Issue #19 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/internal/controllers/import_controller.go
+++ b/internal/controllers/import_controller.go
@@ -65,7 +65,8 @@ const (
 
 // CAPIImportReconciler represents a reconciler for importing CAPI clusters in Rancher.
 type CAPIImportReconciler struct {
-	client.Client
+	Client           client.Client
+	RancherClient    client.Client
 	recorder         record.EventRecorder
 	WatchFilterValue string
 	Scheme           *runtime.Scheme
@@ -178,7 +179,7 @@ func (r *CAPIImportReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		errs = append(errs, fmt.Errorf("error reconciling cluster: %w", err))
 	}
 
-	if err := r.Patch(ctx, capiCluster, patchBase); err != nil {
+	if err := r.Client.Patch(ctx, capiCluster, patchBase); err != nil {
 		errs = append(errs, fmt.Errorf("failed to patch cluster: %w", err))
 	}
 
@@ -191,7 +192,7 @@ func (r *CAPIImportReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 
 func (r *CAPIImportReconciler) reconcile(ctx context.Context, capiCluster *clusterv1.Cluster) (ctrl.Result, error) {
 	// fetch the rancher clusters
-	rancherClusterHandler := rancher.NewClusterHandler(ctx, r.Client)
+	rancherClusterHandler := rancher.NewClusterHandler(ctx, r.RancherClient)
 	rancherClusterName := turtelesnaming.Name(capiCluster.Name).ToRancherName()
 
 	rancherCluster, err := rancherClusterHandler.Get(client.ObjectKey{Namespace: capiCluster.Namespace, Name: rancherClusterName})
@@ -346,7 +347,7 @@ func (r *CAPIImportReconciler) getClusterRegistrationManifest(ctx context.Contex
 	log := log.FromContext(ctx)
 
 	key := client.ObjectKey{Name: clusterRegistrationTokenName, Namespace: clusterName}
-	tokenHandler := rancher.NewClusterRegistrationTokenHandler(ctx, r.Client)
+	tokenHandler := rancher.NewClusterRegistrationTokenHandler(ctx, r.RancherClient)
 
 	token, err := tokenHandler.Get(key)
 	if err != nil {
--- a/main.go
+++ b/main.go
@@ -28,6 +28,9 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	"k8s.io/component-base/version"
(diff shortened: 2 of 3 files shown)


main.go
```diff
@@ -97,6 +101,9 @@ func initFlags(fs *pflag.FlagSet) {
 
 	fs.IntVar(&concurrencyNumber, "concurrency", 1,
 		"Number of resources to process simultaneously")
+
+	fs.StringVar(&rancherKubeconfig, "rancher-kubeconfig", "",
+		"Path to a kubeconfig file. Only required if out-of-cluster.")
```
richardcase, line 106:
```suggestion
		"Path to the Rancher kubeconfig file. Only required if running out-of-cluster.")
```

main.go
```diff
@@ -154,12 +161,48 @@ func setupChecks(mgr ctrl.Manager) {
...
+			os.Exit(1)
+		}
+	} else {
+		setupLog.Info("in-cluster installation of rancher-turtles")
+		c = mgr.GetClient()
+	}
+
 	if err := (&controllers.CAPIImportReconciler{
-		Client:           mgr.GetClient(),
+		Client:           c,
```
richardcase, line 190: We actually need 2 clients for `CAPIImportReconciler`:

- 1 for the controller itself for the CAPI Clusters
- 1 for Rancher

So maybe something like this:

```go
type CAPIImportReconciler struct {
	Client client.Client
        RancherClient client.Client
```

In the scenario where Rancher Manager and CAPI are in the same cluster then:
- **Client** is set to `mgr.GetClient()`
- **RancherClient**  is set to `mgr.GetClient()`

In the scenario for this change (i.e. where the CAPI management cluster is different to the Rancher Manager cluster) , then:
- **Client** is set to `mgr.GetClient()`
- **RancherClient** is set to client created from the kubeconfig specified via **rancherKubeconfig**

How does that sound?

salasberryfin: @richardcase updated the PR with your suggestions.

main.go
```diff
@@ -97,6 +101,9 @@ func initFlags(fs *pflag.FlagSet) {
 
 	fs.IntVar(&concurrencyNumber, "concurrency", 1,
 		"Number of resources to process simultaneously")
+
+	fs.StringVar(&rancherKubeconfig, "rancher-kubeconfig", "",
```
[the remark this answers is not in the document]
salasberryfin: Yes, this has to available in the pod.

internal/controllers/import_controller.go
```diff
@@ -257,7 +258,7 @@ func (r *CAPIImportReconciler) reconcileNormal(ctx context.Context, capiCluster
 		Namespace: capiCluster.Namespace,
 	}
 
-	remoteClient, err := r.remoteClientGetter(ctx, capiCluster.Name, r.Client, clusterKey)
+	remoteClient, err := r.remoteClientGetter(ctx, capiCluster.Name, r.RancherClient, clusterKey)
```
Danil-Grigorev, line 261: If the providers are installed outside of rancher management cluster, then likely the secret to connect to the CAPI cluster will live inside this “other then rancher manager” cluster. So here for me it makes sense to use regular in-cluster client to connect to the provisioned cluster by CAPI from the cluster where CAPI is installed.

main.go
```diff
@@ -155,11 +162,54 @@ func setupChecks(mgr ctrl.Manager) {
...
+		RancherClient:    rancherClient,
 		WatchFilterValue: watchFilterValue,
 	}).SetupWithManager(ctx, mgr, controller.Options{MaxConcurrentReconciles: concurrencyNumber}); err != nil {
 		setupLog.Error(err, "unable to create capi controller")
 		os.Exit(1)
 	}
 }
+
+// setupRancherClient can either create a client for an in-cluster installation (rancher and rancher-turtles in the same cluster)
+// or create a client for an out-of-cluster installation (rancher and rancher-turtles in different clusters).
```
furkatgofurov7, line 182:
```suggestion
// setupRancherClient can either create a client for an in-cluster installation (rancher and rancher-turtles in the same cluster)
// or create a client for an out-of-cluster installation (rancher and rancher-turtles in different clusters) based on the existence of Rancher kubeconfig file.
```

furkatgofurov7 (commented):
While this is clear to the internal team, it would be nice to have a small document describing what is (out-of/in)-cluster support, the difference between them and why it is needed. 

Please disregard if we have a separate issue to document it in rancher-turtles-docs

Danil-Grigorev:
Same thought as @furkatgofurov7, but additionally we need to have some clear process to pass the kubeconfig to the pod, which could be documented in rancher-turtles-docs + not necessary in the same PR but would prefer to have e2e test coverage here to be sure.

richardcase:
@furkatgofurov7 @Danil-Grigorev @salasberryfin  - i created #94 for us to create an ADR for this.

furkatgofurov7:
> @furkatgofurov7 @Danil-Grigorev @salasberryfin - i created #94 for us to create an ADR for this.

Except [this](https://github.com/rancher-sandbox/rancher-turtles/pull/82#discussion_r1311217975), I have nothing major to not merge it

salasberryfin:
Thanks for the reviews @richardcase @Danil-Grigorev @furkatgofurov7. I'll be taking on the ADR proposal to have this well documented.

salasberryfin:
I addressed the latest comments and created a PR for the ADR #95.

richardcase (dismissed):
This looks good to me, thanks @salasberryfin 

We will need to create a documentation issue (in racher-turtles-docs) to cover:
- Handling the RBAC, service account etc
- How to create the secret and mount it

salasberryfin:
I created rancher-sandbox/rancher-turtles-docs#13 to track this.

furkatgofurov7:
@salasberryfin can we rebase it?

salasberryfin:
The changes are now up to date after rebase, please take a look.

furkatgofurov7 (approved):
Thanks @salasberryfin!

Danil-Grigorev (approved):

alexander-demicev (approved):

Issue #86, opened 2023-08-23
Gaps in rancher cluster deletion process


Danil-Grigorev:
Currently there are several issues with implementation of deletion strategy for the rancher cluster.

1. `imported=true` annotation is being set conditionally based on the fact it should also be present on the rancher cluster. It [shouldn’t](https://github.com/rancher-sandbox/rancher-turtles/pull/85/files#diff-7c0fda45174e6cff5feacf49979ba7f3f56d797cb698ce3a8f9dc4f654124ee7R18), the behavior supposed to be automatic upon cluster deletion.
2. Events coming from updates of the rancher cluster are not propagated to the CAPI cluster due to incorrect name mapping between the two.
3. Controller-runtime mapper methods do not support predicate filtering logic on enqueued requests, it triggers reconciliation on the resource directly by enqueuing the `NamespaceName` of the object and passing it into the [work queue](https://github.com/kubernetes-sigs/controller-runtime/blob/c20ea143a236a34fb331e6c04820b75aac444e7d/pkg/internal/controller/controller.go#L249). Which causes the following race condition:
    - Rancher cluster is getting deleted. This generates a mapped reconcile request to the CAPI cluster, which triggers a reconcile loop without passing predicate logic for the enqueued CAPI cluster. Rancher cluster has deletion timestamp, so the CAPI cluster is annotated and nothing else happens.
    - At some point Rancher cluster is finally removed, which, again, triggers reconcile on CAPI cluster and it follows to the reconcileNormal scenario. The Rancher cluster is re-created again and is visible in UI.

furkatgofurov7:
I think this can be closed

PR #85, opened 2023-08-22
📖 Add deletion strategy ADR

furkatgofurov7:
kind/documentation

**What this PR does / why we need it**:
Adds deletion strategy ADR

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #63 

**Special notes for your reviewer**:

**Checklist**:

- [X] squashed commits into logical changes
- [X] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

PR #87, opened 2023-08-23
:bug: Allow rancher cluster to be deleted from UI

Danil-Grigorev:
kind/bug

This addresses the problems described in #86.

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/internal/controllers/import_controller.go
+++ b/internal/controllers/import_controller.go
@@ -24,8 +24,10 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
@@ -87,7 +89,7 @@ func (r *CAPIImportReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Ma
 
 	err = c.Watch(
 		source.Kind(mgr.GetCache(), u),
-		handler.EnqueueRequestsFromMapFunc(r.rancherClusterToCapiCluster(ctx)),
+		handler.EnqueueRequestsFromMapFunc(r.rancherClusterToCapiCluster(ctx, capiPredicates)),
 		//&handler.EnqueueRequestForOwner{OwnerType: &clusterv1.Cluster{}},
 	)
 	if err != nil {
@@ -97,7 +99,7 @@ func (r *CAPIImportReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Ma
 	ns := &corev1.Namespace{}
 	err = c.Watch(
 		source.Kind(mgr.GetCache(), ns),
-		handler.EnqueueRequestsFromMapFunc(r.namespaceToCapiClusters(ctx)),
+		handler.EnqueueRequestsFromMapFunc(r.namespaceToCapiClusters(ctx, capiPredicates)),
 	)
 
 	if err != nil {
@@ -366,7 +368,7 @@ func shouldImport(obj metav1.Object) (hasLabel bool, labelValue bool) {
 	return true, autoImport
 }
 
-func (r *CAPIImportReconciler) rancherClusterToCapiCluster(ctx context.Context) handler.MapFunc {
+func (r *CAPIImportReconciler) rancherClusterToCapiCluster(ctx context.Context, clusterPredicate predicate.Funcs) handler.MapFunc {
 	log := log.FromContext(ctx)
 
 	return func(_ context.Context, o client.Object) []ctrl.Request {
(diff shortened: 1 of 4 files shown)


internal/controllers/import_controller.go
```diff
@@ -303,27 +306,21 @@ func (r *CAPIImportReconciler) shouldAutoImport(ctx context.Context, capiCluster
...
-	// then annotate the CAPI cluster so that we don't auto-import again.
-	if turtlesannotations.HasClusterImportAnnotation(rancherCluster) {
-		log.Info(fmt.Sprintf("rancher cluster %s has %s annotation, so annotating CAPI cluster %s",
-			rancherCluster.Name, turtlesannotations.ClusterImportedAnnotation, capiCluster.Name))
-
-		annotations := capiCluster.GetAnnotations()
-		if annotations == nil {
-			annotations = map[string]string{}
-		}
+	// If the Rancher Cluster was already imported, then annotate the CAPI cluster so that we don't auto-import again.
```
furkatgofurov7, line 313: where is the check in the code from this comment, I seem to be missing it

Danil-Grigorev: https://github.com/rancher-sandbox/rancher-turtles/pull/87/files#diff-df81a5620bfede50d1c70fd0ebb982e652d28721e15b825c52dddcf97ceda0deR384

furkatgofurov7: can we log the message in that case, it was not obvious from first look

furkatgofurov7 (approved):

internal/controllers/import_controller.go
Danil-Grigorev: Operator does not reconcile non-imported rancher clusters. Imported rancher cluster follows the `<capi-cluster>-capi` naming convention. Therefore it is not possible here that any of the reconcile code, including reconcile delete would be executed on non-imported rancher cluster, unless it follows given naming convention. So log message on the  https://github.com/rancher-sandbox/rancher-turtles/pull/87/files#diff-df81a5620bfede50d1c70fd0ebb982e652d28721e15b825c52dddcf97ceda0deR311 and https://github.com/rancher-sandbox/rancher-turtles/pull/87/files#diff-df81a5620bfede50d1c70fd0ebb982e652d28721e15b825c52dddcf97ceda0deR314 are both describing what is happening with the cluster. Don’t see a need to make it more vivid than that.

Danil-Grigorev:
> How does this affect #85?

It fully follows the proposal at the state as it is in right [now](https://github.com/rancher-sandbox/rancher-turtles/pull/85/commits/cc80c0a9435c10cdeeea53388e7b4496fa4b5e28).

internal/helpers/name_converter.go
```diff
@@ -0,0 +1,21 @@
+package helpers
```
alexander-demicev, line 1: let's use `util` package

internal/controllers/import_controller.go
```diff
@@ -303,27 +306,21 @@ func (r *CAPIImportReconciler) shouldAutoImport(ctx context.Context, capiCluster
 	return autoImport, nil
 }
 
-func (r *CAPIImportReconciler) reconcileDelete(ctx context.Context, capiCluster *clusterv1.Cluster,
-	rancherCluster *rancher.Cluster,
-) (ctrl.Result, error) {
+func (r *CAPIImportReconciler) reconcileDelete(ctx context.Context, capiCluster *clusterv1.Cluster) (ctrl.Result, error) {
 	log := log.FromContext(ctx)
-	log.Info("Reconciling CAPI Cluster Delete")
+	log.Info("Reconciling rancher cluster deletion")
```
alexander-demicev, line 311: but we are reconciling CAPI cluster deletion

internal/controllers/import_controller.go
```diff
@@ -303,27 +306,21 @@ func (r *CAPIImportReconciler) shouldAutoImport(ctx context.Context, capiCluster
...
-	if turtlesannotations.HasClusterImportAnnotation(rancherCluster) {
-		log.Info(fmt.Sprintf("rancher cluster %s has %s annotation, so annotating CAPI cluster %s",
-			rancherCluster.Name, turtlesannotations.ClusterImportedAnnotation, capiCluster.Name))
-
-		annotations := capiCluster.GetAnnotations()
-		if annotations == nil {
-			annotations = map[string]string{}
-		}
+	// If the Rancher Cluster was already imported, then annotate the CAPI cluster so that we don't auto-import again.
+	log.Info(fmt.Sprintf("rancher cluster is being remoted, annotating CAPI cluster %s", capiCluster.Name))
```
alexander-demicev, line 314:
```suggestion
	log.Info(fmt.Sprintf("Rancher cluster is being removed, annotating CAPI cluster %s", capiCluster.Name))
```

internal/controllers/import_controller.go
alexander-demicev: It will also be logged in the predicate itself. My concern is that as far as I know predicates might not always work as expected and we should structure our code like they don't exist, it's nice if they filter out resources for us but don't expect them to always do that. I wouldn't remove the code.

alexander-demicev (changes requested):

mjura (approved):
lgtm

internal/controllers/import_controller.go
Danil-Grigorev: Right now we have to act upon the fact that rancher cluster is getting deleted

alexander-demicev:
@Danil-Grigorev It happened about 2 years ago when we worked on externally managed infrastructure for CAPI, I can't find this discussion. If not working predicate appears, let's fix it when it happens :)

Danil-Grigorev:
Can we merge this? Capi cluster event causes rancher cluster to be re-imported as of right now. @furkatgofurov7 @alexander-demicev I don't understand the delay in reviews

furkatgofurov7:
> Can we merge this? Capi cluster event causes rancher cluster to be re-imported as of right now. @furkatgofurov7 @alexander-demicev I don't understand the delay in reviews

I did pass through it once already and not blocking it from my side as per comments?

furkatgofurov7 (approved):

richardcase (approved):

alexander-demicev (approved):

PR #93, opened 2023-08-31
Add CAPI operator as a dependency to chart

alexander-demicev:

Add CAPI operator to helm chart dependency, I will also open a follow-up PR that creates core provider before Rancher turtles are installed.
.

Fixes https://github.com/rancher-sandbox/rancher-turtles/issues/91

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,10 +109,6 @@ HELM_VER := v3.8.1
 HELM_BIN := helm
 HELM := $(TOOLS_BIN_DIR)/$(HELM_BIN)-$(HELM_VER)
 
-CAPI_OPERATOR_VER := 0.5.0
-CAPI_OPERATOR := $(abspath $(TOOLS_BIN_DIR)/capi-operator.tgz)
-CAPI_OPERATOR_URL := https://github.com/kubernetes-sigs/cluster-api-operator/releases/download/v$(CAPI_OPERATOR_VER)/cluster-api-operator-$(CAPI_OPERATOR_VER).tgz
-
 GOLANGCI_LINT_VER := v1.53.3
 GOLANGCI_LINT_BIN := golangci-lint
 GOLANGCI_LINT := $(abspath $(TOOLS_BIN_DIR)/$(GOLANGCI_LINT_BIN))
@@ -384,9 +380,6 @@ $(SETUP_ENVTEST): # Build setup-envtest from tools folder.
 $(GINKGO): # Build ginkgo from tools folder.
 	GOBIN=$(TOOLS_BIN_DIR) $(GO_INSTALL) $(GINKGO_PKG) $(GINKGO_BIN) $(GINGKO_VER)
 
-$(CAPI_OPERATOR):
-	wget $(CAPI_OPERATOR_URL) -O $(CAPI_OPERATOR)
-
 $(GOLANGCI_LINT): # Download and install golangci-lint
 	hack/ensure-golangci-lint.sh \
 		-b $(TOOLS_BIN_DIR) \
@@ -438,7 +431,7 @@ build-chart: $(HELM) $(KUSTOMIZE) $(RELEASE_DIR) $(CHART_RELEASE_DIR) $(CHART_PA
 	sed -i'' -e 's@tag: .*@tag: '"$(RELEASE_TAG)"'@' $(CHART_RELEASE_DIR)/values.yaml
 	sed -i'' -e 's@image: .*@image: '"$(CONTROLLER_IMG)"'@' $(CHART_RELEASE_DIR)/values.yaml
 	sed -i'' -e 's@imagePullPolicy: .*@imagePullPolicy: '"$(PULL_POLICY)"'@' $(CHART_RELEASE_DIR)/values.yaml
-	$(HELM) dependency update
+	cd $(CHART_RELEASE_DIR) && $(HELM) dependency update 
 	$(HELM) package $(CHART_RELEASE_DIR) --app-version=$(HELM_CHART_TAG) --version=$(HELM_CHART_TAG) --destination=$(CHART_PACKAGE_DIR)
 
 .PHONY: release-chart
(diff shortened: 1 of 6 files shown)


charts/rancher-turtles/charts/cluster-api-operator-0.5.1.tgz
Danil-Grigorev, line 1: Is this supposed to be there?

charts/rancher-turtles/values.yaml
```diff
@@ -3,3 +3,5 @@ rancherTurtles:
   tag: v0.0.0
   imagePullPolicy: Never
   namespace: rancher-turtles-system
+capiOperator:
+  enabled: true
```
Danil-Grigorev, line 7: I would prefer leaving it on disabled by default for the out-of-cluster support

charts/rancher-turtles/requirements.yaml
```diff
@@ -0,0 +1,5 @@
+dependencies:
+  - name: cluster-api-operator
+    version: v0.5.1
```
Danil-Grigorev, line 3: Could this version be parametrized?

charts/rancher-turtles/values.yaml
richardcase: I think its ok to have it enabled by default. The main is that there is an option to control it.

charts/rancher-turtles/values.yaml
```diff
@@ -3,3 +3,5 @@ rancherTurtles:
   tag: v0.0.0
   imagePullPolicy: Never
   namespace: rancher-turtles-system
+capiOperator:
```
richardcase, line 6: Would it also be worth exposing some of the values of the CAPI operator? That way when we install Rancher Turtles we can also say which providers to install (or not install cert-manager)  for example

charts/rancher-turtles/charts/cluster-api-operator-0.5.1.tgz
alexander-demicev: Yes, helm requires dependencies to be vendored like this

charts/rancher-turtles/requirements.yaml
alexander-demicev: It depends, we can pass it from Makefile by patching this file or if we want to have a somehow functional chart in `charts`(not like in capi operator where it's assembled on release command) just leave it as it is. The same applies to keeping CAPI operator chart tarball.

charts/rancher-turtles/values.yaml
alexander-demicev: Helm supports passing arguments to dependencies, you can do `--set cluster-api-operator.enabled=true`

charts/rancher-turtles/values.yaml
alexander-demicev: you can do it, `--set cluster-api-operator.infrastructure=aws` should work. We tried it with cert-manager in CAPI operator

charts/rancher-turtles/charts/cluster-api-operator-0.5.1.tgz
Danil-Grigorev: I checked and it could be added to gitignore. The `build-chart` makefile task already downloads the file, so the presence of the file is ensured during release. https://github.com/helm/charts/blob/390ee6614b4bb0da96b397f9b382e3ddca9f59a4/.gitignore#L35

charts/rancher-turtles/values.yaml
Danil-Grigorev: Could you please update e2e tests so they won't break because of this change and incorporate the described functionality?

alexander-demicev:
I added embedded CAPI chart to e2e, ptal

Danil-Grigorev (dismissed):
LGTM

charts/rancher-turtles/values.yaml
```diff
@@ -3,3 +3,5 @@ rancherTurtles:
   tag: v0.0.0
   imagePullPolicy: Never
   namespace: rancher-turtles-system
+cluster-api-operator:
+  enabled: true
```
Danil-Grigorev, line 7: Should this also include `create-namespace` and `namespace` flags? So it won’t be installed in the default namespace.

Makefile
```diff
@@ -438,7 +431,7 @@ build-chart: $(HELM) $(KUSTOMIZE) $(RELEASE_DIR) $(CHART_RELEASE_DIR) $(CHART_PA
 	sed -i'' -e 's@tag: .*@tag: '"$(RELEASE_TAG)"'@' $(CHART_RELEASE_DIR)/values.yaml
 	sed -i'' -e 's@image: .*@image: '"$(CONTROLLER_IMG)"'@' $(CHART_RELEASE_DIR)/values.yaml
 	sed -i'' -e 's@imagePullPolicy: .*@imagePullPolicy: '"$(PULL_POLICY)"'@' $(CHART_RELEASE_DIR)/values.yaml
-	$(HELM) dependency update
+	cd $(CHART_RELEASE_DIR) && $(HELM) dependency update
```
furkatgofurov7, line 434: nitty: space at the end of the line

test/e2e/e2e_suite_test.go
```diff
@@ -234,16 +233,17 @@ func initBootstrapCluster(bootstrapClusterProxy framework.ClusterProxy, config *
...
 	Expect(os.MkdirAll(logFolder, 0750)).To(Succeed(), "Invalid argument. Log folder can't be created for initBootstrapCluster")
 
-	initCAPIOperator(bootstrapClusterProxy, config)
 	initRancherTurtles(bootstrapClusterProxy, config)
 	initRancher(bootstrapClusterProxy, config)
 }
 
 func initRancherTurtles(clusterProxy framework.ClusterProxy, config *clusterctl.E2EConfig) {
+	By("Adding docker variables secret")
+	Expect(clusterProxy.Apply(ctx, dockerVariablesSecret)).To(Succeed())
```
furkatgofurov7, line 245: Q: where is this `dockerVariablesSecret` being created or supplied from? What does it contain?

charts/rancher-turtles/values.yaml
alexander-demicev: it gets the same namespace as rancher turtles by default, which I guess is ok for us

test/e2e/e2e_suite_test.go
alexander-demicev: It's taken from https://github.com/rancher-sandbox/rancher-turtles/blob/main/test/e2e/resources/config/docker-infra-secret.yaml, this is needed for CAPI operator to substitute variables in provider components.

Danil-Grigorev (approved):

furkatgofurov7 (approved):

salasberryfin (approved):

PR #109, opened 2023-09-06
✨ Install CAPI using operator after chart is installed

alexander-demicev:
kind/feature

Add new values and template to helm chart so we can deploy CAPI with one command, assuming rancher and cert-manager are running users will be able to install extension, CAPI operator and CAPI like this:
```
helm repo add rancher-turtles https://rancher-sandbox.github.io/rancher-turtles
helm repo update
helm install rancher-turtles rancher-turtles/rancher-turtles --create-namespace -n rancher-turtles-system
```

Fixes https://github.com/rancher-sandbox/rancher-turtles/issues/18
Fixes https://github.com/rancher-sandbox/rancher-turtles/issues/105

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/test_chart.yaml
+++ b/.github/workflows/test_chart.yaml
@@ -59,7 +59,7 @@ jobs:
         run: kind load docker-image ${{ env.MANIFEST_IMG }}:${{ env.TAG }}
 
       - name: Run chart-testing (install)
-        run: helm install rancher-turtles out/charts/rancher-turtles/ -n rancher-turtles-system --create-namespace --wait --set cluster-api-operator.cert-manager.enabled=true --set clusterAPI.enabled=false
+        run: helm install rancher-turtles out/charts/rancher-turtles/ -n rancher-turtles-system --create-namespace --wait --set cluster-api-operator.cluster-api.enabled=false --set cluster-api-operator.enabled=false
 
       - name: Run chart-testing (un-install)
         run: helm uninstall rancher-turtles -n rancher-turtles-system --wait
--- a/.github/workflows/test_chart.yaml
+++ b/.github/workflows/test_chart.yaml
@@ -59,7 +59,7 @@ jobs:
         run: kind load docker-image ${{ env.MANIFEST_IMG }}:${{ env.TAG }}
 
       - name: Run chart-testing (install)
-        run: helm install rancher-turtles out/charts/rancher-turtles/ -n rancher-turtles-system --create-namespace --wait --set cluster-api-operator.cert-manager.enabled=true
+        run: helm install rancher-turtles out/charts/rancher-turtles/ -n rancher-turtles-system --create-namespace --wait --set cluster-api-operator.cert-manager.enabled=true --set clusterAPI.enabled=false
 
       - name: Run chart-testing (un-install)
         run: helm uninstall rancher-turtles -n rancher-turtles-system --wait
--- a/README.md
+++ b/README.md
@@ -41,29 +41,31 @@ rancherTurtles:
   tag: v0.0.0 # tag to use for the extension
   imagePullPolicy: Never # image pull policy to use for the extension
   namespace: rancher-turtles-system # namespace to deploy to (default: rancher-turtles-system)
-clusterAPI:
-  enabled: true # indicates if core CAPI controllers should be installed (default: true)
-  version: v1.4.6 # version of CAPI to install (default: v1.4.6)
-  configSecret:
-    name: "" # name of the config secret to use for core CAPI controllers, used by the CAPI operator. See [CAPI operator](https://github.com/kubernetes-sigs/cluster-api-operator/tree/main/docs#installing-azure-infrastructure-provider) docs for more details.
-    namespace: "" # namespace of the config secret to use for core CAPI controllers, used by the CAPI operator.
-  core:
-    namespace: capi-system
-    fetchConfig: # (only required for airgapped environments)
-      url: ""  # url to fetch config from, used by the CAPI operator. See [CAPI operator](https://github.com/kubernetes-sigs/cluster-api-operator/tree/main/docs#provider-spec) docs for more details.
-      selector: ""  # selector to use for fetching config, used by the CAPI operator.
-  kubeadmBootstrap:
-    namespace: capi-kubeadm-bootstrap-system
-    fetchConfig:
-      url: ""
-      selector: ""
-  kubeadmControlPlane:
-    namespace: capi-kubeadm-control-plane-system
-    fetchConfig:
-      url: ""
-      selector: ""
 cluster-api-operator: 
   enabled: true # indicates if CAPI operator should be installed (default: true)
+  cert-manager:
+    enabled: true # indicates if cert-manager should be installed (default: true)
+  cluster-api:
+    enabled: true # indicates if core CAPI controllers should be installed (default: true)
+    version: v1.4.6 # version of CAPI to install (default: v1.4.6)
+    configSecret:
+      name: "" # name of the config secret to use for core CAPI controllers, used by the CAPI operator. See [CAPI operator](https://github.com/kubernetes-sigs/cluster-api-operator/tree/main/docs#installing-azure-infrastructure-provider) docs for more details.
+      namespace: "" # namespace of the config secret to use for core CAPI controllers, used by the CAPI operator.
+    core:
+      namespace: capi-system
+      fetchConfig: # (only required for airgapped environments)
+        url: ""  # url to fetch config from, used by the CAPI operator. See [CAPI operator](https://github.com/kubernetes-sigs/cluster-api-operator/tree/main/docs#provider-spec) docs for more details.
+        selector: ""  # selector to use for fetching config, used by the CAPI operator.
+    kubeadmBootstrap:
+      namespace: capi-kubeadm-bootstrap-system
+      fetchConfig:
+        url: ""
+        selector: ""
+    kubeadmControlPlane:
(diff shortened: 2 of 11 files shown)


alexander-demicev:
Should the kubeadm bootstrap and control plane also be included by default? or at least give an option to install them too. I'm fine with both approaches but would like to know what others think.

charts/rancher-turtles/values.yaml
```diff
@@ -3,5 +3,14 @@ rancherTurtles:
   tag: v0.0.0
   imagePullPolicy: Never
   namespace: rancher-turtles-system
+capi:
+  enabled: true
+  namespace: capi-system
+  configSecret:
```
Danil-Grigorev, line 9: Should this be included before 0.5.2 version is released?

test/e2e/e2e_suite_test.go
```diff
@@ -254,11 +254,50 @@ func initRancherTurtles(clusterProxy framework.ClusterProxy, config *clusterctl.
 	}
 	_, err := chart.Run(map[string]string{
 		"cluster-api-operator.cert-manager.enabled": "true",
```
Danil-Grigorev, line 256: As the proxy flags for capi have a different name, have you considered unified naming patter here?

test/e2e/e2e_suite_test.go
```diff
@@ -254,11 +254,50 @@ func initRancherTurtles(clusterProxy framework.ClusterProxy, config *clusterctl.
...
-		"cluster-api-operator.infrastructure":       config.GetVariable(capiInfrastructure),
-		"cluster-api-operator.secretName":           "variables",
-		"cluster-api-operator.secretNamespace":      "default",
+		"capi.configSecret.namespace":               "default",
+		"capi.configSecret.name":                    "variables",
 	})
 	Expect(err).ToNot(HaveOccurred())
+
```
Danil-Grigorev, line 261: IMO this test case should be added in the CAPI operator repository (nit)

README.md
```diff
@@ -10,7 +10,59 @@ Currently this project has the following functionality:
...
+    namespace: # namespace of the config secret to use for core CAPI controllers, used by the CAPI operator.
+  fetchConfig: # (only required for airgapped environments)
+    url: # url to fetch config from, used by the CAPI operator. See [CAPI operator](https://github.com/kubernetes-sigs/cluster-api-operator/tree/main/docs#provider-spec) docs for more details.
+    selector: # selector to use for fetching config, used by the CAPI operator.
+cluster-api-operator:
+  enabled: # indicates if CAPI operator should be installed (default: true)
+```
+### Installing CAPI providers
+
+The Rancher turtles extension does not install any CAPI providers, you will need to install them yourself using [CAPI operator](https://github.com/kubernetes-sigs/cluster-api-operator/tree/main/docs).
```
Danil-Grigorev, line 58: But you can, using infrastructure flag passed to helm values, no?

.github/workflows/test_chart.yaml
```diff
@@ -59,7 +59,7 @@ jobs:
         run: kind load docker-image ${{ env.MANIFEST_IMG }}:${{ env.TAG }}
 
       - name: Run chart-testing (install)
-        run: helm install rancher-turtles out/charts/rancher-turtles/ -n rancher-turtles-system --create-namespace --wait --set cluster-api-operator.cert-manager.enabled=true
+        run: helm install rancher-turtles out/charts/rancher-turtles/ -n rancher-turtles-system --create-namespace --wait --set cluster-api-operator.cert-manager.enabled=true --set capi.enabled=false
```
Danil-Grigorev, line 62: This workflow is the closest to quickstart process of the rancher turtles. Does this mean that capi operator will be disabled by default?

test/e2e/e2e_suite_test.go
alexander-demicev: you mean capi-operator vs cluster-api-operator?

.github/workflows/test_chart.yaml
alexander-demicev: Uninstalling doesn't really work for us right now properly due to CAPI operator not cleaning up providers before removing the controller. We can fix it on the CAPI operator side(preferable way) or document that users have to remove all providers manually before deleting the chart. 
I don't want CAPI operator blocking our progress on the extension so disabling CAPI is a temporary solution

alexander-demicev: We are planning a v0.6.0 release for operator so fix can be included there

README.md
alexander-demicev: Not really, in this case, CAPI operator chart will deploy the core CAPI provider too. We tried to follow the same pattern as clusterctl here, our deployment model is a bit more complicated. What can help here is a clusterctl plugin for operator or alternatively rancher users will get a UI for providers

charts/rancher-turtles/values.yaml
alexander-demicev: v1alpha2 is for v0.6.0 we can reuse the same values, the difference will be in the provider manifest, we also have conversion webhooks to handle the transition.

test/e2e/e2e_suite_test.go
alexander-demicev: Can you elaborate? what exactly has to be tested in CAPI operator?

Danil-Grigorev: Is [this](https://github.com/rancher-sandbox/rancher-turtles/pull/109/files#diff-eb068e1dbcbf03afb60148e8974e03075b485f799c46b05643c3c2292c0b2154R1-R41 ) supposed to be installed together with operator by the users? Why testing kubeadm provider here when it could be tested in the CAPI operator repository? Additionally with helm uninstall issue, this blocks other operations within e2e tests, it surely missing manual uninstall part in this specific test case. It is not scalable.

alexander-demicev: I will add a manual uninstall until this is fixed in operator

richardcase:
> Should the kubeadm bootstrap and control plane also be included by default? or at least give an option to install them too. I'm fine with both approaches but would like to know what others think.

I think initially, let's just install the core controllers. We want to favour Rancher solutions so i think Kubeadm will always be an optional install.

test/e2e/e2e_suite_test.go
alexander-demicev: so after some thinking, I don't believe we have to uninstall manually providers in e2e, we tear down the entire management cluster and if we need to debug the test run and skip cleanup it's better to leave the providers running, wdyt?

test/e2e/e2e_suite_test.go
alexander-demicev: I've just tried different approaches here, there will still be inconsistency. Helm doesn't like`-` character in go template so `cluster-api.something` is not an option, I changed it to `clusterAPI` and then `cluster-api-operator.enabled` to `clusterAPIOperator.enabled`, and if you want to pass values to capi operator chart you have to use `chartname.something` which is `cluster-api-operator.something`

alexander-demicev: I guess it's better to keep `cluster-api-operator.enabled` to match with other actions related to operator

alexander-demicev:
@richardcase @Danil-Grigorev please take a look again, all works in e2e

test/e2e/e2e_suite_test.go
Danil-Grigorev: This implementation fixes uninstall process in capi operator - https://github.com/kubernetes-sigs/cluster-api-operator/pull/251 so no need to do some unnessesary cluster teardowns, we just need to merge it. Creating and tearing down management clusters is costly, not scalable and will prolong the e2e test execution.

charts/rancher-turtles/templates/core-provider.yaml
```diff
@@ -0,0 +1,71 @@
+{{- if .Values.clusterAPI.enabled }}
```
Danil-Grigorev, line 1: I believe you’ve hit the issue with the naming here. I looked into this in the uninstall [PR](https://github.com/kubernetes-sigs/cluster-api-operator/pull/251/files#diff-5f2ca550f95c3e34df471651fd2d2a0ce91c55c35127672c7ed6ebbdd0109613R1) You can still use the 
`cluster-api-operator` name like this:
```suggestion
{{- if index .Values "cluster-api-operator" "enabled" }}
```

.github/workflows/test_chart.yaml
```diff
@@ -59,7 +59,7 @@ jobs:
         run: kind load docker-image ${{ env.MANIFEST_IMG }}:${{ env.TAG }}
 
       - name: Run chart-testing (install)
-        run: helm install rancher-turtles out/charts/rancher-turtles/ -n rancher-turtles-system --create-namespace --wait --set cluster-api-operator.cert-manager.enabled=true
+        run: helm install rancher-turtles out/charts/rancher-turtles/ -n rancher-turtles-system --create-namespace --wait --set cluster-api-operator.cert-manager.enabled=true --set clusterAPI.enabled=false
```
Danil-Grigorev, line 62: A bit unrelated to this PR, but why do we need to add capi operator chart repor [there](https://github.com/rancher-sandbox/rancher-turtles/pull/109/files#diff-6e5e920e8ceef9fedf6bc5cd6ac8e0774002e0be57cb1f15b8860a66d64570e7R45)? I think this job will work without it.

Danil-Grigorev: Tried that, see the error from chart-tester, resolved.

richardcase (approved):
Looking good!

PR #112, opened 2023-09-07
docs: rancher integration strategy adr

### LINKED ISSUE #64: ADR: Why we use unstructured

As a developer
I want an ADR to document our use of unstructured
So future engineers understand the decision

### Detailed Description

We need an ADR to document why we use unstructured when reading/creating Rancher specific CRs

_No response_

### Label(s) to be applied

/kind feature

richardcase:

Documents the decision around our integration with Rancher

Fixes #64


furkatgofurov7 (approved):
Thanks!

Danil-Grigorev (approved):

PR #125, opened 2023-09-13
fix: use rancherClient in reconcileNormal for out of cluster support

### LINKED ISSUE #124: [BUG] Client configuration fails with out of cluster installation
## Description

As reported by @alexander-demicev, support for installation of Rancher Turtles in a different cluster from Rancher Manager (https://github.com/rancher-sandbox/rancher-turtles/pull/82) is failing due to an incorrect use of clients ("same-cluster-client" being used when "out-of-cluster-client" should be configured).

In addition to this, the [watcher](https://github.com/rancher-sandbox/rancher-turtles/blob/main/internal/controllers/import_controller.go#L104) is looking for Rancher Cluster in the same k8s cluster. Will be opening a different issue to address the problem with the watcher.

salasberryfin:

This fixes a few bugs reported on the support for Rancher Turtles being installed outside of the Rancher Manager cluster, which includes not setting clients properly for the different scenarios #124.

Fixes #124 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/internal/controllers/import_controller.go
+++ b/internal/controllers/import_controller.go
@@ -214,7 +214,7 @@ func (r *CAPIImportReconciler) reconcileNormal(ctx context.Context, capiCluster 
 ) (ctrl.Result, error) {
 	log := log.FromContext(ctx)
 
-	err := r.Client.Get(ctx, client.ObjectKeyFromObject(rancherCluster), rancherCluster)
+	err := r.RancherClient.Get(ctx, client.ObjectKeyFromObject(rancherCluster), rancherCluster)
 	if apierrors.IsNotFound(err) {
 		shouldImport, err := util.ShouldAutoImport(ctx, log, r.Client, capiCluster, importLabelName)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


alexander-demicev (approved):

Danil-Grigorev (changes requested):
What about watches functionality for the rancher cluster? IMO to support this fully should create 2 managers, one of them leveraging current `loadConfigWithContext` or something else providing `rest.Config` and watching rancher objects we create with that.

alexander-demicev:
@Danil-Grigorev yes, I will open a separate issue for discussing that. I think this PR can be merged as is to at least fix using proper clients.

salasberryfin:
The initial issue I opened includes both clients and watchers #124 but we can split this into two, if you think that makes sense.

furkatgofurov7 (approved):

Danil-Grigorev (approved):
Thanks for updating the issue, lgtm.

PR #137, opened 2023-09-20
Run e2e tests on every PR in an isolated environment

alexander-demicev:

- Introduce `--insecure-skip-verify` flag set to false by default, when set controller will skip tls verification when doing http get request of import manifests. The flag should only be used for debugging/testing.
- Add an option to customize controller flags in helm chart(for usage see e2e suite setup). I will updated docs in a follow-up PR.
- Moved controller Deployment from `config` to `charts` because we need to use arrays in helm templates and kustomize can't work with go template, we don't expect the deployment to change often so can have it static instead of building with kustomize.
- Add an isolated mode for running E2E with CAPD, in this mode instead of ngrok we use kind node IP and sslip.io to generate a server url. In this mode rancher server is only accessible to hosts running on the same network, this allows us to test CAPD(or other providers where hosts run locally).
- When using isolated mode, install nginx ingress instead of ngrok https://kind.sigs.k8s.io/docs/user/ingress/#ingress-nginx
- Add github workflow for running on each PR and a second one as nightly run. We will run only CAPD tests on each PR and a full suite every night, actual separation of the tests will be done in a follow-up as we have some refactoring in progress.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


test/e2e/helpers_test.go
```diff
@@ -47,6 +47,9 @@ var (
 
 	//go:embed resources/config/rancher-setting-patch.yaml
 	rancherSettingPatch []byte
+
+	//go:embed resources/config/nginx-ingress.yaml
```
Danil-Grigorev, line 51: It is better to have nginx installed with helm here, in case we’d need to use different image or version in the future. Plus waiting for rollout could be delegated to helm in `setupIngress`

test/e2e/resources/config/nginx-ingress.yaml
```diff
@@ -0,0 +1,660 @@
...
+apiVersion: networking.k8s.io/v1
+kind: IngressClass
+metadata:
+  labels:
+    app.kubernetes.io/component: controller
+    app.kubernetes.io/instance: ingress-nginx
+    app.kubernetes.io/name: ingress-nginx
+    app.kubernetes.io/part-of: ingress-nginx
+    app.kubernetes.io/version: 1.9.0-beta.0
+  name: nginx
```
Danil-Grigorev, line 624: There is already a setup for rancher ingress in `test/e2e/resources/config/ingress-class-patch.yaml` so you can set name configurable there like ${INGRESS_NAME} (`nginx` or `ngrok`), it will apply the `ingressclass.kubernetes.io/is-default-class` label to it, so rancher will be accessible when it will apply `test/e2e/resources/config/ingress.yaml` as it is

test/e2e/e2e_suite_test.go
```diff
@@ -459,6 +483,51 @@ func dumpBootstrapClusterLogs(bootstrapClusterProxy framework.ClusterProxy) {
 	}
 }
 
+// configureIsolatedEnvironment sets the isolatedHostName variable to the IP of the first and only node in the boostrap cluster. Labels the node with
+// "ingress-ready" so that the nginx ingress controller can pick it up, required by kind. See: https://kind.sigs.k8s.io/docs/user/ingress/#create-cluster
```
Danil-Grigorev, line 487: This ingress-ready label is only needed because of it being specified in the nginx ingress spec. I think it could be dropped for the use of existing `https://kubernetes.io/docs/reference/labels-annotations-taints/#node-role-kubernetes-io-control-plane` label, which is already present

test/e2e/helpers_test.go
alexander-demicev: I like that these manifests are provided as an example with proper ports, services, etc. configured to work with a kind cluster. It's a no-brainer for a test environment. We are very unlikely to change it in the future. Kind docs also list it at a recommended way to install.

test/e2e/resources/config/nginx-ingress.yaml
alexander-demicev: replied above, it's just easier to apply static manifests

test/e2e/e2e_suite_test.go
alexander-demicev: There is a node selector in Deployment that requires this label to be present.

richardcase (commented):
Seems good to me. I was time poor, so will leave to others to approve.

test/e2e/e2e_suite_test.go
Danil-Grigorev: It also requires a kubeadm configuration to set these labels on the node, which we could not pass. But we have 1 control plane and so the destination is always known

test/e2e/helpers_test.go
Danil-Grigorev: Though you still base the connectivity setup on the existing ingress configuration for the rancher pod, where the example suggests to open extra port mappings you don’t (can’t) add, so what we have here is not a  recommended way. Because of that this is not guarantied that this configuration won’t change in the future if we change ingress config.

.github/workflows/e2e-every-pr.yaml
```diff
@@ -0,0 +1,31 @@
+name: Run e2e tests
+
+on:
+  push:
+    branches:
+      - "main"
+  pull_request:
+
+env:
+  RANCHER_PASSWORD: ${{ secrets.RANCHER_PASSWORD }}
```
Danil-Grigorev, line 10: We shouldn’t use any secrets in presubmit. This will work without it.
```suggestion
```

test/e2e/e2e_suite_test.go
```diff
@@ -88,6 +90,10 @@ var (
 
 	// chartPath is the path to the operator chart.
 	chartPath string
+
+	// isolatedMode instructs the test to run without ngrok and exposing the cluster to the internet. This setup will only work with CAPD
+	// or other providers that run in the same network as the bootstrap cluster.
+	isolatedMode bool
```
Danil-Grigorev, line 96: Ngrok ingress is provisioned in full if it has env set described in README. Isolated mode could be a default option, when the environment is not set.

Danil-Grigorev (approved):
I left some comments because this has some things to improve. Though it works, and having some minimal e2e on PR is desirable until we get it with prow in full.

.github/workflows/e2e-nightly-full.yaml
```diff
@@ -1,9 +1,8 @@
-name: Run e2e tests
+name: Run nightly e2e tests
 
 on:
-  push:
```
Danil-Grigorev, line 4: I think this job could run on PR merge and in nightly. No need to disable what works.

salasberryfin (approved):

test/e2e/e2e_suite_test.go
alexander-demicev: I'd like ngrok to be the default, when we run tests locally ngrok works for all setups. Using kind node IP will not work with docker desktop on mac :)

.github/workflows/e2e-every-pr.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+  e2e:
+    concurrency: test-e2e
```
Danil-Grigorev, line 14: This line has to be removed, we are no longer required to run only one e2e job at a time.

PR #139, opened 2023-09-21
feat: update chart to enable crs and topology features

### LINKED ISSUE #130: Update the chart to enable the CRS & Topology features
Update the chart so that when the specific 1.4.x version of core CAPI is being installed we enable the following features:

- CRS
- Topology

This will give a better experience for users when they get started.

salasberryfin:

This PR updates the chart so that  when it deploys the CoreProvider it also sets the environment variables for the features:

- CRS
- Topology

These features are currently set by [manually creating a secret](https://docs.rancher-turtles.com/docs/getting-started/install_turtles_operator) with the desired values for the environment variables `CLUSTER_TOPOLOGY` and `EXP_CLUSTER_RESOURCE_SET`. This secret will now be created as part of the chart installation and both features will default to `true`. These can otherwise be disabled passing the parameters to Helm:

`--set cluster-api-operator.cluster-api.secretConfig.name="user-secret-name"`
`--set cluster-api-operator.cluster-api.secretConfig.namespace="user-secret-namespace`

When both these parameters are provided, the secret is expected to exist in the namespace and the user must manage its contents, including (but not limited to) `CLUSTER_TOPOLOGY` and `EXP_CLUSTER_RESOURCE_SET`.

Fixes #130 

README is updated with the new values and a new issue is created to track `rancher-turtles-docs` updates rancher-sandbox/rancher-turtles-docs#27.

**Checklist**:

- [x] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/rancher-turtles/templates/core-provider.yaml
+++ b/charts/rancher-turtles/templates/core-provider.yaml
@@ -21,11 +21,12 @@ spec:
   additionalManifests:
     name: capi-additional-rbac-roles
     namespace: {{ index .Values "cluster-api-operator" "cluster-api" "core" "namespace" }}
-{{- if index .Values "cluster-api-operator" "cluster-api" "configSecret" "name" }}
+{{- if and (index .Values "cluster-api-operator" "cluster-api" "configSecret" "name") (index .Values "cluster-api-operator" "cluster-api" "configSecret" "namespace") }}
   secretName: {{ index .Values "cluster-api-operator" "cluster-api" "configSecret" "name" }}
-{{- if index .Values "cluster-api-operator" "cluster-api" "configSecret" "namespace" }}
   secretNamespace: {{ index .Values "cluster-api-operator" "cluster-api" "configSecret" "namespace" }}
-{{- end }}
+{{ else }}
+  secretName: {{ index .Values "cluster-api-operator" "cluster-api" "configSecret" "defaultName" }}
+  secretNamespace: {{ index .Values "cluster-api-operator" "cluster-api" "core" "namespace" }}
 {{- end }}
 {{- if or (index .Values "cluster-api-operator" "cluster-api" "core" "fetchConfig" "url") (index .Values "cluster-api-operator" "cluster-api" "core" "fetchConfig" "selector") }}
   fetchConfig:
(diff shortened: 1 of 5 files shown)


test/e2e/helpers_test.go
```diff
@@ -30,9 +30,6 @@ var (
 	//go:embed resources/testdata/fleet-capi-test.yaml
 	fleetCAPITestdata []byte
 
-	//go:embed resources/config/capi-providers-secret.yaml
-	capiProvidersSecret []byte
```
Danil-Grigorev, line 34: Why is this removed? This secret is a “patch” on top of existing resource as well, and is used for providing additional env depending on type of test. Like in e2e for azure PR, it adds all azure env secrets.

salasberryfin: It was removed just because I did not see it being used other than in the `Adding CAPI variables secret` section of [initRancherTurtles](https://github.com/rancher-sandbox/rancher-turtles/blob/ed9b78bb4cd64dab12dee4feb5dac19cab538ced/test/e2e/e2e_suite_test.go#L249), which is no longer required, but I have no problem in keeping it if it's needed.

charts/rancher-turtles/templates/core-provider.yaml
```diff
@@ -68,4 +64,17 @@ data:
       - apiGroups: ["rke-machine.cattle.io"]
         resources: ["*"]
         verbs: ["*"]
+---
```
Danil-Grigorev, line 67: What if a user would want to use his own secret, already created in the cluster?

test/e2e/helpers_test.go
Danil-Grigorev: For this to be no longer required a secret content should be fully customizable. Right now you can only set 2 feature flags.

Danil-Grigorev (changes requested):

charts/rancher-turtles/templates/core-provider.yaml
salasberryfin: I will add a flag to use an existing secret. The default behavior would be to create the pre-configured secret but the user can pass this parameter to avoid this.

test/e2e/helpers_test.go
salasberryfin: If we give the user the possibility to use their own secret, do we still need to let them customize the content of the "default" secret?

charts/rancher-turtles/templates/core-provider.yaml
Danil-Grigorev: The simplest way how to achieve the goal - to not create anything if name and namespace is not set.

salasberryfin: The initial idea was to use an opt-out strategy rather than opt-in, so the secret would be created by default and the values for these two features set to true. They would be able to explicitly turn them off.

test/e2e/helpers_test.go
Danil-Grigorev: This functionality used in e2e tests. There is a PR implementing test case for [azure](https://github.com/rancher-sandbox/rancher-turtles/pull/133), it was opened a couple days prior to that. This is where the secret is used and extended.

salasberryfin:
I would be more in favor of @alexander-demicev's proposal. I think this is a simpler solution and gives the user the freedom to manage their own secret if they want to do so, which I would say covers both point 2 and 3 in @Danil-Grigorev's list of use cases.

The deletion of the secret patching functionality was just an initial clean-up proposal which I'll undo. This could use a user-managed secret (current status) with any custom environment variables/feature flags.

Danil-Grigorev (approved):

richardcase (dismissed):

salasberryfin:
There was a condition missing in `core-provider.yaml` when setting `CoreProvider` https://github.com/rancher-sandbox/rancher-turtles/blob/d73fd08d01e71aa21ecdecfa23d0269f94cfc3d6/charts/rancher-turtles/templates/core-provider.yaml#L24 It should now be fixed

Danil-Grigorev (approved):

alexander-demicev (approved):

PR #190, opened 2023-10-04
🌱 Apply embedded capi feature flag after rancher install

Danil-Grigorev:
kind/chore

**What this PR does / why we need it**:
This PR converts all existing gitops scenarios to a rancher installation with post-install step applying required feature gates as per documented [installation process](https://docs.rancher-turtles.com/docs/getting-started/rancher#installing-rancher).

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #188 

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits into logical changes
- [ ] includes documentation
- [ ] adds unit tests
- [x] adds or updates e2e tests

PR #192, opened 2023-10-04
fix: use pre-install helm hook to prepare rancher for turtles

### LINKED ISSUE #185: Add option to delete old Rancher webhooks

As a operator
When i have an existing Rancher Manager Cluster
And i install Rancher Turtles 
And disable the  "embedded CAPI" feature in Rancher
I expect that i can apply CAPI cluster definitions to the cluster

### Detailed Description

Currently doing the above results in validating/mutating webhooks being left behind after the feature is disabled. If you try and apply a CAPI manifest you will get an error:

```
time="2023-10-03T14:18:36Z" level=error msg="error syncing 'cluster-fleet-local-local-1a3d67d0a899/clusters-clusters': handler bundle-deploy: failed to create  │
│ resource: Internal error occurred: failed calling webhook \"default.cluster.cluster.x-k8s.io\": failed to call webhook: Post \"https://capi-webhook-service.cat/ │
│ tle-provisioning-capi-system.svc:443/mutate-cluster-x-k8s-io-v1beta1-cluster?timeout=10s\": service \"capi-webhook-service\" not found, requeuing"
```

Disabling the feature doesn't remove the webhooks. To fix this manually you can run the following commands after disabling the feature:

```bash

kubectl delete mutatingwebhookconfigurations.admissionregistration.k8s.io mutating-webhook-configuration
kubectl delete validatingwebhookconfigurations.admissionregistration.k8s.io validating-webhook-configuration
```

We need to:

- [ ] Update our docs so people can run those commands
- [ ] Look at incorporating feature flag enabled functionality to automatically delete the webhooks (i.e. engineer a fix). Like we are doing in #178 for labels on the kubeconfig secrets
- [ ] Create a fix upstream in Rancher

_No response_

### Label(s) to be applied

/kind feature

salasberryfin:

Currently, installing Rancher Turtles in an existing Rancher Manager after disabling embedded CAPI, results in validating/mutating webhooks being left behind after the feature is disabled. If you try and apply a CAPI manifest you will get an error:

```
time="2023-10-03T14:18:36Z" level=error msg="error syncing 'cluster-fleet-local-local-1a3d67d0a899/clusters-clusters': handler bundle-deploy: failed to create  │
│ resource: Internal error occurred: failed calling webhook \"default.cluster.cluster.x-k8s.io\": failed to call webhook: Post \"https://capi-webhook-service.cat/ │
│ tle-provisioning-capi-system.svc:443/mutate-cluster-x-k8s-io-v1beta1-cluster?timeout=10s\": service \"capi-webhook-service\" not found, requeuing"
```

To avoid having to manually delete these webhooks and disable the feature flag, this PR adds a `pre-install` Helm hook that:
1. Disables embedded CAPI.
2. Deletes mutating/validating webhooks (creating a Service Account, Cluster Role and Cluster Role Binding).

This is not the fanciest of solutions but it allows to install Rancher Turtles just using `helm install` in an existing Rancher Manager. We can later change Rancher and remove this hook if we find a more robust option.

Fixes #185 

When this gets merged, we'll need to update the docs accordingly.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/test_chart.yaml
+++ b/.github/workflows/test_chart.yaml
@@ -8,6 +8,8 @@ env:
   MANIFEST_IMG: controller
   CONTROLLER_IMG: controller
   PULL_POLICY: Never
+  CERT_MANAGER_VERSION: v1.12.3
+  RANCHER_VERSION: v2.7.5
 
 jobs:
   lint-test:
@@ -54,9 +56,22 @@ jobs:
         uses: helm/kind-action@v1.8.0
         with:
           cluster_name: kind
+          node_image: kindest/node:v1.26.3
 
       - name: Add local docker image
         run: kind load docker-image ${{ env.MANIFEST_IMG }}:${{ env.TAG }}
+
+      - name: Add cert-manager chart repo
+        run: helm repo add jetstack https://charts.jetstack.io
+
+      - name: Add rancher chart repo
+        run: helm repo add rancher-stable https://releases.rancher.com/server-charts/stable
+
+      - name: Install cert-manager
+        run: helm install cert-manager jetstack/cert-manager --namespace cert-manager --create-namespace --version ${{ env.CERT_MANAGER_VERSION }} --set installCRDs=true --wait
+
+      - name: Install Rancher
+        run: helm install rancher rancher-stable/rancher --namespace cattle-system --create-namespace --set bootstrapPassword=rancheradmin --set replicas=1 --set hostname="e2e.dev.rancher" --set 'extraEnv[0].name=CATTLE_FEATURES' --set global.cattle.psp.enabled=false --version ${{ env.RANCHER_VERSION }} --wait
 
       - name: Run chart-testing (install)
         run: helm install rancher-turtles out/charts/rancher-turtles/ -n rancher-turtles-system --create-namespace --wait --set cluster-api-operator.cluster-api.enabled=false --set cluster-api-operator.enabled=false
--- a/charts/rancher-turtles/templates/pre-install-job.yaml
+++ b/charts/rancher-turtles/templates/pre-install-job.yaml
@@ -1,0 +1,99 @@
+{{- if index .Values "rancherTurtles" "features" "embedded-capi" "disabled" }}
+---
+apiVersion: management.cattle.io/v3
+kind: Feature
+metadata:
+  name: embedded-cluster-api
+  annotations:
+    "helm.sh/hook": pre-install
+    "helm.sh/hook-weight": "1"
+spec:
+  value: false
+{{- end }}
+{{- if index .Values "rancherTurtles" "features" "rancher-webhook" "cleanup" }}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: pre-install-job
+  namespace: rancher-turtles-system
+  annotations:
+    "helm.sh/hook": pre-install
+    "helm.sh/hook-weight": "1"
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: pre-install-job-delete-webhooks
+  annotations:
+    "helm.sh/hook": pre-install
+    "helm.sh/hook-weight": "1"
+rules:
+- apiGroups:
+  - admissionregistration.k8s.io
+  resources:
+  - validatingwebhookconfigurations
+  - mutatingwebhookconfigurations
+  verbs:
+  - delete
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: pre-install-job-webhook-cleanup
+  annotations:
+    "helm.sh/hook": pre-install
+    "helm.sh/hook-weight": "1"
+subjects:
+  - kind: ServiceAccount
+    name: pre-install-job
+    namespace: rancher-turtles-system
+roleRef:
+  kind: ClusterRole
+  name: pre-install-job-delete-webhooks
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: rancher-mutatingwebhook-cleanup
+  annotations:
+    "helm.sh/hook": pre-install
+    "helm.sh/hook-weight": "2"
+spec:
+  ttlSecondsAfterFinished: 300
+  template:
+    spec:
+      serviceAccountName: pre-install-job
+      containers:
+        - name: rancher-mutatingwebhook-cleanup
+          image: {{ index .Values "rancherTurtles" "features" "rancher-webhook" "kubectlImage" }}:{{ .Capabilities.KubeVersion }}
+          args:
+          - delete
+          - mutatingwebhookconfigurations.admissionregistration.k8s.io
+          - mutating-webhook-configuration
+          - --ignore-not-found=true
+      restartPolicy: Never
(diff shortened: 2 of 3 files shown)


Danil-Grigorev (approved):

salasberryfin:
We may need to install Rancher before Rancher Turtles in the E2E environment for the `Feature` kind to be available. Looks like this is triggering CI failures.

Danil-Grigorev (approved):

Danil-Grigorev:
[E2e](https://github.com/rancher-sandbox/rancher-turtles/pull/190) tests are passing with this change, so failure here could be ignored.

salasberryfin:
Thanks @Danil-Grigorev. Should we consider this PR superseded by yours #190 and close it?

Danil-Grigorev:
I don’t know. Can go both ways, you can cherry pick the commit #190 and the other one could be closed. Or this PR could be merged without e2e tests passing, and followed by #190 separately.

salasberryfin:
Let's merge this first and then #190 and keep them as independent PRs.

charts/rancher-turtles/templates/pre-install-job.yaml
```diff
@@ -0,0 +1,73 @@
+---
```
richardcase, line 1: Could we have a value that can be used to enable or disable this? Perhaps in the values.yaml we can have something like this:

```yaml
features:
  cleanup-rancher-webhooks: true
```

I can then add to the values.yaml in this section to enable/delete the labels update (#178)

charts/rancher-turtles/templates/pre-install-job.yaml
```diff
@@ -0,0 +1,73 @@
...
+    "helm.sh/hook": pre-install
+    "helm.sh/hook-weight": "2"
+spec:
+  ttlSecondsAfterFinished: 300
+  template:
+    spec:
+      serviceAccountName: pre-install-job
+      containers:
+        - name: rancher-webhook-cleanup
+          image: bitnami/kubectl
```
richardcase, line 67: Rancher does something similar. Do you know if it uses the same image?

richardcase, line 67: Also, i think we'd want to have the image configurable in the values.yaml for air-gapped scenarios.

richardcase (commented):
In my head i was thinking of something actually in the Rancher Turtles binary but this is an interesting approach.

charts/rancher-turtles/templates/pre-install-job.yaml
salasberryfin: I was investigating a bit and Rancher uses this https://github.com/rancher/kubectl utility that directly runs the command you pass in the container and terminates it rather than having a bash session that you can run `kubectl` commands in. The only issue with this is that you must specify the `kubectl` version as tag so having custom images won't be really possible. I guess this is fine considering it is a Rancher-owned image but it may be worth a look.

salasberryfin:
I added flags for disabling embedded capi and deleting rancher webhooks:
```bash
rancherTurtles:
  image: controller
  tag: v0.0.0
  imagePullPolicy: Never
  managerArguments: {}
  imagePullSecrets: []
  features:
    disable-embedded-capi:
      enabled: true
    cleanup-rancher-webhook:
      enabled: true
      image: rancher/kubectl
```
Also, changed the deleting webhook `Job` to use `rancher/kubectl` image. This has to now be separated into two different jobs: one for each webhook deletion (mutating/validating).

richardcase (approved):
Looks good, thanks for adding the flags in the values.

Although there is a chart linrt failure.

charts/rancher-turtles/values.yaml
```diff
@@ -5,6 +5,12 @@ rancherTurtles:
   namespace: rancher-turtles-system
   managerArguments: {}
   imagePullSecrets: []
+  features:
+    disable-embedded-capi:
+      enabled: true
```
Danil-Grigorev, line 10:
```suggestion
    embedded-capi:
      disabled: true
```

charts/rancher-turtles/values.yaml
```diff
@@ -5,6 +5,12 @@ rancherTurtles:
...
+    cleanup-rancher-webhook:
+      enabled: true
```
Danil-Grigorev, line 12:
```suggestion
   rancher-webhook:
      cleanup: true
```

.github/workflows/test_chart.yaml
```diff
@@ -54,10 +56,23 @@ jobs:
...
+        run: helm repo add rancher-stable https://releases.rancher.com/server-charts/stable
+
+      - name: Install cert-manager
+        run: helm install cert-manager jetstack/cert-manager --namespace cert-manager --create-namespace --version ${{ env.CERT_MANAGER_VERSION }} --set installCRDs=true --wait
+
+      - name: Install Rancher
+        run: helm install rancher rancher-stable/rancher --namespace cattle-system --create-namespace --set bootstrapPassword=rancheradmin --set replicas=1 --set hostname="e2e.dev.rancher" --set 'extraEnv[0].name=CATTLE_FEATURES' --set global.cattle.psp.enabled=false --version ${{ env.RANCHER_VERSION }} --wait
+
       - name: Run chart-testing (install)
         run: helm install rancher-turtles out/charts/rancher-turtles/ -n rancher-turtles-system --create-namespace --wait --set cluster-api-operator.cluster-api.enabled=false --set cluster-api-operator.enabled=false
```
Danil-Grigorev, line 77:
```suggestion
        run: helm install rancher-turtles out/charts/rancher-turtles/ -n rancher-turtles-system --create-namespace --wait --set cluster-api-operator.cluster-api.enabled=false --set cluster-api-operator.enabled=false --set features. disable-embedded-capi.enabled=false --set features.cleanup-rancher-webhook.enabled=false
```
As this workflow represents the default flow of installing rancher-turtles in a fresh cluster, we might not need to bring rancher with it. Just making sure the chart is installable.

Danil-Grigorev (dismissed):
Some minor suggestions, but LGTM as is.

.github/workflows/test_chart.yaml
```diff
@@ -54,10 +56,23 @@ jobs:
...
+
+      - name: Add rancher chart repo
+        run: helm repo add rancher-stable https://releases.rancher.com/server-charts/stable
+
+      - name: Install cert-manager
+        run: helm install cert-manager jetstack/cert-manager --namespace cert-manager --create-namespace --version ${{ env.CERT_MANAGER_VERSION }} --set installCRDs=true --wait
+
+      - name: Install Rancher
+        run: helm install rancher rancher-stable/rancher --namespace cattle-system --create-namespace --set bootstrapPassword=rancheradmin --set replicas=1 --set hostname="e2e.dev.rancher" --set 'extraEnv[0].name=CATTLE_FEATURES' --set global.cattle.psp.enabled=false --version ${{ env.RANCHER_VERSION }} --wait
+
```
Danil-Grigorev, line 75:
```suggestion
```

In addition to https://github.com/rancher-sandbox/rancher-turtles/pull/192/files#r1348559835

.github/workflows/test_chart.yaml
salasberryfin: I understand your point but I would say, considering these flags are enable by default, we should test them as well to validate the chart.

richardcase (approved):

PR #218, opened 2023-10-19
chore: encode gh actions outputs

### LINKED ISSUE #211: [Release] Prod signing failing
Whilst trying to release v0.2.0 we encountered an issue with the **prod-sign** job:

```
Error: requires at least 1 arg(s), only received 0
main.go:74: error during command execution: requires at least 1 arg(s), only received 0
Error: Process completed with exit code 1.
```

We tried a number of things related the `REGISTRY_ENDPOINT` but this didn't work.

salasberryfin:

GitHub detects outputs containing substrings that are part of secrets as insecure and blocks us from sharing them with downstream jobs. This is a proposal to fix this issue by double encoding outputs so that they are obfuscated. Then, these outputs are decoded in the signing/provenance jobs.

These outputs are not actually sensitive data but just the image location and, worst case scenario (user decodes the string output), this data is public anyway.

Fixes #211 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -4,8 +4,6 @@ on:
   push:
     tags:
     - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
-env:
-  PROD_REGISTRY: registry.rancher.com/rancher
 
 permissions:
   contents: write # Allow to create a release.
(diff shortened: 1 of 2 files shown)


Danil-Grigorev:
I think the better way to handle this is by using existing github tooling: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#workflow

salasberryfin:
> I think the better way to handle this is by using existing github tooling: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#workflow

The problem with this approach is that it uses an external secret store.

furkatgofurov7 (approved):

richardcase (approved):

PR #270, opened 2023-11-19
chore: send alarm message when release test is cancelled

### LINKED ISSUE #269: Notify when release test is cancelled
### Description

The release testing nightly workflow is configured to send alarm messages to Slack only when a failure occurs in any of the workflow's jobs. With the addition of re-usable workflows, it looks like one of them failing causes the whole workflow to have a `cancelled()` status and not a `failure()`. The cancelled executions should be notified as a failed tests.

salasberryfin:

The first few executions of the nightly release test are cancelled after one of the re-used workflows fails which causes a cancelled status and not a failure. The notification job in the workflow is only configured to run if `failure()` and this PR adds the `cancelled()` state to the condition to get notified when this occurs.

Fixes #269 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/nightly-test-release.yaml
+++ b/.github/workflows/nightly-test-release.yaml
@@ -110,7 +110,7 @@ jobs:
   notify-failure:
     name: Notify failure in Slack
     needs: [release]
-    if: failure()
+    if: ${{ failure() || cancelled() }}
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


furkatgofurov7 (approved):
Don't recall seeing `failure` reports on slack, but maybe I missed it. Changes LGTM

richardcase (approved):

PR #309, opened 2023-12-12
test: allow limiting which suites are run

richardcase:

Small change to allow you to run a specific suite when running tests and splits up the self-hosted runner test into a step per e2e suite.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/e2e-long-test.yaml
+++ b/.github/workflows/e2e-long-test.yaml
@@ -17,7 +17,8 @@ env:
   AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
 
 jobs:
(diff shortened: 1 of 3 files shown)


furkatgofurov7 (approved):
Maybe we could think of using a reusable workflows in the future to avoid some repetitions, overall looks good to me

richardcase:
> Maybe we could think of using a reusable workflows in the future to avoid some repetitions, overall looks good to me

💯 When we get to the bottom of why the e2e are failing then we can do this for sure 👍

Danil-Grigorev (approved):

PR #335, opened 2024-01-05
🌱 Implement full cluster artifact gathering for both clusters after each test run

### LINKED ISSUE #333: Collect details about failed e2e test clusters in CI
Our e2e tests do not collect necessary debugging information, required to determine conditions, when the jobs has failed.
We need a tool comparable to must-gather in order to do so.

Danil-Grigorev:
kind/chore

Fixes #333 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/test/e2e/suites/v2prov/v2prov_test.go
+++ b/test/e2e/suites/v2prov/v2prov_test.go
@@ -22,6 +22,7 @@ package v2prov
 import (
 	"fmt"
 	"os"
+	"path"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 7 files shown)


Danil-Grigorev:
@richardcase PTAL

test/e2e/suites/v2prov/v2prov_test.go
```diff
@@ -155,6 +164,13 @@ var _ = Describe("[v2prov] [Azure] Creating a cluster with v2prov should still w
 		}, rancherConnectRes)
 		Expect(rancherConnectRes.Error).NotTo(HaveOccurred(), "Failed getting nodes with Rancher Kubeconfig")
 		Expect(rancherConnectRes.ExitCode).To(Equal(0), "Getting nodes return non-zero exit code")
+	})
+
+	AfterEach(func() {
+		err := testenv.CollectArtifacts(ctx, rancherKubeconfig.TempFilePath, path.Join(flagVals.ArtifactFolder, setupClusterResult.BootstrapClusterProxy.GetName(), clusterName))
+		if err != nil {
+			fmt.Printf("Failed to artifacts for the child cluster: %v\n", err)
```
salasberryfin, line 172:
```suggestion
			fmt.Printf("Failed to collect artifacts for the child cluster: %v\n", err)
```

test/framework/rancher_helpers.go
```diff
@@ -101,6 +107,101 @@ func RancherGetClusterKubeconfig(ctx context.Context, input RancherGetClusterKub
 	result.TempFilePath = tempFile.Name()
 }
 
+// RancherGetOriginalKubeconfig will get the unmodified Kubeconfig for a cluster from Rancher.
+func RancherGetOriginalKubeconfig(ctx context.Context, input RancherGetClusterKubeconfigInput, result *RancherGetClusterKubeconfigResult) {
```
salasberryfin, line 111: Would it make sense to have this and `RancherGetClusterKubeconfig` into a single function? Looks like the differences in the logic are rather small. Maybe we can have an extra field in `RancherGetClusterKubeconfigInput` to set whether we're getting one kubeconfig or the other? Just wondering if this would be viable and so we minimize code repetition.

salasberryfin (commented):
LGTM. Just a couple of nitpicking comments.

test/framework/rancher_helpers.go
salasberryfin: Thanks, that makes sense.

Danil-Grigorev: Generally speaking the separate kubeconfig is required only for one scenario, when a rancher API server is not available. So it is not required to use the second method, if this failure scenario is not assumed.

salasberryfin (approved):

richardcase (approved):

PR #342, opened 2024-01-11
✨Implement initial rancher credentials mapping

### LINKED ISSUE #5: Implement credentials mapping from Rancher to CAPI
**Story**

As a rancher operator,
I want to use Rancher Manager cloud credentials with CAPI,
So that I don't have to enter credentials twice in the UI.

**Description**
We would like to prove out credentials mapping from Rancher cloud credentials to CAPI credentials.

We'd need to watch for AWS cloud-credentials in the Rancher manager cluster and then create the relevant CAPI credential type for the provider.

For example, for AWS credentials we would create an instance of `AWSClusterStaticIdentity`.

**Unanswered Questions**
Do we do this ahead of time? Or at the point of creating the cluster?

Danil-Grigorev:
This implements basic CAPI credentials mapping for Rancher cloud credential secrets. Currently for Azure, AWS, GCP, DigitalOcean and VSphere.

TODO:
- [x] report conditions on missing required keys in the source rancher secret
- [ ] e2e tests
- [ ] Azure tenantID handling question - is reporting this key as missing enough?

issue: https://github.com/rancher-sandbox/rancher-turtles/issues/346

Fixes #5 

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests


main.go
```diff
@@ -132,18 +134,26 @@ func main() {
...
+		Client: client.Options{
+			Cache: &client.CacheOptions{
+				DisableFor: []client.Object{
+					&corev1.ConfigMap{},
+					&corev1.Secret{},
+				},
+			},
+		},
+		Cache: cache.Options{
+			SyncPeriod: &syncPeriod,
```
alexander-demicev, line 156: Why is the sync period moved to a new location? is related to the controller-runtime bump?

internal/sync/provider_sync.go
```diff
@@ -26,17 +27,27 @@ import (
...
+	Destination api.Provider
 }
 
 // NewProviderSync creates a new mirror object.
 func NewProviderSync(cl client.Client, capiProvider *turtlesv1.CAPIProvider) Sync {
+	template := ProviderSync{}.Template(capiProvider)
+
+	destination, ok := template.(api.Provider)
+	if !ok {
+		return nil
```
alexander-demicev, line 45: can we add a log entry here?

richardcase (dismissed):
Great work @Danil-Grigorev 

It would be good to create an issue to follow up on the e2e and Azure tenant id question.

internal/sync/secret_mapper_sync.go
```diff
@@ -0,0 +1,314 @@
...
+	*SecretSync
+	RancherSecret *corev1.Secret
+}
+
+// NewSecretMapperSync creates a new secret mapper object sync.
+func NewSecretMapperSync(cl client.Client, capiProvider *turtlesv1.CAPIProvider) Sync {
+	if capiProvider.Spec.Credentials == nil ||
+		capiProvider.Spec.Credentials.RancherCloudCredential == "" ||
+		knownProviderRequirements[capiProvider.Spec.Name] == nil {
+		return nil
```
alexander-demicev, line 196: same as above, some debug logs maybe? Just to avoid cases of not knowing when nil silently returned

alexander-demicev (commented):
Looks good, just a couple of small comments

main.go
richardcase: It is related to controller-runtime

internal/sync/provider_sync.go
Danil-Grigorev: This will never happen, annoyingly linter thinks that the `ok` value should be checked here. And the initialization is covered with tests, so this is proven.

alexander-demicev:
How are we going to approach cases when authentication is done using CRDs like in CAPZ?https://capz.sigs.k8s.io/topics/multitenancy#deprecated-identity-types ```The capability to set credentials using environment variables has been removed, the required approach is to use `AzureClusterIdentity` as seen in the [identities](identities.md) page.```

internal/sync/secret_mapper_sync.go
Danil-Grigorev: The nil logic is handled as a normal case in sync interface. This just skips the component entirely for the reconciled CAPIProvider resource. This may generate many lines of unnecessary information, not sure about that. Maybe log.V(9) or something as a level?

Danil-Grigorev:
> How are we going to approach cases when authentication is done using CRDs like in CAPZ?https://capz.sigs.k8s.io/topics/multitenancy#deprecated-identity-types `` The capability to set credentials using environment variables has been removed, the required approach is to use `AzureClusterIdentity` as seen in the [identities](identities.md) page. ``

This information can be provided as a variable in the cluster class definition. It’s another topic from rancher credentials mapping.

richardcase:
> How are we going to approach cases when authentication is done using CRDs like in CAPZ?https://capz.sigs.k8s.io/topics/multitenancy#deprecated-identity-types `` The capability to set credentials using environment variables has been removed, the required approach is to use `AzureClusterIdentity` as seen in the [identities](identities.md) page. ``

The current approach is only for the default provider credentials. CAPZ doesn't support default credentials, only credentials when you create a cluster. For providers that support other credential mechanisms when enabling a provider, like CAPA, we will extend `CAPIProvider`.

For cluster creation, if someone wants to use a credential from a Rancher cloud credential we will automatically create the CAPI representations from the rancher cloud credential...the actual mechanism is TBD but could be via a webhook on creation.

internal/sync/secret_mapper_sync.go
Danil-Grigorev: Fixed

Danil-Grigorev:
Bumped helm chart as well as per @alexander-demicev request and added some debug logs. PTAL @alexander-demicev @richardcase, should be good now.

alexander-demicev (approved):
thanks for working on this

Danil-Grigorev:
@richardcase Added a separate issue to track e2e coverage: https://github.com/rancher-sandbox/rancher-turtles/issues/346

richardcase (approved):

PR #349, opened 2024-01-17
fix: installation on eks

### LINKED ISSUE #347: Installing Rancher Turtles on EKS fails

I setup a Rancher Manager cluster on EKS then installed Rancher Turtles to the local cluster via the app marketplace.

The job that performs clean up tasks fails with an `ErrImagePull` error. Its trying to pull `docker.io/rancher/kubectl:v1.26.12-eks-5e0fdde` which doesn't exist.

I expect Rancher Turtles to install without issues.

### How to reproduce it?

1. Setup a Rancher Manager cluster on EKS using [these instructions](https://ranchermanager.docs.rancher.com/getting-started/installation-and-upgrade/install-upgrade-on-a-kubernetes-cluster/rancher-on-amazon-eks)
2. Install Rancher Turtles via the app store to the **local** cluster

### Rancher Turtles version

0.4.0

_No response_

### Label(s) to be applied

/kind bug

richardcase:

There is an issue installing Rancher Turtles on EKS using the chart as the kubernetes version in EKS doesn't follow semver. For example, EKS will return a version like `v1.26.12-eks-5e0fdde` and this means that we then try and pull the `docker.io/rancher/kubectl:v1.26.12-eks-5e0fdde` image when performing the cleanup tasks.

The function to strip the suffix from the kubernetes version has been updated to handle EKS, AKS and GKE.

Fixes #347 

**Checklist**:

### DIFF
--- a/charts/rancher-turtles/templates/_helpers.tpl
+++ b/charts/rancher-turtles/templates/_helpers.tpl
@@ -1,9 +1,22 @@
 {{/*
     This removes the part after the + in the kubernetes version string.
     v1.27.4+k3s1 -> v1.27.4
+    v1.26.12-eks-5e0fdde -> v1.26.12
+    v1.26.12-gke.1 -> v1.26.12
     v1.28.0 -> v1.28.0
 */}}
 {{- define "strippedKubeVersion" -}}
-{{- $parts := split "+" .Capabilities.KubeVersion.Version -}}
-{{- print $parts._0 -}}
+    {{- if (.Capabilities.KubeVersion.Version | contains "-eks-") -}}
+        {{- $parts := split "-eks-" .Capabilities.KubeVersion.Version -}}
+        {{- print $parts._0 -}}
+    {{- else if (.Capabilities.KubeVersion.Version | contains "-gke.") -}}
+        {{- $parts := split "-gke." .Capabilities.KubeVersion.Version -}}
+        {{- print $parts._0 -}}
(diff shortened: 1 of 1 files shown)


charts/rancher-turtles/templates/_helpers.tpl
```diff
@@ -1,9 +1,22 @@
...
+        {{- $parts := split "-eks-" .Capabilities.KubeVersion.Version -}}
+        {{- print $parts._0 -}}
+    {{- else if (.Capabilities.KubeVersion.Version | contains "-gke.") -}}
+        {{- $parts := split "-gke." .Capabilities.KubeVersion.Version -}}
+        {{- print $parts._0 -}}
+    {{- else if (.Capabilities.KubeVersion.Version | contains "-aks") -}}
+        {{- $parts := split "-aks" .Capabilities.KubeVersion.Version -}}
+        {{- print $parts._0 -}}
+    {{- else -}}
+        {{- $parts := split "+" .Capabilities.KubeVersion.Version -}}
```
Danil-Grigorev, line 19: {{- $parts := split "-" .Capabilities.KubeVersion.Version -}}

I think we can just strip the suffix here generically? WDYT? But just a nit, lgtm

Danil-Grigorev (approved):

charts/rancher-turtles/templates/_helpers.tpl
richardcase: I did think about just stripping from `-` so went with something specific to each case.

Happy to change it.

PR #357, opened 2024-01-25
docs: managementv3 cluster support strategy

### LINKED ISSUE #361: ADR: support `management.cattle.io/v3` cluster import
### Description

The current behavior of Rancher Turtles is to create a `provisioning.cattle.io/v1` cluster when a CAPI cluster is configured to be imported. In light of future changes in Rancher, Turtles must support an alternative option to create `management.cattle.io/v3` resources when importing clusters.

- [ ] Create a proposal that describes how this alternative import strategy needs to be implemented.

salasberryfin:

This ADR is a proposal on how support for `management.cattle.io/v3` clusters is implemented, based on investigation during the development process. 

**The main point of concern here is how we manage cluster naming**. Since `management.cattle.io/v3` clusters are required to have a name that satisfies the string `c-` followed by exactly five characters, we cannot use the display name (the one that appears in Rancher dashboard) as we do with `provisioning.cattle.io/v1` clusters (where name and display name are the same).

Creating a randomized 5-character suffix presents the challenge of not being able to track the resource while reconciling like we do in the existing `provisioning.cattle.io/v1` import controller, since we fetch the cluster on `Name` and `Namespace` (refer to [here](https://github.com/rancher-sandbox/rancher-turtles/blob/8109fda7fe2d77f549ea03fcb0c9b2095e5dd55a/internal/controllers/import_controller.go#L199)).

~A reasonable alternative is to compute the SHA256 hash of the display name and then truncate the result to fit in the 5-character prefix of the name. This would make it easy and cheap to deterministically get the resource name based on the display name. The problem with this approach is that we could potentially encounter collisions if the number of clusters is large.~

We will be tracking the cluster based on two new labels that will be applied to all `management.cattle.io/v3` Rancher clusters that uniquely reference the CAPI cluster that owns the resource. Using this approach, we can leverage the built-in functionality of Kubernetes by passing the field `GenerateName` with the prefix `c-` so that it randomly generates a 5 character suffix that satisfies the regular expression, and retries if a name collision occurs.

Fixes #361 

**Checklist**:

- [x] includes documentation
- [ ] adds unit tests

### DIFF
--- a/docs/adr/0008-managementv3-clusters-support.md
+++ b/docs/adr/0008-managementv3-clusters-support.md
@@ -1,0 +1,51 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [8. Management V3 clusters support](#8-management-v3-clusters-support)
+  - [Context](#context)
+  - [Decision](#decision)
+  - [Consequences](#consequences)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# 8. Management V3 clusters support
+
+* Status: proposed
+* Date: 2024-01-25
+* Authors: @salasberryfin
+* Deciders: @richardcase @alexander-demicev @furkatgofurov7 @Danil-Grigorev @mjura
+
+## Context
+
+As an operator, I'd like to have an alternative import strategy that will generate a `management.cattle.io/v3` cluster instead of a `provisioning.cattle.io/v1` cluster. This is to ensure that we can support this scenario in light of future changes to Rancher.
+
+## Decision
+
+The current import controller logic creates a `provisioning.cattle.io/v1` cluster when a new CAPI cluster is detected. To support a new import strategy, based on a new feature flag, Rancher Turtles will either start the existing `provisioning.cattle.io` controller or a new one for `management.cattle.io`.
+
+Adding support for this resource presents some challenges:
+- Owner Reference cannot be set: `management.cattle.io/v3` clusters are cluster-scoped while CAPI clusters are namespaced.
+- Display name and resource name are different.
+- Cluster registration token exists in a namespace with the same name as the `management.cattle.io/v3` cluster.
+
+Of all three, the naming strategy is the main concern. The `GenerateName` field allows us to set a prefix `c-` and automatically assigns a random name that satisfies the regular expression.
+
+Using this random name approach, we cannot rely on `Name` and `Namespace` to be able to control the status of the cluster during reconciliation, as we do in the existing import controller. After validation, the decision is that we set labels on the Rancher clusters that make it possible to link them back to the underlying CAPI cluster. The labels are:
+
+```
+"cluster-api.cattle.io/capi-cluster-owner": name of the CAPI cluster.
+"cluster-api.cattle.io/capi-cluster-owner-ns": namespace of the CAPI cluster.
+```
+
+These two labels allow us to uniquely identify each Rancher cluster, since only one CAPI cluster with a given name can exist in a given namespace. During the reconciliation process, we apply a `MatchingLabels` filter to strictly retrieve only the single cluster that contains the labels that associate it with the CAPI cluster.
+
+From an end-user perspective:
+
+* The default behavior remains unchanged. If no feature flag is set, the current `provisioning.cattle.io` import strategy will be used.
+* User provided name will be used in Rancher dashboard while the underlying `management.cattle.io/v3` will have a name that satisfies Rancher's naming requirement: `c-` followed by 5 randomly-generated characters.
+
+## Consequences
+
+- Kubernetes' native garbage collector using the owner reference chain is not a viable option due to the namespaced vs global scoped conflict. This means that there needs to be a custom logic to manage deletion of resources.
+- Cluster name will be used to access cluster registration token namespace.
+- Display name and cluster name will be different.


docs/adr/0008-managementv3-clusters-support.md
```diff
@@ -0,0 +1,42 @@
...
+Of all three, the naming strategy is the main concern. To be able to control the status of the cluster during reconciliation, there must be a deterministic name that identifies the resource. The decision is that we can create a hash from the cluster's display name and truncate it to the required five characters. This operation has low cost and a sufficiently randomized result yet it is deterministic.
+
+From an end-user perspective:
+
+* The default behavior remains unchanged. If no feature flag is set, the current `provisioning.cattle.io` import strategy will be used.
+* User provided name will be used in Rancher dashboard while the underlying `management.cattle.io/v3` wil have a name that satisfies Rancher's naming requirement: `c-` followed by 5 semi-random characters.
+
+## Consequences
+
+- Kubernetes' native garbage collector using the owner reference chain is not a viable option due to the namespaced vs global scoped conflict. This means that there needs to be a custom logic to manage deletion of resources.
```
Danil-Grigorev, line 40: Ownership references can't be set from the namespace scoped object, but it can be set from cluster scoped. This proposal has larger implications on managed / non-managed scenario. We still need to set ownerships to CAPI cluster from V3 cluster for managed, and don't set for non-managed.

docs/adr/0008-managementv3-clusters-support.md
```diff
@@ -0,0 +1,42 @@
...
+
+As an operator, I'd like to have an alternative import strategy that will generate a `management.cattle.io/v3` cluster instead of a `provisioning.cattle.io/v1` cluster. This is to ensure that we can support this scenario in light of future changes to Rancher.
+
+## Decision
+
+The current import controller logic creates a `provisioning.cattle.io/v1` cluster when a new CAPI cluster is detected. To support a new import strategy, based on a new feature flag, Rancher Turtles will either start the existing `provisioning.cattle.io` controller or a new one for `management.cattle.io`.
+
+Adding support for this resource presents some challenges:
+- Owner Reference cannot be set: `management.cattle.io/v3` clusters are cluster-scoped while CAPI clusters are namespaced.
+- Display name and resource name are different: the resource name must satisfy the regex `^(c-[a-z0-9]{5}|local)$`.
```
Danil-Grigorev, line 28: If a conflict occurs, name can be randomly generated again until the cluster is applied. Using only hash does not allow this.

docs/adr/0008-managementv3-clusters-support.md
```diff
@@ -0,0 +1,42 @@
...
+From an end-user perspective:
+
+* The default behavior remains unchanged. If no feature flag is set, the current `provisioning.cattle.io` import strategy will be used.
+* User provided name will be used in Rancher dashboard while the underlying `management.cattle.io/v3` wil have a name that satisfies Rancher's naming requirement: `c-` followed by 5 semi-random characters.
+
+## Consequences
+
+- Kubernetes' native garbage collector using the owner reference chain is not a viable option due to the namespaced vs global scoped conflict. This means that there needs to be a custom logic to manage deletion of resources.
+- Cluster name will be used to access cluster registration token.
+- Display name and cluster name will be different, and cluster name will need to satisfy the regular expression `^(c-[a-z0-9]{5}|local)$`.
```
Danil-Grigorev, line 42: Local is the name of self-import for the rancher cluster. We can't use it.

docs/adr/0008-managementv3-clusters-support.md
salasberryfin: If we randomly generate the name, we know that a conflict occurs when creating the cluster resource [here](https://github.com/rancher-sandbox/rancher-turtles/blob/8109fda7fe2d77f549ea03fcb0c9b2095e5dd55a/internal/controllers/import_controller.go#L234), but then, how can we actually fetch the correct cluster resource [here](https://github.com/rancher-sandbox/rancher-turtles/blob/8109fda7fe2d77f549ea03fcb0c9b2095e5dd55a/internal/controllers/import_controller.go#L199) if that was a random string?

docs/adr/0008-managementv3-clusters-support.md
salasberryfin: This is just the exact regex copied from Rancher. Didn't mean that `local` is a valid option. I'll update it to avoid misunderstandings.

docs/adr/0008-managementv3-clusters-support.md
Danil-Grigorev: When an owner reference is set on the CAPI cluster, we can track it from there. Otherwise the cluster v3 needs to be labeled with the CAPI cluster name to allow lookup.

docs/adr/0008-managementv3-clusters-support.md
salasberryfin: Based on the note on this document https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/, owner/dependent must either be both in the same namespace or both cluster-scoped.

Danil-Grigorev: The doc is incorrect in that assumption. Only namespaced resource can’t set cross-namespace owner reference, not the cluster scoped.

docs/adr/0008-managementv3-clusters-support.md
richardcase: When we create the `v3.Cluster` we can just use the builtin functionality of Kubernetes to ensure the resource has a unique name. Using `GenerateName: c-` and not setting `Name` when we create will be enough.

salasberryfin:
I updated the ADR document and the PR description based on the feedback. Thanks @Danil-Grigorev @richardcase.

docs/adr/0008-managementv3-clusters-support.md
```diff
@@ -0,0 +1,49 @@
...
+## Context
+
+As an operator, I'd like to have an alternative import strategy that will generate a `management.cattle.io/v3` cluster instead of a `provisioning.cattle.io/v1` cluster. This is to ensure that we can support this scenario in light of future changes to Rancher.
+
+## Decision
+
+The current import controller logic creates a `provisioning.cattle.io/v1` cluster when a new CAPI cluster is detected. To support a new import strategy, based on a new feature flag, Rancher Turtles will either start the existing `provisioning.cattle.io` controller or a new one for `management.cattle.io`.
+
+Adding support for this resource presents some challenges:
+- Owner Reference cannot be set: `management.cattle.io/v3` clusters are cluster-scoped while CAPI clusters are namespaced.
```
Danil-Grigorev, line 27: We can still set and ownership reference on the CAPI cluster, to ensure the deletion is followed one way at least.

salasberryfin: If I understand correctly, you mean setting the Rancher cluster as the owner and the CAPI cluster as the dependent object. If that's the case, wouldn't it conflict with the way we manage un-importing clusters? A user may want to remove a cluster from Rancher but keep the underlying CAPI cluster. I'm afraid setting owner references this way would break this functionality.

Danil-Grigorev: No, it will not. When creating a v3 cluster for unmanaged scenario we should follow  https://kubernetes.io/docs/tasks/administer-cluster/use-cascading-deletion/#set-orphan-deletion-policy, and apply the `orphan` finalizer to the v3 cluster object. This way, when a regular delete is issued to v3 cluster, this will not follow with deletion of dependents, and will simply remove the ownership reference.

For managed scenario the finalizer can be omitted. If we go this route, we might not need labels too, as ownership reference contains the name of the v3 object to lookup.

docs/adr/0008-managementv3-clusters-support.md
```diff
@@ -0,0 +1,49 @@
...
+- Display name and resource name are different: the resource name must satisfy the regex `^(c-[a-z0-9]{5})$`.
```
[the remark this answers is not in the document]
salasberryfin: The regex is set by Rancher [here](https://github.com/rancher/rancher/blob/676c3eb3dfa5a5d7d359315f942b630f8d3bf479/pkg/controllers/provisioningv2/cluster/controller.go#L50) and only clusters that satisfy it will be shown in the dashboard. I'd say it is relevant to mention its current state as a reason why the decision was made. I think that it would be best that, if it were to change in the future, we create a new document or simply update the code accordingly.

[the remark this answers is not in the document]
Danil-Grigorev: Rancher has a similar [issue](https://github.com/rancher/rancher/issues/40852) for the regex. As I said, it’s enough to say that we use `generateName: c-` for cluster creation. This will cover it, and we won’t propagate the problem further in a design document. It’s not a good practice to rely on internal functionality prone to change, and mention it in a design documents for the same reason.

docs/adr/0008-managementv3-clusters-support.md
Danil-Grigorev: We’ve validated use of the ownership references in this scenario, unfortunately most of the CLI clients specify the default `background` policy during deletion, overriding the `orphan` policy on the resource, and this irreversibly causes capi cluster deletion. It’s better to use label matching, as it’s currently being suggested.

docs/adr/0008-managementv3-clusters-support.md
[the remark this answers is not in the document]
salasberryfin: References to the current regular expression are now removed from the document.

Danil-Grigorev (approved):

richardcase (approved):

furkatgofurov7 (approved):

salasberryfin:
Thanks for the comments and reviews folks.

PR #362, opened 2024-01-29
✨ Support creating management v3 cluster

### LINKED ISSUE #298: Add ability to generate v3.Cluster
Add an alternative import strategy to Rancher Turtles that will generate a management.v3.Cluster instead of a provisioning.v1.Cluster.

This is to ensure that we can support this scenario in light of future changes to Rancher.

A small amount of investigation was done: https://github.com/rancher/highlander/issues/42

salasberryfin:

This PR adds support for representing imported CAPI clusters as `management.cattle.io/v3` clusters instead of the current `provisioning.cattle.io/v1` resource.

[ADR 8](https://github.com/rancher-sandbox/rancher-turtles/blob/main/docs/adr/0008-managementv3-clusters-support.md) contains the rationale behind choosing this implementation approach and the discussions on different alternatives before agreeing to a decision.

This includes a new import controller specific for the new type of resource, which is only enabled if using a feature flag, so the default behavior of Rancher Turtles is to continue with the current import strategy. Some of the functions that are used by both controllers have been centralized so that we have less code to maintain. Still, some of the controller code is very similar to one another but re-utilizing it would probably increase complexity and reduce readability.

An E2E specification equivalent to the existing `import gitops` has been created for when Rancher Turtles is executed with this feature flag enabled.

There are definitely improvements to the way the code is structured. Especially in terms of increasing the re-usability, for both controller and E2E, but my opinion is that this PR is large enough to also add a larger refactor. Probably it would be better if we can have a couple of follow-ups to simplify the code where possible. But this is obviously a team decision. Additionally, there will be a discussion on whether we should move completely to this new controller and deprecate `provisioning.cattle.io`, so it may not make much sense to invest time in increasing re-usable components at this point. 

Fixes #298 

~**The DeletionTimestamp seems to not be set when the Rancher Cluster is deleted.** The logic is the same that we use for other controller, though. I would appreciate any feedback on this. I'll leave a comment below.~

The new feature flag is added to the docs [here](https://github.com/rancher-sandbox/rancher-turtles-docs/pull/59).

**Checklist**:

- [x] includes documentation
- [x] adds unit tests


internal/controllers/import_controller_v3.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+
+	if !capiCluster.ObjectMeta.DeletionTimestamp.IsZero() {
+		err := r.deleteDependentRancherCluster(ctx, capiCluster)
+		if err != nil {
+			return ctrl.Result{Requeue: true}, fmt.Errorf("error deleting associated managementv3.Cluster resources: %w", err)
+		}
+	}
+
+	if !rancherCluster.ObjectMeta.DeletionTimestamp.IsZero() {
```
salasberryfin, line 218: I've seen this failing but, looking at other projects, it looks like this is the proper way to handle the resource's deletion. Why would it not set a `DeletionTimestamp`? This could break the "un-import" strategy of Rancher Turtles, since CAPI cluster wouldn't be annotated and it would try to reconcile the Rancher Cluster again.

richardcase: Potentially if there isn't a finalizer.

salasberryfin: That's what I initially thought, but the resource's yaml shows multiple finalizers. I just re-checked and validated this:
```
API Version:  management.cattle.io/v3      
Kind:         Cluster                  
Metadata:                                
  Creation Timestamp:  2024-02-07T18:50:28Z
  Finalizers:                     
    controller.cattle.io/cluster-agent-controller-cleanup
    controller.cattle.io/cluster-scoped-gc 
    controller.cattle.io/cluster-provisioner-controller                                                
    controller.cattle.io/mgmt-cluster-rbac-remove
    wrangler.cattle.io/mgmt-cluster-remove 
    ```

Comment (richardcase) [reply internal/controllers/import_controller_v3.go:218]: Maybe its a difference between wrangler and controller-runtime

Comment (richardcase) [reply internal/controllers/import_controller_v3.go:218]: I was also thinking, that perhaps something else is updating the it? Before the update caused by the delete?

Comment (salasberryfin) [reply internal/controllers/import_controller_v3.go:218]: I'd double check this. Even if I set a new finalizer that I never delete from the resource, it gets deleted. I expected this to make it stay in deleting status, so there must something that is interfering with this behavior.

Comment (salasberryfin) [reply internal/controllers/import_controller_v3.go:220]: Ultimately the issue was that, when configuring the watcher, it was not translating the rancher cluster into the capi cluster properly [here](https://github.com/rancher-sandbox/rancher-turtles/blob/dcacd4ac52a868d4b532bdd465c8ce506f32ee18/internal/controllers/import_controller_v3.go#L91), which caused the reconcile to not behave as expected.

File: internal/controllers/import_controller_v3.go [hunk 2]
```diff
@@ -0,0 +1,356 @@
...
+				GenerateName: "c-",
+				Labels: map[string]string{
+					capiClusterOwner:          capiCluster.Name,
+					capiClusterOwnerNamespace: capiCluster.Namespace,
+					ownedLabelName:            "",
+				},
+			},
+			Spec: managementv3.ClusterSpec{
+				DisplayName: capiCluster.Name,
+				Description: "CAPI cluster imported to Rancher",
```
richardcase, line 256: It would be good to have a way to override this default description (probably via annotation on the CAPI **Cluster**). But we can do that it a following PR>

internal/controllers/import_controller_v3.go
```diff
@@ -0,0 +1,356 @@
...
+	return ctrl.Result{}, nil
+}
+
+func (r *CAPIImportManagementV3Reconciler) rancherClusterToCapiCluster(ctx context.Context, clusterPredicate predicate.Funcs) handler.MapFunc {
+	log := log.FromContext(ctx)
+
+	return func(_ context.Context, o client.Object) []ctrl.Request {
+		labels := o.GetLabels()
+		capiCluster := &clusterv1.Cluster{ObjectMeta: metav1.ObjectMeta{
+			Name:      labels[capiClusterOwner],
```
richardcase, line 309: Should we do an explicit check for the label first?

richardcase (commented):
Great work @salasberryfin. This is looking good.

internal/controllers/import_controller_v3.go
```diff
@@ -0,0 +1,356 @@
...
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=clusters;clusters/status,verbs=get;list;watch;create;update;delete;patch
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machinepools,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machinepools/status,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machinepools/finalizers,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machines;machines/status,verbs=get;list;watch;delete;create;update;patch
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machinesets,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machinesets/status,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machinesets/finalizers,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=*,verbs=get;list;watch;create;update;patch;delete
```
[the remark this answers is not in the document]
salasberryfin: Removed a number of permissions which we don't seem to be using. E2E tests are passing after the edit.

salasberryfin:
@richardcase review suggestions have been addressed.

internal/controllers/import_controller_v3.go
salasberryfin: Created https://github.com/rancher-sandbox/rancher-turtles/issues/385 to track this enhancement.

internal/controllers/import_controller_test.go
```diff
@@ -54,6 +49,7 @@ var _ = Describe("reconcile CAPI Cluster", func() {
 		clusterRegistrationToken *managementv3.ClusterRegistrationToken
 		capiKubeconfigSecret     *corev1.Secret
 		clusterName              = "generated-rancher-cluster"
+		testObjectsSuffix        = "provisioningv1"
```
Danil-Grigorev, line 52: Is this suffix required? It seems a bit out of place.

internal/controllers/import_controller.go
```diff
@@ -208,7 +183,7 @@ func (r *CAPIImportReconciler) reconcile(ctx context.Context, capiCluster *clust
 	}
 
 	if !rancherCluster.ObjectMeta.DeletionTimestamp.IsZero() {
-		return r.reconcileDelete(ctx, capiCluster)
+		return reconcileDelete(ctx, capiCluster)
```
Danil-Grigorev, line 186: I don't think we need to unassign these methods from CAPIImportReconciler. Even if for now this is just capiCluster reference here, we might need to use different clients for operations later, and current code is better suited for that.

internal/controllers/helpers.go
```diff
@@ -0,0 +1,235 @@
+/*
```
Danil-Grigorev, line 1: I think that having helpers here is unnessesary. All of these methods were previously a part of reconcile object, they still are. It may be we only need to have a selector on the v1 or v3 cluster object inside the structure, to decide logic path.

internal/controllers/import_controller_v3.go
```diff
@@ -0,0 +1,365 @@
...
+	Client             client.Client
+	RancherClient      client.Client
+	recorder           record.EventRecorder
+	WatchFilterValue   string
+	Scheme             *runtime.Scheme
+	InsecureSkipVerify bool
+
+	controller         controller.Controller
+	externalTracker    external.ObjectTracker
+	remoteClientGetter remote.ClusterClientGetter
```
Danil-Grigorev, line 61: This structure is the same as previous CAPIImportReconciler. Why we need two?

internal/controllers/import_controller_v3.go
```diff
@@ -0,0 +1,365 @@
...
+	log.Info("Successfully applied import manifest")
+
+	return ctrl.Result{}, nil
+}
+
+func (r *CAPIImportManagementV3Reconciler) rancherClusterToCapiCluster(ctx context.Context, clusterPredicate predicate.Funcs) handler.MapFunc {
+	log := log.FromContext(ctx)
+
+	return func(_ context.Context, o client.Object) []ctrl.Request {
+		cluster, ok := o.(*managementv3.Cluster)
```
Danil-Grigorev, line 300: This code previously didn't have any static assumption on the source object for the events, so there was no dependency on v1.Cluster. Now it has a v3.Cluster check here I think is also not needed. This is all driven by the controller watches setup, where you specify what cluster to watch. I the v1 vs v3 difference in labels could be a simple if case with additional logic here.

internal/controllers/import_controller_v3.go
```diff
@@ -0,0 +1,365 @@
...
+
+	if len(rancherClusterList.Items) != 0 {
+		if len(rancherClusterList.Items) > 1 {
+			log.Info("More than one rancher cluster found. Will default to using the first one.")
+		}
+
+		rancherCluster = &rancherClusterList.Items[0]
+	}
+
+	if !capiCluster.ObjectMeta.DeletionTimestamp.IsZero() {
```
Danil-Grigorev, line 206: Do we set a finalizer on the CAPI Cluster object to prevent situations where we miss the event of cluster removal, so the Rancher v3.Cluster is left orphaned?

internal/rancher/management/v3/cluster.go
```diff
@@ -0,0 +1,151 @@
...
+// ClusterList contains a list of ClusterList.
+// +kubebuilder:object:root=true
+type ClusterList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []Cluster `json:"items"`
+}
+
+// ClusterCondition is the struct representing a condition of a Rancher Cluster.
+type ClusterCondition struct {
```
Danil-Grigorev, line 73: I think this custom logic is better to be replaced with `"sigs.k8s.io/cluster-api/util/conditions"` and implement the Getter interface: https://github.com/kubernetes-sigs/cluster-api/blob/96beb3577f5bbd5b762ac6e2cbfb05e9f8721aa9/util/conditions/getter.go#L29-L34

internal/controllers/import_controller.go
```diff
@@ -103,12 +86,11 @@ func (r *CAPIImportReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Ma
 		return fmt.Errorf("creating new controller: %w", err)
 	}
 
-	// Watch Rancher provisioningv2 clusters
+	// Watch Rancher provisioningv1 clusters
```
richardcase, line 89: He he, i meant v2prov when i wrote the comment and not meaning a reference to the group api version :)

This makes more sense now :)

internal/controllers/import_controller_test.go
richardcase: Makes it easier to see which test this relates to i guess.

internal/controllers/import_controller_v3.go
richardcase: I don't think there is any harm in this instance.

internal/controllers/import_controller_v3.go
richardcase: No we don't and thats a good point Danil. 

@salasberryfin - would you be able to add a finalizer as part of the reconcile.

internal/controllers/import_controller_v3.go
richardcase: Its the normal convention to check the type and return nil if you can't cast it:

- https://github.com/rancher-sandbox/rancher-turtles/blob/b3ea1a3a63c448f40b0b2f13015ca586fff1b193/internal/controllers/import_controller.go#L383
- https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controlplane/eks/controllers/awsmanagedcontrolplane_controller.go#L434

The fact we didn't have it previously is a mistake i would say.

internal/rancher/management/v3/cluster.go
richardcase: I agree :+1:

richardcase (commented):
This is generally looking good to me. I think the main things are adding the finalizer and also the conditions.

Personally i don't think we need to worry too much if there is a small amount of duplication. We'll be revisitng this if we make v3.Cluster the default (and only option?).

internal/controllers/import_controller_v3.go
salasberryfin: Thanks, I'll be fixing this

internal/controllers/import_controller_test.go
salasberryfin: With the existing configuration, running both `provisioning` and `management` tests would generate conflicts, as there are resources with the same name used by both test suites. This is used to keep them separated and avoid conflicts.

internal/controllers/import_controller_v3.go
salasberryfin: This check can also be found in upstream projects, for example: https://github.com/kubernetes-sigs/cluster-api/blob/0045baf02ce542fad0f96e42ab2decc41822256e/controlplane/kubeadm/internal/controllers/controller.go#L576

internal/controllers/import_controller_v3.go
salasberryfin: It now checks explicitly for the two required labels.

internal/rancher/management/v3/cluster.go
salasberryfin: This custom logic is taken from `rancher/wrangler` [here](https://github.com/rancher/wrangler/blob/004384a454ec6062420e10439bf9fa136fa38e4d/pkg/condition/condition.go). Had to re-use the functions as they're not exported in the original code. I think it makes sense to stick to the way `rancher` does it, but I'm happy to change it if you think otherwise.

internal/controllers/helpers.go
salasberryfin: This was just an effort to reduce code duplication. I'd say that, if we're deprecating the `provisioning.cattle.io` controller in the near future, we can just address this formatting when we do that. Then we can use a structure equivalent to what we've had until now.

internal/controllers/import_controller_v3.go
salasberryfin: Using specific structs just for the sake of simplicity and readability.

Danil-Grigorev: I agree, it’s just making it harder to review. There are many changes, but functionally there should be only: 
1. Apply v3 cluster instead of v1 
2. Watch on v3 cluster, map events to CAPI cluster via newly added labels on the resource
3. Collect registration token from the v3 cluster name (named) namespace.
4. Manually issue delete on both objects as their lifecycle is tied together in removal.
5. Optionally: Featuregate this functionality.
It is making difficult to verify there is only those change with so many code changes. Also because of that, there seems to be some test case duplication.

internal/controllers/import_controller.go
salasberryfin: For now these would be two identical functions. Can we address this when we make `management.cattle.io` the only controller? We could go back to the configuration we had before for `provisioning.cattle.io` and it would be extensible for potential future operations with other clients.

internal/controllers/import_controller_v3.go
Danil-Grigorev: This was a convention when controller-runtime may receive a partial object from an API server which will not be valid by schema for deserialization. This is not happening now. As we only need a namespace/name here, and we basically rely on conformity with client.Object interface, I don’t think all conventions should be preserved. We don’t rely on any api specific fields from v1 cluster or v3 cluster here. You can quote me on that, but I don’t see harm in dropping this anywhere now for the new projects.

internal/rancher/management/v3/cluster.go
Danil-Grigorev: The conditions is a kubernetes API, each project has to implement this equally to match API expectation. Wrangler is never a reliable source or example for anything, but here it does not matter what code to use, just importing this has 3 main benefits:
1. CAPI implementation is reliable, and is already used in our project.
2. If this copied code will have a security vulnerability, we will be responsible to fix it here. CAPI imported code will likely just be a dependency bump for us. 
3. Less boilerplate code in general opposed to copying.

richardcase: I think its probably best to follow conventions of upstream projects over wrangler (which we do not use).

internal/controllers/import_controller_v3.go
salasberryfin: Last version adds a finalizer to the capi cluster (only if the cluster is not annotated) which is deleted whenever the rancher cluster is removed.

internal/controllers/import_controller.go
salasberryfin: With the addition of the capi cluster finalizer, the deletion logic requires the finalizer be removed and `reconcileDelete` functions are no longer equivalent, so they are methods of the reconciler struct again.

internal/controllers/import_controller_v3_test.go
```diff
@@ -0,0 +1,422 @@
...
+		cluster := rancherClusters.Items[0]
+		Expect(cluster.Name).To(ContainSubstring("c-"))
+
+		cluster.Status.Conditions = []managementv3.ClusterCondition{
+			{
+				Type:   managementv3.ClusterConditionType(managementv3.ClusterConditionAgentDeployed),
+				Status: corev1.ConditionTrue,
+			},
+		}
+		Expect(managementv3.ClusterConditionAgentDeployed.IsTrue(&cluster)).To(BeTrue())
```
Danil-Grigorev, line 284:
```suggestion
		conditions.Set(cluster, conditions.TrueCondition(managementv3.ClusterConditionAgentDeployed))
```

internal/controllers/import_controller_test.go
Danil-Grigorev: I suspect these conflicts come from the fact we don’t create a temporary namespace for each test case. Envtest is very bad at removing things in time for the next test case. What you can use (also I’m working on introducing this in more places) is a new helper from `github.com/rancher-sandbox/rancher-turtles/internal/test/helpers`, which includes two things:
```go
		ns, err = testEnv.CreateNamespace(ctx, "ns-prefix") // create a unique ns
		Expect(err).ToNot(HaveOccurred())
// and to use in AfterSuite
                Expect(testEnv.Cleanup(ctx, ns, anyNumberOfObjects)).To(Succeed())
```

internal/rancher/management/v3/cluster.go
salasberryfin: Thanks for the suggestion. Using the CAPI conditions is quite a neat solution.

internal/controllers/import_controller_test.go
salasberryfin: I've tried to refactor the v3 tests with the new helpers and I'm still getting namespace collisions. This happens on the namespaces created from `import_sample.yaml`, specifically `cattle-system`, which I don't think we have control over during testing, other than modifying the yaml itself. Any ideas on whether this can be solved or we need to parse the yaml for each test?

internal/controllers/import_controller_v3.go
richardcase: That's a good point, if there has been a change in controller-runtime that makes this check redundant then yes for sure we should update the conventions we use/follow.

@Danil-Grigorev - out of curiosity so you know when the change was introduced into CR?

salasberryfin:
This is now updated and rebased with the recent changes to the test environment.

internal/controllers/import_controller_v3.go
salasberryfin: Thanks @Danil-Grigorev. Considering this, I removed the type check.

salasberryfin: Thanks @Danil-Grigorev. Considering this, I removed the type check.

Danil-Grigorev: @richardcase [Here](https://github.com/kubernetes-sigs/controller-runtime/blob/56159419231e985c091ef3e7a8a3dee40ddf1d73/pkg/reconcile/reconcile.go#L137) is for example a conversion without error check and it is a generic one every controller will now use. Additionally here we only need namespace/name, so for the sake of this method only this info required. And as we have it in available, then it was typechecked before.

salasberryfin:
Reminder that there's also this documentation PR https://github.com/rancher-sandbox/rancher-turtles-docs/pull/59 to document the new feature flag.

internal/controllers/helpers.go
```diff
@@ -0,0 +1,214 @@
...
+		}
+
+		_, autoImport := util.ShouldImport(ns, importLabelName)
+		if !autoImport {
+			log.V(2).Info("Namespace doesn't have import annotation label with a true value, skipping")
+			return nil
+		}
+
+		capiClusters := &clusterv1.ClusterList{}
+		if err := rancherClient.List(ctx, capiClusters, client.InNamespace(o.GetNamespace())); err != nil {
```
Danil-Grigorev, line 110: Previously the r.Client was used here, not r.RancherClient. This is confusing, even if the invocation is using the correct r.Client.

internal/controllers/import_controller.go
```diff
@@ -271,7 +247,7 @@ func (r *CAPIImportReconciler) reconcileNormal(ctx context.Context, capiCluster
 	}
 
 	// get the registration manifest
-	manifest, err := r.getClusterRegistrationManifest(ctx, rancherCluster.Status.ClusterName, capiCluster.Namespace)
+	manifest, err := getClusterRegistrationManifest(ctx, rancherCluster.Status.ClusterName, capiCluster.Namespace, r.Client, r.InsecureSkipVerify)
```
Danil-Grigorev, line 250: Here the r.RancherClient should be used, see https://github.com/rancher-sandbox/rancher-turtles/pull/362/files#diff-df81a5620bfede50d1c70fd0ebb982e652d28721e15b825c52dddcf97ceda0deL337

salasberryfin: Thanks for noticing. These two are unintended errors because of naming similarities. Fixing them now.

internal/controllers/import_controller_v3.go
```diff
@@ -0,0 +1,393 @@
...
+	if err := r.Client.Get(ctx, req.NamespacedName, capiCluster); err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{Requeue: true}, nil
+		}
+
+		return ctrl.Result{Requeue: true}, err
+	}
+
+	if capiCluster.ObjectMeta.DeletionTimestamp.IsZero() && !turtlesannotations.HasClusterImportAnnotation(capiCluster) {
+		if !controllerutil.ContainsFinalizer(capiCluster, managementv3.CapiClusterFinalizer) {
```
Danil-Grigorev, line 142:
```suggestion
	if capiCluster.ObjectMeta.DeletionTimestamp.IsZero() && !turtlesannotations.HasClusterImportAnnotation(capiCluster) && !controllerutil.ContainsFinalizer(capiCluster, managementv3.CapiClusterFinalizer) {
```
But to be honest this feels like it belongs under reconcileNormal scenario

internal/controllers/import_controller_v3.go
```diff
@@ -0,0 +1,393 @@
...
+		if len(rancherClusterList.Items) > 1 {
+			log.Info("More than one rancher cluster found. Will default to using the first one.")
+		}
+
+		rancherCluster = &rancherClusterList.Items[0]
+	}
+
+	if !capiCluster.ObjectMeta.DeletionTimestamp.IsZero() {
+		if err := r.deleteDependentRancherCluster(ctx, capiCluster); err != nil {
+			return ctrl.Result{Requeue: true}, fmt.Errorf("error deleting associated managementv3.Cluster resources: %w", err)
```
Danil-Grigorev, line 229:
```suggestion
			return ctrl.Result{}, fmt.Errorf("error deleting associated managementv3.Cluster resources: %w", err)
```
Error causes automatic requeue.

internal/controllers/import_controller_v3.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+
+	return r.reconcileNormal(ctx, capiCluster, rancherCluster)
+}
+
+func (r *CAPIImportManagementV3Reconciler) reconcileNormal(ctx context.Context, capiCluster *clusterv1.Cluster,
+	rancherCluster *managementv3.Cluster,
+) (ctrl.Result, error) {
+	log := log.FromContext(ctx)
+
```
Danil-Grigorev, line 244:
```suggestion
createCopy := rancherCluster.DeepCopy()
```

internal/controllers/import_controller_v3.go
```diff
@@ -0,0 +1,393 @@
...
+				Labels: map[string]string{
+					capiClusterOwner:          capiCluster.Name,
+					capiClusterOwnerNamespace: capiCluster.Namespace,
+					ownedLabelName:            "",
+				},
+			},
+			Spec: managementv3.ClusterSpec{
+				DisplayName: capiCluster.Name,
+				Description: "CAPI cluster imported to Rancher",
+			},
```
Danil-Grigorev, line 270:
```suggestion
		if err := r.RancherClient.Create(ctx, createCopy); err != nil {
```

internal/controllers/import_controller_v3.go
```diff
@@ -0,0 +1,393 @@
...
+		}); err != nil {
+			return ctrl.Result{}, fmt.Errorf("error creating rancher cluster: %w", err)
+		}
+
+		return ctrl.Result{Requeue: true}, nil
```
Danil-Grigorev, line 275:
```suggestion
rancherCluster = createCopy
```
Don’t think we need to return here. Rancher cluster was created and stored, we just need to swap the pointers with the copy.

test/e2e/suites/managementv3/managementv3_test.go
```diff
@@ -0,0 +1,125 @@
...
+			CAPIClusterCreateWaitName:      "wait-rancher",
+			DeleteClusterWaitName:          "wait-controllers",
+			CapiClusterOwnerLabel:          e2e.CapiClusterOwnerLabel,
+			CapiClusterOwnerNamespaceLabel: e2e.CapiClusterOwnerNamespaceLabel,
+			OwnedLabelName:                 e2e.OwnedLabelName,
+		}
+	})
+})
+
+var _ = Describe("[Azure] [AKS] Create and delete CAPI cluster functionality should work with namespace auto-import", Label(e2e.DontRunLabel), func() {
```
Danil-Grigorev, line 64:
```suggestion
var _ = Describe("[Azure] [AKS] [management.cattle.io/v3] Create and delete CAPI cluster functionality should work with namespace auto-import", Label(e2e.DontRunLabel), func() {
```

internal/controllers/import_controller_v3_test.go
```diff
@@ -0,0 +1,441 @@
...
+					Name:      capiCluster.Name,
+				},
+			})
+			g.Expect(err).ToNot(HaveOccurred())
+			g.Expect(res.RequeueAfter).To(Equal(defaultRequeueDuration))
+		}).Should(Succeed())
+	})
+
+	It("should reconcile a CAPI cluster when rancher cluster doesn't exist", func() {
+		capiCluster.Labels = map[string]string{
```
Danil-Grigorev, line 163: I think here a namespace import label should be removed, so that this statement would be tested.

Danil-Grigorev (dismissed):
Some nits, overall looks good

internal/controllers/import_controller_v3.go
salasberryfin: Not sure about this change. It is definitely cleaner but I think it doesn't help with readability.

salasberryfin: And the cluster object defined in `reconcile` is not exactly what we need to create later in `reconcileNormal`, as it needs to have a `name` and `namespace` to be testable, which we cannot use when actually creating the resource.

internal/controllers/import_controller_v3.go
```diff
@@ -0,0 +1,393 @@
...
+	rancherCluster := &managementv3.Cluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "c-",
+			Namespace: capiCluster.Namespace,
+			Labels: map[string]string{
+				capiClusterOwner:          capiCluster.Name,
+				capiClusterOwnerNamespace: capiCluster.Namespace,
+			},
+		},
+	}
```
Danil-Grigorev, line 202:
```suggestion
	rancherCluster := &managementv3.Cluster{}
```
Just noticed that this is an invalid object, it has `c-` name on it. If this is needed for tests to work, then what tests are testing? We should not put mocks in production code just for testing. Removing this will also help with readability.

internal/controllers/helpers.go
```diff
@@ -0,0 +1,214 @@
...
+	log := log.FromContext(ctx)
+
+	return func(_ context.Context, o client.Object) []ctrl.Request {
+		ns, ok := o.(*corev1.Namespace)
+		if !ok {
+			log.Error(nil, fmt.Sprintf("Expected a Namespace but got a %T", o))
+			return nil
+		}
+
+		_, autoImport := util.ShouldImport(ns, importLabelName)
```
alexander-demicev, line 103:
```suggestion
		if _, autoImport := util.ShouldImport(ns, importLabelName); !autoImport {
```

internal/controllers/import_controller_v3.go
```diff
@@ -0,0 +1,393 @@
...
+		For(&clusterv1.Cluster{}).
+		WithOptions(options).
+		WithEventFilter(capiPredicates).
+		Build(r)
+	if err != nil {
+		return fmt.Errorf("creating new controller: %w", err)
+	}
+
+	// Watch Rancher managementv3 clusters
+	err = c.Watch(
```
alexander-demicev, line 92:
```suggestion
	if err := c.Watch(
```

internal/controllers/import_controller_v3.go
alexander-demicev: If I'm not mistaken in case of error `Requeue: true` will be ignored. Controller runtime will handle exponential backoff when error is returned.

internal/controllers/import_controller_v3.go
alexander-demicev: what's exactly the issue with tests?

alexander-demicev (commented):
looks great, couple minor comments

internal/controllers/import_controller_v3.go
salasberryfin: Just re-tested this and it is not needed anymore, I'll be fixing it. Honestly I couldn't reproduce the scenario that caused this issue, but I remember tests were complaining about the resource not having name nor namespace and that's why I added a placeholder. Thought it could be coming from when we tried to use resource ownership to link capi cluster to rancher cluster, because we then used a `Get` to fetch the cluster instead of a `List`, but it doesn't seem like that's the case. Thanks for reporting.

internal/controllers/helpers.go
```diff
@@ -0,0 +1,214 @@
...
+) (string, error) {
+	log := log.FromContext(ctx)
+
+	token := &managementv3.ClusterRegistrationToken{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      clusterName,
+			Namespace: namespace,
+		},
+		Spec: managementv3.ClusterRegistrationTokenSpec{
+			ClusterName: clusterName,
```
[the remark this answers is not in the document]
salasberryfin: Sorry, I don't understand what you mean. Aren't we doing the `Get` request after setting the object `Spec` already? https://github.com/rancher-sandbox/rancher-turtles/blob/e1ab73b9b19078b10c4fd3c23496ee51e438724f/internal/controllers/helpers.go#L70

internal/controllers/import_controller_v3.go
salasberryfin: The reason I opted for having this here is to avoid problems due to changes to the object. Since the capi cluster is being updated by the capi controller and then we're also applying the finalizer, doing it in `reconcileNormal` would mean that we would have to deal with the changed object and then repeat the same when patching after reconcile [here](https://github.com/rancher-sandbox/rancher-turtles/blob/e1ab73b9b19078b10c4fd3c23496ee51e438724f/internal/controllers/import_controller_v3.go#L174). I think the downside of having it here is that there's a corner case in which the finalizer is set on the capi cluster and then the rancher cluster fails to create, but I'd say this is better than the situation in which we have a rancher cluster created but we fail to add the finalizer to the capi cluster.

Danil-Grigorev (approved):

richardcase (approved):

internal/controllers/helpers.go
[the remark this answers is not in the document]
richardcase: I think what @alexander-demicev is saying that generally when you do a get you only fill in the **ObjectMetadata** and not the **Spec**.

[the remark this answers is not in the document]
salasberryfin: Ok, sorry I did not get it at first. This is taken as-is from the existing `provisioning.cattle.io` controller. Perhaps we can carry changing this over to the "Move to v3 cluster" follow-up issue?

alexander-demicev (approved):

PR #398, opened 2024-02-23
:bug: Fix dependabot duplicate rule issue

Danil-Grigorev:
Apparently dependabot does not support separation of groups for the same dependency engine (gomod). 

https://github.com/rancher-sandbox/rancher-turtles/runs/21812211327

`The property '#/updates/1' is a duplicate. Update configs must have a unique combination of 'package-ecosystem', 'directory', and 'target-branch'`

The problem appears to be due to: https://github.com/dependabot/dependabot-core/issues/1778

To still benefit from early CAPI Operator version bump, changing interval to daily.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -4,7 +4,7 @@ updates:
   - package-ecosystem: "gomod"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "daily"
     labels:
       - "kind/cleanup"
       - "area/dependency"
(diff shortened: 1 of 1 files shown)


richardcase (approved):

furkatgofurov7 (commented):
I am afraid the repo will be flooded with dependabot PRs after this change, weekly used to work fine so far. Would not just removing ignore be sufficient?

Danil-Grigorev:
@furkatgofurov7 Our spec is grouping updates together. Worst which may happen, we will always have 1 PR open, which should not be much of an issue. Flooding of dependabot PRs is happening in operator repos ATM. I have filters set to separate email notifications for dependency bumps and general inbox. This helps.

furkatgofurov7 (approved):

PR #433, opened 2024-03-14
:bug: Fix agent node affinity settings to always prefer control plane nodes

Danil-Grigorev:

Fixes #

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,7 @@ GINKGO_ARGS ?=
 SKIP_RESOURCE_CLEANUP ?= false
 USE_EXISTING_CLUSTER ?= false
 ISOLATED_MODE ?= false
+CNI ?= calico # alternatively: kindnet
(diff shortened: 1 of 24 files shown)


richardcase:
I think we need to hold off on this change for the following reasons:

- We've not seen any issues in our testing specifically related to the scheduling/placement of the agent pod(s). 
- We don't have time to thoroughly test the impact of this before the release
- We should consider fixing this upstream instead

richardcase:
Lets split out the helm addon and changing the tests to use calico instead of kindnet.

furkatgofurov7:
@Danil-Grigorev let's break this PR into two, the one with the changes @richardcase suggested and the latter we can go through in more detail.

Danil-Grigorev:
I’ll close this PR and open a separate one with calico changes only later.

PR #452, opened 2024-03-28
🌱 Downgrade CAPI to 1.4.6

Danil-Grigorev:
Due to rancher 2.8.3 still using 1.4.6 CAPI version, we will use this by default too.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/rancher-turtles/templates/clusterctl-config.yaml
+++ b/charts/rancher-turtles/templates/clusterctl-config.yaml
@@ -9,7 +9,7 @@ data:
     providers:
     # Cluster API core provider
     - name:         "cluster-api"
-      url:          "https://github.com/kubernetes-sigs/cluster-api/releases/v1.6.2/core-components.yaml"
+      url:          "https://github.com/kubernetes-sigs/cluster-api/releases/v1.4.6/core-components.yaml"
       type:         "CoreProvider"
 
     # Infrastructure providers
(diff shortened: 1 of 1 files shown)


furkatgofurov7:
How about other provider versions in the file, i.e CAPA v2.3.5 [uses](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/v2.3.5/go.mod#L45) CAPI v1.5.3, should they all also be updated to versions which uses v1.4.X version of CAPI?

richardcase (dismissed):

richardcase:
> How about other provider versions in the file, i.e CAPA v2.3.5 [uses](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/v2.3.5/go.mod#L45) CAPI v1.5.3, should they all also be updated to versions which uses v1.4.X version of CAPI?

I think it should be ok as its based on the api contract. But i will check something quickly....

richardcase:
Doing:

```bash
clusterctl init --core cluster-api:v1.4.6 -i aws
```

Results in CAPA **v2.4.1** being installed

Danil-Grigorev:
Should we upgrade it then? :)

furkatgofurov7:
> Doing:
> 
> ```shell
> clusterctl init --core cluster-api:v1.4.6 -i aws
> ```
> 
> Results in CAPA **v2.4.1** being installed

That is expected to fetch the latest patch if not specified? My point was how provider version which does not support the CAPI version can be installed together )

Danil-Grigorev:
We were testing these versions in our CI with CAPI 1.4.6. Current set of defaults is safe from this perspective.

richardcase:
> We were testing these versions in our CI with CAPI 1.4.6. Current set of defaults is safe from this perspective.

I agree. We have tested these combinations. Also the API version are the same.

furkatgofurov7 (approved):

richardcase (approved):

PR #484, opened 2024-04-18
:seedling: cover cluster reimport after running e2e

### LINKED ISSUE #266: Add an e2e scenario covering re-import of rancher manifests
We need to add a test scenario covering re-import of the removed rancher cluster instance, after the `ClusterImportedAnnotation` is removed from the CAPI cluster instance, preventing automatic import.

salasberryfin:

This change adds a re-import check after running the existing `import_gitops` suite. The validation sits behind a condition passed as input `TestClusterReimport` which is proposed to only be enabled on CAPD tests to limit the effect on running time, which is already considerably long. This is also the reason why the new re-import validation is added to an existing test suite and not as a new one.

Both controller types are supported for now (v1 and v3 clusters) until we move completely to only v3.

To simplify and reduce code duplication, the validation on the Rancher cluster resource is now part of a `validateRancherCluster` function which is called during the "normal" test and then again after re-import.

The process of re-import is as follows:
- Rancher cluster is removed, which causes the CAPI cluster to be annotated: `imported=true`.
- Validate that the cluster is no longer imported and annotation is properly set.
- Retrieve CAPI cluster annotations and remove `imported=true`.
- Validate annotation is removed and wait for Rancher cluster to appear.
- Re-run the existing tests on Rancher cluster (agent deployed, cluster ready, etc.).

Fixes #266 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/test/e2e/specs/import_gitops.go
+++ b/test/e2e/specs/import_gitops.go
@@ -76,6 +76,9 @@ type CreateUsingGitOpsSpecInput struct {
 	SkipDeletionTest bool
 
 	LabelNamespace bool
+
+	// TestClusterReimport defines whether to test un-importing and re-importing the cluster after initial test.
+	TestClusterReimport bool
 }
 
 // CreateUsingGitOpsSpec implements a spec that will create a cluster via Fleet and test that it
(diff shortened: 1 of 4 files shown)


alexander-demicev (approved):
LGTM, thanks for taking care of this

Danil-Grigorev (approved):
Thanks for adding this, LGTM

PR #505, opened 2024-04-23
Add ETCD Snapshot and Restore API skeleton

alexander-demicev:

This PR only adds the ETCD snapshot/restore API skeleton and hides its installation behind helm chart value feature flag

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


alexander-demicev:
@furkatgofurov7 I moved all manifests under `config/exp/*` so CRDs and RBAC for example will live here `config/exp/crd`, `config/exp/rbac`, thanks for the suggestion. For your second comment, I'd prefer to keep experimental APIs separated because it's easier to template them for helm so they are hidden behind feature flag. CAPI always releases and installs all experimental APIs, they don't run controllers for them by default but CRDs are installed. Since we are using helm, we can provide a better UX and not install some CRDs unless it's necessary.

Makefile
```diff
@@ -495,6 +503,8 @@ release: clean-release $(RELEASE_DIR)  ## Builds and push container images using
 .PHONY: build-chart
 build-chart: $(HELM) $(KUSTOMIZE) $(RELEASE_DIR) $(CHART_RELEASE_DIR) $(CHART_PACKAGE_DIR) ## Builds the chart to publish with a release
 	$(KUSTOMIZE) build ./config/chart > $(CHART_DIR)/templates/rancher-turtles-components.yaml
+	$(KUSTOMIZE) build ./config/exp > $(CHART_DIR)/templates/rancher-turtles-exp-components.yaml
```
furkatgofurov7, line 506: I think it makes sense to move kustomization/kustomizeconfig yAML files to new crd folder as well, since these are crd specific
```suggestion
	$(KUSTOMIZE) build ./config/exp/crd > $(CHART_DIR)/templates/rancher-turtles-exp-components.yaml
```

furkatgofurov7 (commented):
One nit, otherwise LGTM:

Makefile
alexander-demicev: Current command includes everything in `./config/exp/*`, at the moment there are only CRDs but when we add controllers we're going to see RBAC also present there.

furkatgofurov7 (approved):

Danil-Grigorev:
This seems out of place for this repo, have you considered implementing this as an addon/helm chart and ship by default with turtles but from another repo? This has no dependency on turtles and vice-versa, so keeping this up-to-date will not be a problem as there is no compatibility matrix. Experimental phase allows to use latest at all times. Additionally your POC can be reused, as this seems to be doing a duplication of initial work.

alexander-demicev:
Yes, we considered a separate repository. We want day 2 operations to be a part of the turtles extension and also we'd like to keep it here to reduce the number of repositories and save time setting up new repos. We can follow a similar to CAPI approach where some subprojects live in the core repo. @furkatgofurov7 @richardcase please correct me if I'm wrong.

Danil-Grigorev:
CAPI scenario is different from this. To be closer to upstream way, this needs to be a subproject, and have an independent `go.mod` file too. CAPI places experimental API types separately with the expectation that they will eventually become core. Additionally experimental can import core CAPI types. ETCD snapshot/restore shouldn't.

salasberryfin (approved):
LGTM. I think that, considering that in terms of functionality this is a core part of Turtles, I'd favor having it inside the main Turtles repository, even if there is no direct dependency at the code level. Especially, it will reduce and simplify chores for the future.

config/exp/etcdrestore/kustomization.yaml
```diff
@@ -0,0 +1,35 @@
...
+    group: apiextensions.k8s.io
+    version: v1
+    kind: CustomResourceDefinition
+    name: etcdmachinesnapshots.turtles-capi.cattle.io
+  path: crd/patches/turtles-capi.cattle.io_etcdmachinesnapshots.yaml
+- target:
+    group: apiextensions.k8s.io
+    version: v1
+    kind: CustomResourceDefinition
+    name: capiproviders.turtles-capi.cattle.io
```
furkatgofurov7, line 22: this needs a change found by @yiannistri https://github.com/rancher/turtles/pull/506#discussion_r1586502725

furkatgofurov7 (changes requested):

config/exp/etcdrestore/crd/patches/turtles-capi.cattle.io_etcdmachinesnapshots.yaml
Danil-Grigorev, line 1: This and another patch can be dropped. This behavior is possible to achieve with kubebuilder defaults now

Danil-Grigorev (approved):

furkatgofurov7 (approved):

PR #506, opened 2024-04-23
Add ETCD Snapshot and Restore Controllers/Webhooks/Reconcilers skeleton

### LINKED ISSUE #495: [etcdbackup&restore] Add skeleton for controllers/reconcilers/webhooks
We should add main.go, and basic skeleton structure for controllers/reconcilers/webhooks from the POC [code](https://github.com/furkatgofurov7/turtles-etcd-restore/tree/main) and hide it behind the feature gate.

furkatgofurov7:
Fixes #495 
Needs rebase after https://github.com/rancher/turtles/pull/505 is merged

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests


salasberryfin (approved):
LGTM when https://github.com/rancher/turtles/pull/505 is merged.

yiannistri (approved):
LGTM, nice work :+1:

config/exp/crd/kustomization.yaml
```diff
@@ -0,0 +1,35 @@
...
+    group: apiextensions.k8s.io
+    version: v1
+    kind: CustomResourceDefinition
+    name: etcdmachinesnapshots.turtles-capi.cattle.io
+  path: patches/turtles-capi.cattle.io_etcdmachinesnapshots.yaml
+- target:
+    group: apiextensions.k8s.io
+    version: v1
+    kind: CustomResourceDefinition
+    name: capiproviders.turtles-capi.cattle.io
```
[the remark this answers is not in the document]
furkatgofurov7: Good catch, it is cherry picked change from https://github.com/rancher/turtles/pull/505/files#diff-5e1f0bd75a6d1f2eb59d8a9d5f8347c0b25db9a7d92afa3327275b33f2646916R22 and will be solved either there or here. 

cc @alexander-demicev

exp/etcdrestore/go.mod
```diff
@@ -0,0 +1,92 @@
+module github.com/rancher/turtles/exp/etcdrestore
+
+go 1.22.0
```
alexander-demicev, line 3: do we want a separate go mod? can we keep the main one from the root directory for simplicity? We can always add it back if we see issues with that

Danil-Grigorev: With a single go.mod there will be issues with CAPI version because of a clash on RKE2 provider import and CAPI operator import.

furkatgofurov7: Initially, I went on and used the main one, however, after an offline discussion with @Danil-Grigorev and his concerns around adding new dependencies in the root go.mod, I decided to keep it separated.

furkatgofurov7: The list of things we might want to avoid by @Danil-Grigorev:

- when the implementation will grow, etcd backup restore being in experemental will introduce new dependencies, I saw [github.com/rancher-sandbox/cluster-api-provider-rke2](http://github.com/rancher-sandbox/cluster-api-provider-rke2) v0.2.7 in the POC repo. This will bring collision between CAPI versions and Controller runtime versions, which will have to be solved for both turtles and etcd backup/restore. May cause us headaches later on.
- Our automated dependabot updates will be even more fragile then they are now, if we grow the go.mod file, because of co-dependencies of the libraries we may not know of.
- There are some singleton things in turtles, like CRDs or webhooks for example, but not only, and I think that some of the setup may overlap with etcd backup restore. Which may require hacks to fix.
- In terms of availability, having both controllers run from single main.go may result in more frequent failures. It may be especially frustrating if we will see a problem with turtles because of etcd backup/restore change, while it is being experimental and disabled in the environment. Separating main.go (and deployments, if it comes to it) will not have this problem.

@alexander-demicev WDYT?

Danil-Grigorev (approved):

config/exp/etcdrestore/default/manager_role_aggregation_patch.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: manager-role
+  labels:
+    turtles-capi.cattle.io/aggregate-to-manager: "true"
```
yiannistri, line 6: Does this label match the label used in the aggregated ClusterRole ?

yiannistri (dismissed):
LGTM

exp/etcdrestore/go.mod
alexander-demicev: sounds good to me, dependencies might harder to manage

alexander-demicev (dismissed):

config/exp/etcdrestore/default/manager_role_aggregation_patch.yaml
furkatgofurov7: Good catch @yiannistri, pushed https://github.com/rancher/turtles/pull/506/commits/952ff064abf6757c008153e7460abf06239ff3c7

salasberryfin (approved):

Danil-Grigorev (approved):

Issue #477, opened 2024-04-10
Extend `v3.Cluster` coverage in E2E


salasberryfin:
### Description

When the support for importing CAPI clusters based on `management.cattle.io` resources was first introduced in https://github.com/rancher/turtles/pull/362, a single E2E test suite was added for this alternative controller (based on the existing `import-gitops` suite).

As we ultimately move to `v3.Cluster` we need to extend the E2E coverage. The initial approach could be to migrate the existing suites to focus on `management.cattle.io` cluster resources rather than `provisioning.cattle.io`. Doing so, we guarantee that we can support these scenarios with the new controller and importing mechanism.

This will mean that we need to cover (at least):
- GitOps approach via Fleet.
- Validate that we do not break v2prov.
- Update labels.
- Deploying Rancher with `embedded-cluster-api` feature disabled.

salasberryfin:
Following on the work done in https://github.com/rancher-sandbox/cluster-api-provider-rke2/pull/306, the E2E tests for the v3 controller must cover CAPRKE2.

PR #519, opened 2024-04-29
Add CAPRKE2 to e2e suite

alexander-demicev:

Fixes https://github.com/rancher-sandbox/rancher-turtles/issues/104

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


salasberryfin (approved):
LGTM. Referenced this in https://github.com/rancher/turtles/issues/477 to cover this for v3 controller too.

Danil-Grigorev (approved):

test/e2e/data/cluster-templates/docker-rke2.yaml
```diff
@@ -0,0 +1,168 @@
...
+      bootstrapTimeout: 15m
+---
+apiVersion: controlplane.cluster.x-k8s.io/v1alpha1
+kind: RKE2ControlPlane
+metadata:
+  name: ${CLUSTER_NAME}-control-plane
+spec: 
+  replicas: ${CONTROL_PLANE_MACHINE_COUNT}
+  agentConfig:
+    version: ${KUBERNETES_VERSION}+rke2r1
```
Danil-Grigorev, line 51:
```suggestion
  replicas: ${CONTROL_PLANE_MACHINE_COUNT}
  version: ${KUBERNETES_VERSION}+rke2r1
```
After the RKE2 is bumped? Is this how it should look like?

test/e2e/data/cluster-templates/docker-rke2.yaml
```diff
@@ -0,0 +1,168 @@
...
+  name: ${CLUSTER_NAME}-md-0
+spec:
+  clusterName: ${CLUSTER_NAME}
+  replicas: ${WORKER_MACHINE_COUNT}
+  selector:
+    matchLabels:
+      cluster.x-k8s.io/cluster-name: ${CLUSTER_NAME}
+  template:
+    spec:
+      version: ${KUBERNETES_VERSION}
```
Danil-Grigorev, line 94:
```suggestion
      version: ${KUBERNETES_VERSION}+rke2r1
```
This template should cause an error due to future logic, probably it has to have correct value?

Danil-Grigorev (approved):

test/e2e/data/cluster-templates/docker-rke2.yaml
alexander-demicev: yes but old version field will also be valid for some time

test/e2e/data/cluster-templates/docker-rke2.yaml
alexander-demicev: It won't break on update, CAPRKE2 will ignore this field if it doesn't contain suffix and fallback to agentConfig.Version

furkatgofurov7 (approved):

test/e2e/data/cluster-templates/docker-rke2.yaml
Danil-Grigorev: The fact that API was still in alpha was the only chance to avoid this “both are valid” behavior. This will be another problem to solve in the future

PR #556, opened 2024-05-27
fix: git add files in dependabot workflow

salasberryfin:

Dependabot CI checks are failing on some PRs after replacing the no longer supported action with manual steps [example](https://github.com/rancher/turtles/actions/runs/9249544453/job/25447774389). Looks like this is because the Action we were using sets a default `git add .` [here](https://github.com/EndBug/add-and-commit?tab=readme-ov-file#inputs) and it is missing now, so the commit returns error code 1 when there are no changes to commit. This PR simply adds a `git add .` which hopefully is enough to mimic the removed action's behavior.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -31,4 +31,5 @@ jobs:
       run: |
         git config --local user.email "49699333+dependabot[bot]@users.noreply.github.com"
         git config --local user.name "github-actions[bot]"
+        git add .
         git commit -m "Update generated code"


Danil-Grigorev (approved):

Danil-Grigorev:
We might want to explore using rancher/renovate bot to do these operation in the future. Current workflow is unfortunately very limited and requires manual patches within the changing requirements

salasberryfin:
Thanks @Danil-Grigorev, I agree with your point. I guess we can keep this as a temporary solution and think about migrating this to a more sophisticated alternative.

Danil-Grigorev:
Yeah, same thinking @salasberryfin. Opened an issue for the future: https://github.com/rancher/turtles/issues/558

alexander-demicev (approved):

PR #561, opened 2024-05-28
Mark propagate labels as experimental

### LINKED ISSUE #560: Leave cluster label propagation as experimental
Currently we need to evaluate usages of label propagation before marking this as enabled by default. Users can always enable it manually through questions or helm chart values.

Danil-Grigorev:

Fixes #560 

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/internal/controllers/suite_test.go
+++ b/internal/controllers/suite_test.go
@@ -23,6 +23,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/rancher/turtles/feature"
 	"github.com/rancher/turtles/internal/test/helpers"
 	"k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -49,6 +50,12 @@ var (
 	kubeConfigBytes []byte
 	ctx             = ctrl.SetupSignalHandler()
 )
+
+func init() {
+	utilruntime.Must(feature.MutableGates.SetFromMap(map[string]bool{
+		string(feature.PropagateLabels): true,
+	}))
+}
 
 func setup() {
 	utilruntime.Must(clusterv1.AddToScheme(scheme.Scheme))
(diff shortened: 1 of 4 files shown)


alexander-demicev (dismissed):

internal/controllers/suite_test.go
```diff
@@ -50,6 +51,12 @@ var (
 	ctx             = ctrl.SetupSignalHandler()
 )
 
+func init() {
+	utilruntime.Must(feature.MutableGates.SetFromMap(map[string]bool{
+		string(feature.PropagateLabels): true,
```
yiannistri, line 56: Just for my understanding: does this indicate that the PropagateLabels feature flag is mutable i.e. it can be turned on or off by the end user?

yiannistri (approved):
LGTM

internal/controllers/suite_test.go
Danil-Grigorev: No, unless they build it with this code. Once the FG enabled by default, it can’t be disabled unless there is a handler like this exposed to the user. The `Gates` is immutable, so this has to happen before it is injected in the variable inside `init()`. This line here just satisfies the existing tests expectation on the FG remaining being set to true.

Danil-Grigorev: Sorry @yiannistri let me correct myself - it is in fact mutable. We expose flag `--feature-gates` to override the defaults: 
```
      --feature-gates mapStringBool            A set of key=value pairs that describe feature gates for alpha/experimental features. Options are:
                                               AllAlpha=true|false (ALPHA - default=false)
                                               AllBeta=true|false (BETA - default=false)
                                               ClusterResourceSet=true|false (BETA - default=true)
                                               ClusterTopology=true|false (ALPHA - default=false)
                                               KubeadmBootstrapFormatIgnition=true|false (ALPHA - default=false)
                                               MachinePool=true|false (BETA - default=true)
                                               MachineSetPreflightChecks=true|false (ALPHA - default=false)
                                               RuntimeSDK=true|false (ALPHA - default=false)
                                               etcd-snapshot-restore=true|false (BETA - default=false)
                                               managementv3-cluster=true|false (BETA - default=false)
                                               propagate-labels=true|false (BETA - default=false)
                                               rancher-kube-secret-patch=true|false (BETA - default=false)
```
Although I’m seeing some leftover features in the list :)

salasberryfin (approved):

PR #562, opened 2024-05-30
test: add rke2 cluster to e2e test cases

salasberryfin:

Provision RKE2 CAPI clusters in short e2e tests. This PR also updates the version of the Docker provider, as we need to use some of the fields that don't exist in the old provider version.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


test/e2e/config/operator.yaml
```diff
@@ -27,7 +27,7 @@ intervals:
 
 variables:
   RANCHER_VERSION: "v2.8.1"
-  KUBERNETES_VERSION: "v1.27.0"
+  KUBERNETES_VERSION: "v1.27.3"
```
furkatgofurov7, line 30: how about using newer version here? kind image was already v1.28.0 prior to this change

furkatgofurov7 (commented):
Overall looks good thanks @salasberryfin, one question/suggestion inline:

test/e2e/data/capi-operator/capi-providers.yaml
```diff
@@ -12,6 +12,7 @@ metadata:
 spec:
   name: docker
   type: infrastructure
+  version: v1.7.0
```
Danil-Grigorev, line 15: This needs to be checked that it's working

test/e2e/data/cluster-templates/docker-rke2.yaml
```diff
@@ -71,13 +73,19 @@ metadata:
...
 apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
 kind: RKE2ConfigTemplate
 metadata:
   name: ${CLUSTER_NAME}-md-0
+spec: 
+  template:
+    spec:
+      agentConfig:
+        nodeAnnotations:
+          test: "true"
```
Danil-Grigorev, line 88: Node annotations is a specific feature, I don’t see it making sense to duplicate tests from RKE2 provider previously implemented here
```suggestion
spec: {}
```

test/e2e/data/cluster-templates/docker-rke2.yaml
```diff
@@ -17,10 +17,10 @@ spec:
   clusterNetwork:
     pods:
       cidrBlocks:
-      - 10.45.0.0/16
+      - 10.1.0.0/16
```
Danil-Grigorev, line 20: Thanks for updating this, I see the change follows in CAAPF steps, which is great. This range also does not overlap with anything for my machine.

test/e2e/data/capi-operator/capi-providers.yaml
Danil-Grigorev: Checked this, seems to be working. Which is good, as by default we still install 1.4.6 but users can upgrade to latest using default repo index in CAPI. Will not work for providers which are not added to CAPI, like elemental or CAAPF, k3s, etc. However we need to keep `clusterctl.yaml` config map on 1.4.6 because this updated version will break rancher provisioning.
```
│ NAMESPACE↑                      NAME                     TYPE               PROVIDERNAME      INSTALLEDVERSION      PHASE         │
│ capi-system                     cluster-api              core               cluster-api       v1.7.2                Ready
```

Danil-Grigorev (approved):

PR #594, opened 2024-07-10
test: restructure e2e v1v3 suites

### LINKED ISSUE #477: Extend `v3.Cluster` coverage in E2E

salasberryfin:

Structure E2E in differentiated suites for v3 checks, including adding validations for `embedded-capi-disabled` which so far only has v1 test coverage.

Additionally, this disables tests involving CAPRKE2 + CAPD until we bump CAPI version.

Fixes #477

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/test/e2e/data/capi-operator/capi-providers.yaml
+++ b/test/e2e/data/capi-operator/capi-providers.yaml
@@ -12,7 +12,7 @@ metadata:
 spec:
   name: docker
   type: infrastructure
-  version: v1.7.0
+  version: v1.4.6
   configSecret:
     name: variables
 ---
--- a/test/e2e/suites/embedded-capi-disabled-v3/embedded_capi_disabled_v3_test.go
+++ b/test/e2e/suites/embedded-capi-disabled-v3/embedded_capi_disabled_v3_test.go
@@ -1,0 +1,62 @@
+//go:build e2e
+// +build e2e
+
+/*
+Copyright © 2023 - 2024 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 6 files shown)


Danil-Grigorev (approved):

alexander-demicev (approved):

alexander-demicev:
After some investigation, we found the CAPD is not the source of the error but it still makes sense to keep it with the same CAPI version, maybe we can also include the CAPD version downgrade in this PR.

salasberryfin:
Thanks @alexander-demicev @Danil-Grigorev, I included the downgrade in the pr.

alexander-demicev (approved):

furkatgofurov7:
> Additionally, this disables tests involving CAPRKE2 + CAPD until we bump CAPI version.

Is this really necessary to do? I had an impression we will go down that path if the problem was with CAPD, but might be misunderstood .

alexander-demicev:
@furkatgofurov7 It's probably not mandatory but we can potentially avoid any issues related to version differences. We will bump CAPI to the latest soon and bring back these tests. We would also need to do some other CAPRKE2 tests that use providers like AWS to test something real that users might deploy.

furkatgofurov7 (approved):

PR #597, opened 2024-07-10
🐛 Apply finalizer on rancher cluster to avoid re-create race condition

### LINKED ISSUE #589: Cluster getting re-imported after removal from Rancher

- Install turtles and UI extension
- Import CAPI cluster, v3 cluster gets created in Rancher. Wait for it to be Active
- Remove the imported cluster from Rancher, it gets re-imported. 
Not always reproducible though.

https://github.com/rancher/turtles/assets/73099870/e22a5858-daf8-4adf-8fbf-7939447b7c6c

Is there a race condition to annotate cluster as _imported_ and its removal.
Cluster description after getting re-imported:
```
Name:         cluster1
Namespace:    default
Labels:       app.kubernetes.io/managed-by=Helm
              objectset.rio.cattle.io/hash=d8a48a02f9251fc48f93df5bf362b6004500d2a9
Annotations:  imported: true
              meta.helm.sh/release-name: clusters-clusters
              meta.helm.sh/release-namespace: default
              objectset.rio.cattle.io/id: default-clusters-clusters-cattle-fleet-local-system
API Version:  cluster.x-k8s.io/v1beta1
Kind:         Cluster
Metadata:
  Creation Timestamp:  2024-07-03T12:27:13Z
  Finalizers:
    cluster.cluster.x-k8s.io
  Generation:        2
  Resource Version:  156174
  UID:               6ed57ce2-c6dc-4933-b243-5fb30d57f4d7
Spec:
  Cluster Network:
    Pods:
      Cidr Blocks:
        10.1.0.0/16
    Service Domain:  cluster.local
    Services:
      Cidr Blocks:
        10.10.0.0/16
  Control Plane Endpoint:
    Host:  172.19.0.2
    Port:  6443
  Control Plane Ref:
    API Version:  controlplane.cluster.x-k8s.io/v1beta1
    Kind:         RKE2ControlPlane
    Name:         cluster1-control-plane
    Namespace:    default
  Infrastructure Ref:
    API Version:  infrastructure.cluster.x-k8s.io/v1beta1
    Kind:         DockerCluster
    Name:         cluster1
    Namespace:    default
Status:
  Conditions:
    Last Transition Time:  2024-07-03T12:30:51Z
    Status:                True
    Type:                  Ready
    Last Transition Time:  2024-07-03T12:29:58Z
    Status:                True
    Type:                  ControlPlaneInitialized
    Last Transition Time:  2024-07-03T12:30:51Z
    Status:                True
    Type:                  ControlPlaneReady
    Last Transition Time:  2024-07-03T12:27:59Z
    Status:                True
    Type:                  InfrastructureReady
  Control Plane Ready:     true
  Infrastructure Ready:    true
  Observed Generation:     2
  Phase:                   Provisioned
Events:
  Type    Reason               Age                 From                Message
  ----    ------               ----                ----                -------
  Normal  Provisioning         31m (x2 over 31m)   cluster-controller  Cluster cluster1 is Provisioning
  Normal  InfrastructureReady  30m (x12 over 30m)  cluster-controller  Cluster cluster1 InfrastructureReady is now true
  Normal  Provisioned          30m (x11 over 30m)  cluster-controller  Cluster cluster1 is Provisioned
```

Cluster should not be re-imported after removal.

### How to reproduce it?

Steps mention above

### Rancher Turtles version

v0.9.0

Behavior is mostly seen with v3 cluster, not v1.

### Label(s) to be applied

/kind bug

Danil-Grigorev:

As we are unable to set ownership references from CAPI cluster to a cluster scoped management v3 cluster, we have to manage deletion manually. This comes with a problem of tracking deletion order for rancher/capi cluster, and setting `imported` annotation to prevent re-import scenario.

There is a race condition if removal of management v3 cluster happens too fast - turtles misses the opportunity to set `imported` annotation first and causes immediate re-import of the cluster. To prevent such scenario, we need to set `capicluster.turtles.cattle.io` to the management v3 resource we created at all times, and clean it up within the `imported` annotation handling scope on deletion.

Fixes #589 

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/internal/controllers/import_controller_v3.go
+++ b/internal/controllers/import_controller_v3.go
@@ -150,17 +150,16 @@ func (r *CAPIImportManagementV3Reconciler) Reconcile(ctx context.Context, req ct
 		return ctrl.Result{Requeue: true}, err
 	}
 
+	log = log.WithValues("cluster", capiCluster.Name)
+
 	if capiCluster.ObjectMeta.DeletionTimestamp.IsZero() && !turtlesannotations.HasClusterImportAnnotation(capiCluster) &&
-		!controllerutil.ContainsFinalizer(capiCluster, managementv3.CapiClusterFinalizer) {
-		log.Info("CAPI cluster is imported, adding finalizer")
-		controllerutil.AddFinalizer(capiCluster, managementv3.CapiClusterFinalizer)
+		controllerutil.AddFinalizer(capiCluster, managementv3.CapiClusterFinalizer) {
+		log.Info("CAPI cluster is marked for import, adding finalizer")
 
 		if err := r.Client.Update(ctx, capiCluster); err != nil {
 			return ctrl.Result{}, fmt.Errorf("error adding finalizer: %w", err)
 		}
 	}
-
-	log = log.WithValues("cluster", capiCluster.Name)
 
 	// Wait for controlplane to be ready. This should never be false as the predicates
 	// do the filtering.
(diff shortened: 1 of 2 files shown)


alexander-demicev (approved):
the change makes sense, thank for catching it

internal/controllers/import_controller_v3.go
```diff
@@ -150,18 +150,17 @@ func (r *CAPIImportManagementV3Reconciler) Reconcile(ctx context.Context, req ct
 		return ctrl.Result{Requeue: true}, err
 	}
 
+	log = log.WithValues("cluster", capiCluster.Name)
```
salasberryfin, line 153: Is this a dev log we can remove?

salasberryfin (commented):
Thanks @Danil-Grigorev. Just nit-picking a bit on a log message but LGTM.

internal/controllers/import_controller_v3.go
Danil-Grigorev: No, it is actually adding capiCluster name to the logs context and is useful, as every log message will contain it later. Moved this line up, as in the `if` later the log message would otherwise not get the capiCluster.Name when it’s known.

salasberryfin (approved):

PR #614, opened 2024-07-24
Pin CAAPF version to v0.3.0

### LINKED ISSUE #615: Add fleet addon to the turtles E2E suite
Fleet should be a required dependency for the e2e suite, as we will install necessary bundles with it.

Danil-Grigorev:
Pinned CAAPF to version 0.3.0 and marked it as non-experimental.

Since fleet released v0.10.1-rc.1, but it is not a part of rancher/charts yet, this PR can’t add full e2e suite, which will convert `CNI` installation to the cluster using fleet and enable `hostNetwork: true` setting.

Fixes #615  

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/rancher-turtles/templates/clusterctl-config.yaml
+++ b/charts/rancher-turtles/templates/clusterctl-config.yaml
@@ -50,6 +50,6 @@ data:
 
     # Addon providers
     - name:         "fleet"
-      url:          "https://github.com/rancher-sandbox/cluster-api-addon-provider-fleet/releases/latest/addon-components.yaml"
+      url:          "https://github.com/rancher-sandbox/cluster-api-addon-provider-fleet/releases/v0.3.0/addon-components.yaml"
       type:         "AddonProvider"
 {{- end }}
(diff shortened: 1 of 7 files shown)


alexander-demicev (approved):
🥳

charts/rancher-turtles/templates/clusterctl-config.yaml
```diff
@@ -50,6 +50,6 @@ data:
 
     # Addon providers
     - name:         "fleet"
-      url:          "https://github.com/rancher-sandbox/cluster-api-addon-provider-fleet/releases/latest/addon-components.yaml"
+      url:          "https://github.com/rancher-sandbox/cluster-api-addon-provider-fleet/releases/v0.3.0/addon-components.yaml"
```
mjura, line 53: are you sure that this has to be hardcoded? maybe it can be templated using config.yaml or Chart.yaml

mjura (approved):
only small comment but in overall lgtm

yiannistri (approved):

charts/rancher-turtles/templates/clusterctl-config.yaml
Danil-Grigorev: We can add that. Just wanted for now to get rid of `latest` since the feature is moved out of experimental and the version needs to be pinned. People can still install versions below `v0.3.0`, but it limits the upper bound or when the version is not specified at all. And this bound is shipped and controlled by each turtles release, so we know what to expect from new installs at all times.

PR #448, opened 2024-03-27
🌱 Use calico as alternative to kindnet in e2e tests, and start testing helm chart proxy

Danil-Grigorev:
**What this PR does / why we need it**:
This PR provides alternative to kindnet network setup with calico. In my environment it is needed to make the local tests pass.

Additionally this introduces tests for helm addon provider as means for installing Calico CNI in the child cluster.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits into logical changes
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

PR #654, opened 2024-08-08
🌱 Perform e2e on CAAPF with hostNetwork: true

### LINKED ISSUE #651: Install CNI using Fleet Addon Provider
Rancher added support for fleet `hostNetwork` in alpha charts for version `v2.9.1-alpha1`. This allows to test CNI deployment using fleet agent in the child clusters.

E2E suite can be updated to use `Fleet` `Bundle` which will deploy either `calico` or `kindnet` CNI in the clusters.

Danil-Grigorev:
This PR performs CNI installation using CAAPF in the child clusters, which replaces CRS usage in docker clusters in e2e suite.
This change currently requires alpha version of the rancher chart, with version above or equal to `v2.9.1-alpha1`

Fixes #651 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


Danil-Grigorev:
In order to use CAAPF for deploying calico CNI as a helm chart we need to replicate the behavior for chart templating:
- https://github.com/rancher/turtles/pull/448/files#diff-6977749061d5021a72c12350e8c159bbb9b257dbacba21a1a727ed8029916721R24

There is an issue for that: https://github.com/rancher-sandbox/cluster-api-addon-provider-fleet/issues/32 which needs to be implemented.

furkatgofurov7:
I don't remember us having discussion on why using CRS should be replaced by CAAPF? 
I would appreciate if we could discuss the benefits over the current CNI installation solution and why can't we have both tested in e2e tests, since CRS seem to be still relevant upstream.

test/e2e/README.md
```diff
@@ -95,7 +95,7 @@ The config is located in `test/e2e/config/operator.yaml`.
 Most notable ones:
 ```yaml
 variables:
-  RANCHER_VERSION: "v2.7.5" # Default rancher version to install
+  RANCHER_VERSION: "v2.9.0" # Default rancher version to install
```
furkatgofurov7, line 98: does this need an update in turtles docs pre-requisite table?

Danil-Grigorev: It is in the test/e2e/README.md. This document has to contain documentation on provisioning e2e tests, and it is not maintained by the team. This change you are requesting is probably unrelated to the PR itself.

Danil-Grigorev:
@furkatgofurov7 This change only *adds* CSI provisioning tests. CRS are tested upstream, we don’t need to duplicate this if there is an alternative. The CAAPF provides an alternative method, which does not yet have test coverage. This change adds it within turtles.

Prior approach to CAAPF is usage of CAAPH, in #448 but since Fleet is installed in every rancher cluster, we can use it to perform same functionality as CAAPH.

test/e2e/README.md
furkatgofurov7: It was just a question @Danil-Grigorev, thanks for explanation

furkatgofurov7:
> @furkatgofurov7 This change only _adds_ CSI provisioning tests. CRS are tested upstream, we don’t need to duplicate this if there is an alternative. The CAAPF provides an alternative method, which does not yet have test coverage. This change adds it within turtles.
> 
> Prior approach to CAAPF is usage of CAAPH, in #448 but since Fleet is installed in every rancher cluster, we can use it to perform same functionality as CAAPH.

Alright that makes sense, by PR description I had the feeling that CRS are not relevant at all in turtles and we completely drop its usage. Using CAAPF instead of CAAPH is a good improvement towards wider adoption of CAAPF.

furkatgofurov7 (approved):

alexander-demicev (approved):

PR #658, opened 2024-08-13
⚠️ Bump dependencies: CAPI, Rancher and K8s versions

furkatgofurov7:
**What this PR does / why we need it**:
This PR bumps few important dependencies consumed in turtles to the newer versions. Breakdown of version updates as follows:
- **CAPI** to v1.7.3
- **K8s** to v1.30.3/v1.30.0 (K8s_VERSION/MGMT_K8S_VERSION/CPI_IMAGE_K8S_VERSION /kind)
- **Rancher** to v2.9.0
- **rancher-agent** v2.9.0
- **RKE2** to v1.30.3+rke2r1
-  **cert-manager** to v1.15.2
- Other 'Makefile' dependencies to newer version

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #659 

**Special notes for your reviewer**:

**Checklist**:

- [x] squashed commits into logical changes
- [ ] includes documentation
- [ ] adds unit tests
- [x] adds or updates e2e tests

PR #660, opened 2024-08-14
Add new rancher setting patch in e2e

alexander-demicev:

For testing purposes we need to set agent-tls-mode to system-store. More details on the setting here: https://github.com/rancher/rancher/pull/45684

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/test/e2e/data/rancher/rancher-setting-patch.yaml
+++ b/test/e2e/data/rancher/rancher-setting-patch.yaml
@@ -15,3 +15,9 @@ kind: Setting
 metadata:
   name: eula-agreed
 value: "2023-08-16T20:39:11.062Z"
+---
+apiVersion: management.cattle.io/v3
+kind: Setting
+metadata:
+  name: agent-tls-mode
+value: "system-store"


furkatgofurov7:
@alexander-demicev I checked new setting was introduced in rancher-2.9? And we recently bumped it in turtles e2e https://github.com/rancher/turtles/pull/658. Should we re-trigger e2e tests once again to test it?

alexander-demicev:
@furkatgofurov7 I rebased the PR

Danil-Grigorev (approved):

furkatgofurov7 (approved):
Deferring on CI, but this looks good

Danil-Grigorev:
Can you elaborate more why is this setting needed?

furkatgofurov7:
> Can you elaborate more why is this setting needed?

@Danil-Grigorev https://github.com/rancher/rancher/issues/45628#issuecomment-2246152604 should give an idea

test/e2e/data/rancher/rancher-setting-patch.yaml
```diff
@@ -15,3 +15,9 @@ kind: Setting
 metadata:
   name: eula-agreed
 value: "2023-08-16T20:39:11.062Z"
+---
+apiVersion: management.cattle.io/v3
+kind: Setting
+metadata:
+  name: agent-tls-mode
+value: "system-store"
```
furkatgofurov7, line 23: By reading the linked issue and PR again, it seems we need to set this to "strict" if using Rancher 2.9

Danil-Grigorev:
@furkatgofurov7 but this does not explain why we test this setting in our suite. Since this has to be enabled with turtles on helm chart installation in this case.

test/e2e/data/rancher/rancher-setting-patch.yaml
Danil-Grigorev: `system-store` will surface the error on the cluster, which is better for debugging purposes.

alexander-demicev:
@Danil-Grigorev  Defaulting to "strict" is a breaking change for anyone using ngrok in the way we're used to, when enabled allows only certificates verified with CAcerts authority provided by `cacerts` rancher setting will be used.
This is something that Rancher users have to configure on their own and shouldn't be related to how turtles import clusters. I believe that by using `system-store` we are not testing new functionality and using the setup we had before.
Here are a couple links to related PRs:
https://github.com/rancher/fleet/pull/2507
https://github.com/rancher/rancher/pull/45909

I'm not against extending the E2E suite to use "strict" mode for all our suites but it can done later separately.

Danil-Grigorev:
I'm confused @alexander-demicev as I was not advocating for using strict mode, see https://github.com/rancher/turtles/pull/660#discussion_r1718214400

So if you saying that strict became new default, and by applying it we re-enables use of ngrok, change makes sense to me.

(Maybe this needs to be mentioned in PR description?)

alexander-demicev:
@Danil-Grigorev We will need to provide CA public key for ngrok(or anything else) if using strict, which shouldn't be a problem for us or users but it can be addressed separately

PR #663, opened 2024-08-15
Disable 'rancher-kubeconfigs' feature gate in charts

furkatgofurov7:
'rancher-kubeconfigs' FG introduced in https://github.com/rancher/turtles/pull/178 was meant to test CAPI >=v1.5.x in turtles at the time Rancher was not ready yet. However, with the newer Rancher version (v2.9.x using CAPI v1.7.x) it should be possible to test newer version of CAPI without this feature gate enabled. Moreover, `update-labels` test [suite](https://github.com/rancher/turtles/blob/main/test/e2e/suites/update-labels/suite_test.go#L166) this FG by explicitly setting it to true anyways, but by default we should not expose it as enabled in charts.

Additionally, it is a follow-up to #658 to bump leftover dependencies to newer versions.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/rancher-turtles/values.yaml
+++ b/charts/rancher-turtles/values.yaml
@@ -8,12 +8,12 @@ rancherTurtles:
   features:
     cluster-api-operator:
       cleanup: true
-      kubectlImage: registry.k8s.io/kubernetes/kubectl:v1.28.0
+      kubectlImage: registry.k8s.io/kubernetes/kubectl:v1.30.0
     embedded-capi:
       disabled: true
     rancher-webhook:
       cleanup: true
-      kubectlImage: registry.k8s.io/kubernetes/kubectl:v1.28.0
+      kubectlImage: registry.k8s.io/kubernetes/kubectl:v1.30.0
     rancher-kubeconfigs:
       label: true
     managementv3-cluster:
--- a/charts/rancher-turtles/values.yaml
+++ b/charts/rancher-turtles/values.yaml
@@ -15,7 +15,7 @@ rancherTurtles:
       cleanup: true
       kubectlImage: registry.k8s.io/kubernetes/kubectl:v1.30.0
     rancher-kubeconfigs:
-      label: true
+      label: false
     managementv3-cluster:
       enabled: true
     managementv3-cluster-migration:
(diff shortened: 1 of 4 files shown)


yiannistri (approved):

Danil-Grigorev:
@furkatgofurov7 After thinking about I’m under the impression we need to do the opposite of what this PR suggests. Always apply label on the secrets, and never disable it.

charts/rancher-turtles/values.yaml
```diff
@@ -8,14 +8,14 @@ rancherTurtles:
...
-      kubectlImage: registry.k8s.io/kubernetes/kubectl:v1.28.0
+      kubectlImage: registry.k8s.io/kubernetes/kubectl:v1.30.0
     embedded-capi:
       disabled: true
     rancher-webhook:
       cleanup: true
-      kubectlImage: registry.k8s.io/kubernetes/kubectl:v1.28.0
+      kubectlImage: registry.k8s.io/kubernetes/kubectl:v1.30.0
     rancher-kubeconfigs:
-      label: true
```
Danil-Grigorev, line 18: We can remove this option entirely, to avoid situation when the kubeconfig might not get labeled.

furkatgofurov7:
> @furkatgofurov7 After thinking about I’m under the impression we need to do the opposite of what this PR suggests. Always apply label on the secrets, and never disable it.
> 
> Unless there is a confidence that existing rancher code performs this task instead of turtles already. If you have a link, then it lgtm

@Danil-Grigorev here you go:

- FG was introduced and merged in 11/2023: https://github.com/rancher/turtles/pull/178
- Rancher issue: https://github.com/rancher/rancher/issues/42333
- Rancher 2.9 PR mered in 1/2024: https://github.com/rancher/rancher/pull/43897

Danil-Grigorev (approved):

furkatgofurov7:
@alexander-demicev what are your thoughts on this, do you foresee any problems with disabling this FG by default?

alexander-demicev:
@furkatgofurov7 It should be fine to disable it

alexander-demicev (approved):

PR #672, opened 2024-08-19
chore: update lower/upper rancher version in chart

salasberryfin:

As we've now moved to newer version of CAPI (and related dependencies) https://github.com/rancher/turtles/commit/1a31ad4f04ea30ab0d2edff76c819c9bbfc2373d and this now requires Rancher v2.9 (as stated in the docs https://github.com/rancher/turtles-docs/pull/150) this change aligns the chart's lower/upper version bounds for Rancher. 

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/rancher-turtles/Chart.yaml
+++ b/charts/rancher-turtles/Chart.yaml
@@ -23,7 +23,7 @@ annotations:
   catalog.cattle.io/namespace: rancher-turtles-system
   catalog.cattle.io/os: linux
   catalog.cattle.io/permits-os: linux
-  catalog.cattle.io/rancher-version: '>= 2.8.1-0'
+  catalog.cattle.io/rancher-version: '>= 2.9.0-1'
   catalog.cattle.io/release-name: rancher-turtles
   catalog.cattle.io/scope: management
   catalog.cattle.io/type: cluster-tool


alexander-demicev (approved):

charts/rancher-turtles/Chart.yaml
```diff
@@ -23,7 +23,7 @@ annotations:
   catalog.cattle.io/namespace: rancher-turtles-system
   catalog.cattle.io/os: linux
   catalog.cattle.io/permits-os: linux
-  catalog.cattle.io/rancher-version: '>= 2.8.1-0'
+  catalog.cattle.io/rancher-version: '>= 2.9.0-1'
```
furkatgofurov7, line 26: `2.9.0` is rancher version, but what does `-1` define in this format?

furkatgofurov7 (approved):

charts/rancher-turtles/Chart.yaml
salasberryfin: This comes from the ADR that defines these annotations and specifies that there needs to be a lower and upper bound:

> There are two annotations: `catalog.cattle.io/rancher-version` and `catalog.cattle.io/kube-version` which must include lower and upper bounds.

Danil-Grigorev (approved):

PR #679, opened 2024-08-20
✨ Check CAPI provider latest version daily

### LINKED ISSUE #673: Implement automatic upgrade of provider versions

As an operator, I’d like to ensure my CAPIProviders are always at latests, so that there is no need to perform version bumps manually.

### Detailed Description

Since turtles gradually bumps versions of providers, we need to ensure that providers without explicitly specified versions are able to upgrade automatically.

_No response_

### Label(s) to be applied

/kind feature

Danil-Grigorev:
- Perform check that CAPIProvider is up-to-date with current latest daily. Longer delay is set to prevent hitting github rate limits.

This change performs daily check for provider upgrades, and performs the actual upgrade automatically, if the `CAPIProvider` version is not set. If it is set, the version is user managed so we don’t upgrade this automatically.

Once upstream (capi-operator) has a condition to notify about new version availability, this daily check can be changed to rely on the status condition.

This change ensures we don’t leave existing users on older versions of providers.

Fixes #673 

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/api/v1alpha1/conditions_consts.go
+++ b/api/v1alpha1/conditions_consts.go
@@ -28,6 +28,9 @@ const (
 	// RancherCredentialSourceMissing occures when a source credential secret is missing.
 	RancherCredentialSourceMissing = "RancherCredentialSourceMissing"
 
-	// LastAppliedConfigurationTime is set as a timestamp infor of the last configuration update byt the CAPI Operator resource.
+	// LastAppliedConfigurationTime is set as a timestamp info of the last configuration update byt the CAPI Operator resource.
 	LastAppliedConfigurationTime = "LastAppliedConfigurationTime"
+
+	// CheckLatestVersionTime is set as a timestamp info of the last timestamp of the latest version being up-to-date for the CAPIProvider.
+	CheckLatestVersionTime = "CheckLatestVersionTime"
 )
--- a/internal/sync/provider_sync.go
+++ b/internal/sync/provider_sync.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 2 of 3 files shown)


salasberryfin (approved):
LGMT but I'm wondering what would happen if a latest version breaks something and user needs to roll back to previous one. Would they need to manually specify a version so this check is not performed?

alexander-demicev (commented):
In my opinion the default behavior should stay as it is today and automatic updates can be enabled with a separate API field.

Danil-Grigorev:
@alexander-demicev This can be addressed in the followup pr.

furkatgofurov7:
Does this try to avoid the need for bumping versions of core CAPI and providers manually (i.e in go.mod)? When CAPI is 
released, let's say 1.8.0, not all providers, release the compatible version to that of CAPI. Would that mean we will test CAPI 1.8.0 and CAPI Provider complaint with CAPI 1.7?

alexander-demicev:
@Danil-Grigorev sure, in that case let's merge this PR after the release, and as @furkatgofurov7 mentioned we would also need to pin e2e suite dependencies if want to test the same CAPI version as now

alexander-demicev:
@Danil-Grigorev Let's take a bit more time to think about this feature, just to avoid any issues before the release

Danil-Grigorev:
@alexander-demicev This change is required if https://github.com/rancher/turtles/pull/672 is planned to shipped this release. Otherwise users will have to update manually to 1.7.x, and there is a room for error. (Side note: this needs to be documented, that version bump needs to be manual).

alexander-demicev:
Manual updates after the Turtles upgrade are ok, automatic updates of CAPI providers should be an optional feature

alexander-demicev:
@Danil-Grigorev would like to make the feature optional in this PR or in a follow-up? We can merge it like it is now if you want to do it in a follow-up

Danil-Grigorev:
Yes please, I'd like to make a follow up PR @alexander-demicev as already mentioned

alexander-demicev (approved):

PR #705, opened 2024-08-30
🌱 Add option to skip rancher pre-requisites

### LINKED ISSUE #627: issue during installation when rancher is not ready

To reproduce you can do:
- Helm chart to install rancher
- Helm chart to install rancher-turtles

To reproduce you have to launch both helm chart at the same time, because rancher will take more time, and the rancher turtles fails because some CRD are not present.
```
+ helm_v3 install --namespace rancher-turtles-system --create-namespace --version 0.1.0+up0.9.1 --set-string global.clusterCIDR=10.42.0.0/16 --set-string global.clusterCIDRv4=10.42.0.0/16 --set-string global.clusterDNS=10.43.0.10 --set-string global.clusterDomain=cluster.local --set-string global.rke2DataDir=/var/lib/rancher/rke2 --set-string global.serviceCIDR=10.43.0.0/16 rancher-turtles /tmp/rancher-turtles.tgz
Error: INSTALLATION FAILED: failed pre-install: unable to build kubernetes object for deleting hook rancher-turtles/templates/pre-install-job.yaml: resource mapping not found for name: "embedded-cluster-api" namespace: "" from "": no matches for kind "Feature" in version "management.cattle.io/v3"
ensure CRDs are installed first
+ exit
```

The expected will be:
- Rancher-turtles should wait to be installed. Helm will use the retries to try to install it again once CRD and resources from rancher are created.
- but the problem is that a clean-up process starts in rancher-turtles, and the cleanup process fails due to:
```
rancher-turtles-system            rancher-capiprovider-cleanup-bwwd5                       0/1     Error              0               7m14s
rancher-turtles-system            rancher-capiprovider-cleanup-jq6cn                       0/1     Error              0               11m
rancher-turtles-system            rancher-capiprovider-cleanup-lgk5t                       0/1     Error              0               9m55s
rancher-turtles-system            rancher-capiprovider-cleanup-n8rs4                       0/1     Error              0               113s
rancher-turtles-system            rancher-capiprovider-cleanup-rbj72                       0/1     Error              0               12m
rancher-turtles-system            rancher-capiprovider-cleanup-tnn5j                       0/1     Error              0               11m
rancher-turtles-system            rancher-capiprovider-cleanup-xtdlk                       0/1     Error              0               12m
```
and the logs for that pods shows:
```
mgmt-cluster-network:~ # k logs rancher-capiprovider-cleanup-rbj72 -n rancher-turtles-system
error: the server doesn't have a resource type "capiproviders"
```
so the installation cannot be uninstalled.

- After timeout, hel cleanup pods disapears, and the helm-install-rancher-turtle pod is in crashlookbackoff:
```
mgmt-cluster-network:~ # helm uninstall rancher-turtles --namespace rancher-turtles-system
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /etc/rancher/rke2/rke2.yaml
WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /etc/rancher/rke2/rke2.yaml
Error: 1 error occurred:
	* timed out waiting for the condition
```
and the following retry to install rancher-turtles the logs shows:
```
+ helm_v3 install --namespace rancher-turtles-system --create-namespace --version 0.1.0+up0.9.1 --set-string global.clusterCIDR=10.42.0.0/16 --set-string global.clusterCIDRv4=10.42.0.0/16 --set-string global.clusterDNS=10.43.0.10 --set-string global.clusterDomain=cluster.local --set-string global.rke2DataDir=/var/lib/rancher/rke2 --set-string global.serviceCIDR=10.43.0.0/16 rancher-turtles /tmp/rancher-turtles.tgz
Error: INSTALLATION FAILED: cannot re-use a name that is still in use
```
because the previous one could not be deleted with the cleanup process.

### How to reproduce it?

- helm install rancher
- helm install rancher-turtles

### Rancher Turtles version

0.9.1

_No response_

### Label(s) to be applied

/kind bug

Danil-Grigorev:
This change should allow to install turtles in development scenario without dependency on rancher, for the expense of added requirement to patch the `embedded-cluster-api` feature gate manually.

This can be done by specifying `--set rancherTurtles.rancherInstalled=false` flag.

Deletion will need to be performed according to https://turtles.docs.rancher.com/getting-started/uninstall_turtles

Fixes #627

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/rancher-turtles/templates/pre-install-job.yaml
+++ b/charts/rancher-turtles/templates/pre-install-job.yaml
@@ -1,4 +1,5 @@
 {{- if index .Values "rancherTurtles" "features" "embedded-capi" "disabled" }}
+{{- if index .Values "rancherTurtles" "rancherInstalled"}}
 ---
 apiVersion: management.cattle.io/v3
 kind: Feature
@@ -9,6 +10,7 @@ metadata:
     "helm.sh/hook-weight": "1"
 spec:
   value: false
+{{- end }}
 {{- end }}
 {{- if index .Values "rancherTurtles" "features" "rancher-webhook" "cleanup" }}
 ---
--- a/charts/rancher-turtles/values.yaml
+++ b/charts/rancher-turtles/values.yaml
@@ -5,6 +5,7 @@ rancherTurtles:
   namespace: rancher-turtles-system
   managerArguments: []
   imagePullSecrets: []
+  rancherInstalled: true
   features:
     cluster-api-operator:
       cleanup: true
--- a/internal/controllers/capiprovider_controller.go
+++ b/internal/controllers/capiprovider_controller.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	operatorv1 "sigs.k8s.io/cluster-api-operator/api/v1alpha2"
@@ -96,8 +97,9 @@ func (r *CAPIProviderReconciler) patchStatus(ctx context.Context, capiProvider *
 }
 
 // SetupWithManager sets up the controller with the Manager.
-func (r *CAPIProviderReconciler) SetupWithManager(_ context.Context, mgr ctrl.Manager) (err error) {
+func (r *CAPIProviderReconciler) SetupWithManager(_ context.Context, mgr ctrl.Manager, options controller.Options) (err error) {
 	b := ctrl.NewControllerManagedBy(mgr).
+		WithOptions(options).
 		For(&turtlesv1.CAPIProvider{})
 
 	resources := []client.Object{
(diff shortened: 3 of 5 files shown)


hardys:
Thanks for working on this @Danil-Grigorev - the change looks good to me, although it's now clear we could potentially use the existing `features.embedded-capi` API to achieve the same thing, unless we expect any other dependencies on Rancher in the pre-install phase.

Note for other reviewers the use-case behind #627 is we want to deploy Rancher and Rancher Turtles via the [RKE2 helm controller](https://docs.rke2.io/helm) - in this situation ordering of the helm installs cannot be controlled, so you end up reliant on retries, so we must ensure the turtles chart can either install, or fail and be uninstalled, in the case where Rancher is not yet installed.

A related issue is in the pre-delete hook we assume the `capiproviders` CRD exists, and the hook fails otherwise - IMO for a robust fix we should consider making the pre-delete hook fail gracefully in this scenario (e.g check if the crd exists, and exit without any error if it's missing), but that could be considered as a follow-up perhaps.

alexander-demicev (approved):

salasberryfin (dismissed):

furkatgofurov7 (commented):
@Danil-Grigorev can we also update the turtles docs regarding this, since current docs have rancher listed as pre-requisite and has a dedicated chapter for installing it before Turtles? I believe this provides a fix for the specific use case edge team has, but I suggest we need to be super clear on what are the consequences if this option is enabled.

furkatgofurov7 (approved):

alexander-demicev (approved):

salasberryfin (approved):

PR #708, opened 2024-08-30
Add EtcdMachineSnapshot controller

### LINKED ISSUE #499: [etcdsnapshot&restore] Implement ETCDMachineSnapshot controller
For implementation details please refer to the proposal and POC code: https://github.com/furkatgofurov7/turtles-etcd-restore/blob/main/internal/controller/etcdmachinebackup_controller.go

furkatgofurov7:

Fixes #499 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


exp/etcdrestore/controllers/etcdmachinesnapshot_controller.go
```diff
@@ -60,18 +67,224 @@ func (r *EtcdMachineSnapshotReconciler) SetupWithManager(_ context.Context, mgr
...
+	ctx context.Context, etcdMachineSnapshot *snapshotrestorev1.EtcdMachineSnapshot,
+) (ctrl.Result, error) {
+	log := log.FromContext(ctx)
+
+	// Handle different phases of the etcdmachinesnapshot creation process
+	switch etcdMachineSnapshot.Status.Phase {
+	case "":
+		// Initial phase, set to Pending
+		etcdMachineSnapshot.Status.Phase = snapshotrestorev1.ETCDSnapshotPhasePending
+		if err := r.Client.Status().Update(ctx, etcdMachineSnapshot); err != nil {
```
alexander-demicev, line 112: Can we use Patch instead of updating here?

exp/etcdrestore/controllers/etcdmachinesnapshot_controller.go
```diff
@@ -60,18 +67,224 @@ func (r *EtcdMachineSnapshotReconciler) SetupWithManager(_ context.Context, mgr
...
+			return ctx.Err()
+		}
+	}
+	return errors.New("operation failed after maximum retries")
+}
+
+// checkSnapshotStatus checks the status of the snapshot creation process.
+func checkSnapshotStatus(ctx context.Context, r *EtcdMachineSnapshotReconciler, etcdMachineSnapshot *snapshotrestorev1.EtcdMachineSnapshot) error {
+	log := log.FromContext(ctx)
+	etcdSnapshotFileList := &unstructured.UnstructuredList{}
```
alexander-demicev, line 219: we should now use a wrapper library instead of unstructured

exp/etcdrestore/controllers/etcdmachinesnapshot_controller.go
```diff
@@ -60,18 +67,224 @@ func (r *EtcdMachineSnapshotReconciler) SetupWithManager(_ context.Context, mgr
...
+		}
+		return ctrl.Result{}, nil
+	case snapshotrestorev1.ETCDSnapshotPhaseFailed:
+		// Handle snapshot creation failure
+		log.Error(nil, "Snapshot creation failed")
+		// Perform retry logic with exponential backoff
+		initialDelay := 1 * time.Second
+		maxRetries := 5
+
+		err := retryWithExponentialBackoff(ctx, initialDelay, maxRetries, func(ctx context.Context) error {
```
alexander-demicev, line 138: maybe we can let the reconciler return an error instead? This way we'll use controller runtime to handle backoff.

alexander-demicev (changes requested):
I have a couple of small comments, thanks for setting up the skeleton

exp/etcdrestore/controllers/etcdmachinesnapshot_controller.go
```diff
@@ -60,18 +67,224 @@ func (r *EtcdMachineSnapshotReconciler) SetupWithManager(_ context.Context, mgr
...
+		// Initial phase, set to Pending
+		etcdMachineSnapshot.Status.Phase = snapshotrestorev1.ETCDSnapshotPhasePending
+		if err := r.Client.Status().Update(ctx, etcdMachineSnapshot); err != nil {
+			log.Error(err, "Failed to update EtcdMachineSnapshot status to Pending")
+			return ctrl.Result{}, err
+		}
+	case snapshotrestorev1.ETCDSnapshotPhasePending:
+		// Transition to Running
+		etcdMachineSnapshot.Status.Phase = snapshotrestorev1.ETCDSnapshotPhaseRunning
+		if err := r.Client.Status().Update(ctx, etcdMachineSnapshot); err != nil {
```
Danil-Grigorev, line 119: It should be possible to use Patch here as well. Best upstream (CAPI) practice to writing controllers uses `defer` to perform patch on modified resources at the end of the full reconcile cycle.

exp/etcdrestore/controllers/etcdmachinesnapshot_controller.go
Danil-Grigorev: Yes, reconcile should never be blocking. It is always possible to just requeue. But I think returning error is not a good option, if this error is transient, as it might result in hitting very long wait periods, if the error persists for too long.

`RequeueAfter` IMO should be sufficient, unless this is the final error state, in which case returning a custom error LGTM too.

alexander-demicev: I agree, it depends on how we want to handle it, in my opinion, we want to retry applying the system agent plan, and if it fails increase the timeout for reapplying to let users figure out possible issues.

exp/etcdrestore/controllers/etcdmachinesnapshot_controller.go
alexander-demicev: We can only set the phase here and return, `Reconcile()` will patch the object and get to the next phase in the next reconciliation cycle. This way we will also signal to the users the exact status of the snapshot.

exp/etcdrestore/controllers/etcdmachinesnapshot_controller.go
```diff
@@ -60,18 +67,224 @@ func (r *EtcdMachineSnapshotReconciler) SetupWithManager(_ context.Context, mgr
 //+kubebuilder:rbac:groups=turtles-capi.cattle.io,resources=etcdmachinesnapshots/finalizers,verbs=update
 
 // Reconcile reconciles the EtcdMachineSnapshot object.
-func (r *EtcdMachineSnapshotReconciler) Reconcile(_ context.Context, _ ctrl.Request) (ctrl.Result, error) {
-	return ctrl.Result{}, nil
+func (r *EtcdMachineSnapshotReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := log.FromContext(ctx)
+
+	etcdMachineSnapshot := &snapshotrestorev1.EtcdMachineSnapshot{ObjectMeta: metav1.ObjectMeta{
+		Name:      req.Name,
```
alexander-demicev, line 74: no need to setup the name and namespace here, it will be filled out later by Get() request.

exp/etcdrestore/controllers/etcdmachinesnapshot_controller.go
furkatgofurov7: Updated to use `RequeueAfter`

exp/etcdrestore/controllers/etcdmachinesnapshot_controller.go
furkatgofurov7: Updated, PTAL

exp/etcdrestore/controllers/etcdmachinesnapshot_controller.go
```diff
@@ -60,18 +68,193 @@ func (r *EtcdMachineSnapshotReconciler) SetupWithManager(_ context.Context, mgr
...
+) (ctrl.Result, error) {
+
+	patchBase := client.MergeFromWithOptions(etcdMachineSnapshot.DeepCopy(), client.MergeFromWithOptimisticLock{})
+
+	// Handle different phases of the etcdmachinesnapshot creation process
+	switch etcdMachineSnapshot.Status.Phase {
+	case "":
+		// Initial phase, set to Pending
+		etcdMachineSnapshot.Status.Phase = snapshotrestorev1.ETCDSnapshotPhasePending
+		if err := r.Client.Status().Patch(ctx, etcdMachineSnapshot, patchBase); err != nil {
```
alexander-demicev, line 129: can we simply return instead and let the resource be patched at the top-level?

exp/etcdrestore/controllers/etcdmachinesnapshot_controller.go
```diff
@@ -60,18 +68,193 @@ func (r *EtcdMachineSnapshotReconciler) SetupWithManager(_ context.Context, mgr
...
+			return nil
+		}
+	}
+
+	// If no matching ready snapshot is found, return an error to retry
+	return errors.New("snapshot not ready")
+}
+
+// extractFieldsFromETCDSnapshotFile extracts the location, nodeName, snapshotName, and readyToUse fields from an ETCDSnapshotFile resource.
+func extractFieldsFromETCDSnapshotFile(snapshotFile k3sv1.ETCDSnapshotFile) (string, string, string, bool, error) {
```
alexander-demicev, line 235: can this function be dropped now since we can access fields directly? Alternatively, it can be changed to only  validating the EtcdSnapshotFile.

furkatgofurov7: Changed the function to validate the EtcdSnapshotFile as suggested to check the fields before directly accessing them. Thanks for suggestion

exp/etcdrestore/controllers/etcdmachinesnapshot_controller.go
furkatgofurov7: yeah, that is simpler, updated the change accordingly

exp/etcdrestore/controllers/etcdmachinesnapshot_controller.go
```diff
@@ -60,18 +68,193 @@ func (r *EtcdMachineSnapshotReconciler) SetupWithManager(_ context.Context, mgr
...
+) (ctrl.Result, error) {
+
+	// Handle different phases of the etcdmachinesnapshot creation process
+	switch etcdMachineSnapshot.Status.Phase {
+	case "":
+		// Initial phase, set to Pending
+		etcdMachineSnapshot.Status.Phase = snapshotrestorev1.ETCDSnapshotPhasePending
+	case snapshotrestorev1.ETCDSnapshotPhasePending:
+		// Transition to Running
+		etcdMachineSnapshot.Status.Phase = snapshotrestorev1.ETCDSnapshotPhaseRunning
```
alexander-demicev, line 129:
```suggestion
		etcdMachineSnapshot.Status.Phase = snapshotrestorev1.ETCDSnapshotPhaseRunning
		return ctrl.Result{}, nil
```

exp/etcdrestore/controllers/etcdmachinesnapshot_controller.go
```diff
@@ -60,18 +68,193 @@ func (r *EtcdMachineSnapshotReconciler) SetupWithManager(_ context.Context, mgr
...
-	_ context.Context, _ *snapshotrestorev1.EtcdMachineSnapshot,
-) (_ ctrl.Result, err error) {
+	ctx context.Context, etcdMachineSnapshot *snapshotrestorev1.EtcdMachineSnapshot,
+) (ctrl.Result, error) {
+
+	// Handle different phases of the etcdmachinesnapshot creation process
+	switch etcdMachineSnapshot.Status.Phase {
+	case "":
+		// Initial phase, set to Pending
+		etcdMachineSnapshot.Status.Phase = snapshotrestorev1.ETCDSnapshotPhasePending
```
alexander-demicev, line 126:
```suggestion
		etcdMachineSnapshot.Status.Phase = snapshotrestorev1.ETCDSnapshotPhasePending
		return ctrl.Result{}, nil
```

exp/etcdrestore/controllers/etcdmachinesnapshot_controller.go
```diff
@@ -60,18 +68,193 @@ func (r *EtcdMachineSnapshotReconciler) SetupWithManager(_ context.Context, mgr
...
+		return ctrl.Result{RequeueAfter: snapshotPhaseRequeueDuration}, nil
+	case snapshotrestorev1.ETCDSnapshotPhaseDone:
+		// Snapshot creation is complete, no further action needed
+		return ctrl.Result{}, nil
+	}
+
+	patchBase := client.MergeFromWithOptions(etcdMachineSnapshot.DeepCopy(), client.MergeFromWithOptimisticLock{})
+
+	// Patch the resource at the top-level
+	if err := r.Client.Status().Patch(ctx, etcdMachineSnapshot, patchBase); err != nil {
```
alexander-demicev, line 149:
```suggestion
	
```

alexander-demicev: no need to patch here if the object will be patched at the top-level Reconcile()

exp/etcdrestore/controllers/etcdmachinesnapshot_controller.go
```diff
@@ -60,18 +68,193 @@ func (r *EtcdMachineSnapshotReconciler) SetupWithManager(_ context.Context, mgr
...
+				log.Error(err, "Failed to delete associated EtcdMachineSnapshot", "snapshotName", snapshot.Name)
+				return ctrl.Result{}, err
+			}
+			log.Info("Deleted associated EtcdMachineSnapshot", "snapshotName", snapshot.Name)
+		}
+	}
+
+	// Remove the finalizer to allow deletion of the EtcdMachineSnapshot
+	controllerutil.RemoveFinalizer(etcdMachineSnapshot, snapshotrestorev1.ETCDMachineSnapshotFinalizer)
+	if err := r.Client.Update(ctx, etcdMachineSnapshot); err != nil {
```
alexander-demicev, line 189: same as above, let's switch to Patch

exp/etcdrestore/controllers/etcdmachinesnapshot_controller.go
```diff
@@ -60,18 +68,193 @@ func (r *EtcdMachineSnapshotReconciler) SetupWithManager(_ context.Context, mgr
...
+	// Example: Delete associated snapshot resources
+	snapshotList := &snapshotrestorev1.EtcdMachineSnapshotList{}
+	if err := r.Client.List(ctx, snapshotList, client.InNamespace(etcdMachineSnapshot.Namespace)); err != nil {
+		log.Error(err, "Failed to list associated EtcdMachineSnapshots")
+		return ctrl.Result{}, err
+	}
+
+	for _, snapshot := range snapshotList.Items {
+		if snapshot.Spec.MachineName == etcdMachineSnapshot.Spec.MachineName {
+			if err := r.Client.Delete(ctx, &snapshot); err != nil {
```
alexander-demicev, line 179: we should probably wait until deleted before removing finalizer

exp/etcdrestore/controllers/etcdmachinesnapshot_controller.go
```diff
@@ -60,18 +68,193 @@ func (r *EtcdMachineSnapshotReconciler) SetupWithManager(_ context.Context, mgr
...
+		// Extract fields directly after validation
+		snapshotName := snapshotFile.Spec.SnapshotName
+		readyToUse := *snapshotFile.Status.ReadyToUse
+
+		// Check if the snapshot is ready to use and matches the machine snapshot name
+		if readyToUse && snapshotName == etcdMachineSnapshot.Name {
+			// Update the status to Done
+			etcdMachineSnapshot.Status.Phase = snapshotrestorev1.ETCDSnapshotPhaseDone
+			if err := r.Client.Status().Patch(ctx, etcdMachineSnapshot, patchBase); err != nil {
+				return fmt.Errorf("failed to update EtcdMachineSnapshot status to Done: %w", err)
```
alexander-demicev, line 231: same as other comment, it can be patched at Reconcile()

alexander-demicev (approved):
We can merge the PR as it is and improve in follow-ups

Danil-Grigorev (approved):

Issue #721, opened 2024-09-04
Verify users RBAC when creating a snapshot/restore


alexander-demicev:
We need to verify users' RBAC permissions on the cluster when creating a snapshot/restore. Users must have admin rights for the cluster they are creating snapshot/restore. This can be done by creating SAR(https://kubernetes.io/docs/reference/kubernetes-api/authorization-resources/self-subject-access-review-v1/) and checking for '*' verb.

PR #732, opened 2024-09-09
Add webhook that validated users RBAC rules

alexander-demicev:

Add a validating webhook for ETCD snapshot and restore that validates users permissions for the referenced cluster object.

Fixes https://github.com/rancher/turtles/issues/721

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/rancher-turtles/templates/rancher-turtles-exp-etcdrestore-components.yaml
+++ b/charts/rancher-turtles/templates/rancher-turtles-exp-etcdrestore-components.yaml
@@ -509,6 +509,13 @@ rules:
   - update
   - watch
 - apiGroups:
+  - authorization.k8s.io
+  resources:
+  - subjectaccessreviews
+  verbs:
+  - create
+  - get
+- apiGroups:
   - bootstrap.cluster.x-k8s.io
   resources:
   - rke2configs
@@ -843,5 +850,57 @@ webhooks:
     - UPDATE
     resources:
     - rke2configs
+  sideEffects: None
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  annotations:
+    cert-manager.io/inject-ca-from: rancher-turtles-system/rancher-turtles-etcdsnapshotrestore-serving-cert
+  labels:
+    turtles-capi.cattle.io: etcd-restore
+  name: rancher-turtles-etcdsnapshotrestore-validating-webhook-configuration
+webhooks:
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: rancher-turtles-etcdsnapshotrestore-webhook-service
+      namespace: rancher-turtles-system
+      path: /validate-turtles-capi-cattle-io-v1alpha1-etcdmachinesnapshot
+  failurePolicy: Fail
+  matchPolicy: Equivalent
+  name: etcdmachinesnapshot.kb.io
+  rules:
+  - apiGroups:
+    - turtles-capi.cattle.io
+    apiVersions:
+    - v1alpha1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - etcdmachinesnapshots
+  sideEffects: None
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: rancher-turtles-etcdsnapshotrestore-webhook-service
+      namespace: rancher-turtles-system
+      path: /validate-turtles-capi-cattle-io-v1alpha1-etcdsnapshotrestore
+  failurePolicy: Fail
+  matchPolicy: Equivalent
+  name: etcdsnapshotrestore.kb.io
+  rules:
+  - apiGroups:
+    - turtles-capi.cattle.io
+    apiVersions:
+    - v1alpha1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - etcdsnapshotrestores
   sideEffects: None
 {{- end }}
--- a/exp/etcdrestore/webhooks/etcdsnapshotrestore.go
+++ b/exp/etcdrestore/webhooks/etcdsnapshotrestore.go
@@ -1,0 +1,92 @@
+/*
+Copyright © 2023 - 2024 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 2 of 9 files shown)


exp/etcdrestore/webhooks/etcdsnapshotrestore.go
```diff
@@ -0,0 +1,92 @@
...
+		For(&snapshotrestorev1.EtcdSnapshotRestore{}).
+		WithValidator(r).
+		Complete()
+}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type.
+func (r *EtcdMachineSnapshotRestoreWebhook) ValidateCreate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+	logger := log.FromContext(ctx)
+
+	logger.Info("Validating EtcdMachineSnapshot")
```
furkatgofurov7, line 56:
```suggestion
	logger.Info("Validating EtcdSnapshotRestore")
```

charts/rancher-turtles/templates/rancher-turtles-exp-etcdrestore-components.yaml
```diff
@@ -844,4 +851,56 @@ webhooks:
...
+    resources:
+    - etcdmachinesnapshots
+  sideEffects: None
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: rancher-turtles-etcdsnapshotrestore-webhook-service
+      namespace: rancher-turtles-system
+      path: /validate-turtles-capi-cattle-io-v1alpha1-etcdmachinesnapshotrestore
```
furkatgofurov7, line 891: can we please align naming of the CRD everywhere: etcdmachinesnapshotrestore -> etcdsnapshotrestore

exp/etcdrestore/webhooks/etcdsnapshotrestore.go
```diff
@@ -0,0 +1,92 @@
...
+
+func (r *EtcdMachineSnapshotRestoreWebhook) validateSpec(ctx context.Context, etcdSnapshotRestore *snapshotrestorev1.EtcdSnapshotRestore) (admission.Warnings, error) {
+	var allErrs field.ErrorList
+
+	if etcdSnapshotRestore.Spec.ClusterName == "" {
+		allErrs = append(allErrs, field.Required(field.NewPath("spec.clusterName"), "clusterName is required"))
+	}
+
+	if len(allErrs) > 0 {
+		return nil, apierrors.NewInvalid(snapshotrestorev1.GroupVersion.WithKind("EtcdMachineSnapshot").GroupKind(), etcdSnapshotRestore.Name, allErrs)
```
furkatgofurov7, line 84:
```suggestion
		return nil, apierrors.NewInvalid(snapshotrestorev1.GroupVersion.WithKind("EtcdSnapshotRestore").GroupKind(), etcdSnapshotRestore.Name, allErrs)
```

exp/etcdrestore/webhooks/rbac_test.go
```diff
@@ -0,0 +1,106 @@
...
+		Expect(validateRBAC(admission.NewContextWithRequest(ctx, admission.Request{
+			AdmissionRequest: admissionv1.AdmissionRequest{
+				UserInfo: authenticationv1.UserInfo{
+					Username: "clusteradmin",
+				},
+			},
+		}), cl, "test-cluster", namespace)).To(Succeed())
+	})
+
+	It("should fail if user is not to access the cluster", func() {
```
furkatgofurov7, line 88:
```suggestion
	It("should fail if user is not allowed to access the cluster", func() {
```

furkatgofurov7 (changes requested):

exp/etcdrestore/webhooks/rbac.go
```diff
@@ -0,0 +1,60 @@
...
+
+func validateRBAC(ctx context.Context, cl client.Client, clusterName, clusterNamespace string) error {
+	admissionRequest, err := admission.RequestFromContext(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get admission request from context: %w", err)
+	}
+
+	sar := authv1.SubjectAccessReview{
+		Spec: authv1.SubjectAccessReviewSpec{
+			ResourceAttributes: &authv1.ResourceAttributes{
```
Danil-Grigorev, line 37: Doesn’t this need a check for access to `Machine` as well? Maybe there needs to be a user group a user needs to belong to and validated against.

alexander-demicev: If the user is allowed to administer `Cluster` object, can we assume they have access to machines? `SubjectAccessReview` should handle all validation on API server side, we provide it all information we have like user name, group and resource we'd like to validate against.
I'd like to avoid doing multiple SARs if possible assuming that there might be a lot of ETCDMachineSnapshots.

furkatgofurov7 (approved):
Thanks!

exp/etcdrestore/webhooks/rbac.go
Danil-Grigorev: My take on this is to have a `etcd-snapshot-restore` cluster role, which a user has to be a part of to perform operation. This does not mean making more `SAR` requests.

Alternatively, what prevents here to check for `Machine` object in this request too?

alexander-demicev: We started by thinking of including custom RBAC verbs like "etcd-snapshot" or "etcd-restore" and then checking for them in SAR. Then we decided it easier to restrict etcd snapshot/restore to admin users initially and it can be easily changed later if we get any requests from users.

alexander-demicev: For more context, a verb in `Role/ClusterRole` can be any word for example:
```
  verbs: ["get", "list", "watch", "create", "update", "patch", "delete", "mycustomverb"]
```
It's up to us later, how to handle it in webhooks/controllers. For example, we might want more granular RBAC rules for some resources and allow only specific field change for some users in a resource

Danil-Grigorev: It is a new pattern, I was not aware of. My suggestion was to stick to classic RBAC management roles, where a user may be `admin` or `view` or a member of a custom `backup`, `backuprestore` group. It is an alternative, which came to me since I was not a part of previous discussion.

Makes sense to me.

Danil-Grigorev (approved):

PR #735, opened 2024-09-09
Replace LocalObjectReference usage with strings in exp/etcdrestore API 

### LINKED ISSUE #729: Replace local object references with strings
Replace local object references with strings, this will simplify the API as we won't depend on external structures.

furkatgofurov7:
Additionally, currently when running `make build-chart & release-chart` Makefile targets throw:

```
/Users/userX/repos/rancher-turtles/hack/tools/bin/helm-v3.15.3 package out/charts/rancher-turtles --app-version=est/v0.11.0 --version=est/v0.11.0 --destination=out/package
Error: Invalid Semantic Version
make: *** [build-chart] Error 1
```
error and that is because of how we construct `HELM_CHART_TAG` variable. In this patch, the extraction command was updated to use `grep` with a regular expression to ensure it extracts a valid semantic version (e.g., v0.11.0) and both helm package commands run as expected. 

Fixes #729 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


Danil-Grigorev (approved):
Thanks!

alexander-demicev (approved):

PR #742, opened 2024-09-12
Implement ClusterctlConfig with controller

Danil-Grigorev:

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


alexander-demicev (approved):

furkatgofurov7:
Implements: https://github.com/rancher/turtles/blob/main/docs/adr/0012-clusterctl-provider.md

internal/controllers/clusterctlconfig_controller.go
```diff
@@ -0,0 +1,147 @@
...
+		client.FieldOwner("clusterctl-controller"),
+	}...)
+	if err != nil {
+		return fmt.Errorf("creating ClusterctlConfig default ConfigMap: %w", err)
+	}
+
+	return nil
+}
+
+//+kubebuilder:rbac:groups=turtles-capi.cattle.io,resources=clusterctlcofigs,verbs=get;list;watch;patch
```
furkatgofurov7, line 90: I don't see this and below CRDs with listed RBAC rules in role.yaml?

internal/controllers/clusterctlconfig_controller.go
```diff
@@ -0,0 +1,147 @@
...
+//+kubebuilder:rbac:groups=turtles-capi.cattle.io,resources=clusterctlcofigs/status,verbs=get;list;watch;patch
+//+kubebuilder:rbac:groups=turtles-capi.cattle.io,resources=clusterctlcofigs/finalizers,verbs=get;list;watch;patch;update
```
furkatgofurov7, line 92: okay, this is the reason:
```suggestion
//+kubebuilder:rbac:groups=turtles-capi.cattle.io,resources=clusterctlconfigs,verbs=get;list;watch;patch
//+kubebuilder:rbac:groups=turtles-capi.cattle.io,resources=clusterctlconfigs/status,verbs=get;list;watch;patch
//+kubebuilder:rbac:groups=turtles-capi.cattle.io,resources=clusterctlconfigs/finalizers,verbs=get;list;watch;patch;update
```

internal/controllers/clusterctlconfig_controller.go
```diff
@@ -0,0 +1,147 @@
...
+//+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;patch
+
+// Reconcile reconciles the EtcdMachineSnapshot object.
+func (r *ClusterctlConfigReconciler) Reconcile(ctx context.Context, req reconcile.Request) (ctrl.Result, error) {
+	log := log.FromContext(ctx)
+
+	configMap := clusterctl.Config()
```
furkatgofurov7, line 99: should not we check for nil after this?

internal/controllers/clusterctlconfig_controller.go
Danil-Grigorev: CAPI provider and internal/controllers with exception of etcd snapshot/restore do not use this kubebuilder generation, to keep the RBAC clean and updated on manual basis. This, if enabled will make duplicate roles, which can be done separately from this PR, as it will need a review.

internal/controllers/clusterctlconfig_controller.go
Danil-Grigorev: This uses the `embeded` state of the `configMap`, so this config map can’t be nil, it is a static memory address in the binary.

Danil-Grigorev:
I’d like to follow up with the changes in the next PR, ensuring `CAPIProvider` will maintain latest version from the in-memory state.

furkatgofurov7 (approved):
Last [comment](https://github.com/rancher/turtles/pull/742#discussion_r1766774105) that needs resolving, otherwise LGTM

internal/controllers/clusterctlconfig_controller.go
Danil-Grigorev: Yes, https://github.com/rancher/turtles/pull/742/files#diff-e4dcbd6018dd457ef50ee530496d1b8c45380f28ed8cbf097d9c8d9ce3025883R3381-R3382 is set to correct value, annotations needs to be fixed, will do in the followup

Issue #765, opened 2024-10-01
Remove etcd snapshot controller during Turtles uninstall


salasberryfin:
### What steps did you take and what happened?

etcd snapshot controller is not cleaned up when Turtles is uninstalled. Helm will successfully delete Turtles resources and mark removal as complete but etcd snapshot controller remains active and running.

### What did you expect to happen?

etcd snapshot controller should be removed together with all other Turtles components.

### How to reproduce it?

With Turtles installed and etcd snapshot controller enabled, simply uninstall Turtles (either via Rancher Dashboard or directly via Helm).

### Label(s) to be applied

/kind bug

PR #775, opened 2024-10-09
🌱 Add a way to run ETCDRestore controller locally

furkatgofurov7:
This patch provides an alternative way to run Rancher Turtles locally using the turtles-dev script and Tilt. A new `USE_TILT_DEV` env var was introduced (set to true by default to keep previous behaviour), and a new way of deploying Turtles uses the locally built helm chart only when set to `false`. 
Additionally, image building for new `turtles-etcd-snapshot-restore` was taken care by Tilt when used, however it is not when there is no Tilt. To run it without tilt, new Dockerfile to build the image for etcdrestore controller and use the built image in local testing. 

Lastly, it changes nightly image publish GH action to publish etcdrestore image as part of the GHA.

Fixes: https://github.com/rancher/turtles/issues/765

When running uninstall locally with these changes, snapshot restore controller was cleaned up together with RT controller without any issues

**Checklist**:

- [x] includes documentation
- [ ] adds unit tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -349,7 +349,7 @@ docker-pull-prerequisites:
 docker-build-etcdrestore: buildx-machine docker-pull-prerequisites ## Build docker image for a specific architecture
 ## reads Dockerfile from stdin to avoid an incorrectly cached Dockerfile (https://github.com/moby/buildkit/issues/1368)
 	# buildx does not support using local registry for multi-architecture images
-	cat ./exp/etcdrestore/Dockerfile | DOCKER_BUILDKIT=1 BUILDX_BUILDER=$(MACHINE) docker buildx build \
+	cat $(EXP_ETCDRESTORE_DIR)/Dockerfile | DOCKER_BUILDKIT=1 BUILDX_BUILDER=$(MACHINE) docker buildx build \
 			--platform $(ARCH) \
 			--load \
 			--build-arg builder_image=$(GO_CONTAINER_IMAGE) \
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -341,6 +345,32 @@ docker-pull-prerequisites:
 	docker pull $(GO_CONTAINER_IMAGE)
 	docker pull gcr.io/distroless/static:latest
 
+.PHONY: docker-build-etcdrestore ## Build the docker image for etcdrestore
+docker-build-etcdrestore: buildx-machine docker-pull-prerequisites ## Build docker image for a specific architecture
+## reads Dockerfile from stdin to avoid an incorrectly cached Dockerfile (https://github.com/moby/buildkit/issues/1368)
+	# buildx does not support using local registry for multi-architecture images
+	cat ./exp/etcdrestore/Dockerfile | DOCKER_BUILDKIT=1 BUILDX_BUILDER=$(MACHINE) docker buildx build \
```
salasberryfin, line 352: Thinking about when the time comes to get `etcdrestore` out of `./exp`. I'd say we could have a variable with the path to `etcdrestore` source code and it'll simplify things going forward. What do you think?

```suggestion
   ETCDRESTORE_DIR ?= ./exp/etcdrestore

	cat $(ETCDRESTORE_DIR)/Dockerfile | DOCKER_BUILDKIT=1 BUILDX_BUILDER=$(MACHINE) docker buildx build \
```
and we could apply this to other occurrences of this path.

salasberryfin (commented):
Thanks for taking care of this @furkatgofurov7. I just suggested a minor change, but it is not a blocker at all, let me know what you think. Otherwise, it looks good to me.

alexander-demicev (approved):

salasberryfin (approved):

Issue #693, opened 2024-08-27
Switch some test cases to ClusterClass provisioning


salasberryfin:
### What would you like to be added (User Story)?

As we work on improving `ClusterClass` support and user experience, as a developer I'd like to have test cases covering cluster provisioning via `ClusterClass`.

### Detailed Description

E2E tests should validate that cluster can be provisioned via `ClusterClass` + `Cluster` definition.

### Anything else you would like to add?

To prevent us from a growing lists of E2E test cases, we should move some of the existing scenarios to using a `ClusterClass` provisioning approach. This prevents us from increasing the E2E runtime which is already considerably long.

salasberryfin:
We can use this issue to also create more cluster class example to share with UI team.

salasberryfin:
- [ ] AWS
- [x] Azure - [coverage for AKS clusters](https://github.com/rancher/turtles/pull/779)

valaparthvi:
I would like to work on CAPD+Kubeadm part, if that is alright?

kkaempf:
All checkmark items and sub-issues are done, closing.

PR #779, opened 2024-10-10
test: cover azure provisioning via cluster class

salasberryfin:

Cover AKS cluster provisioning via cluster topology. I suggest deleting the "standard" provisioning test case as the resulting workload cluster is equivalent and the same validation logic is applied to Turtles import/un-import functionality.

Partially solves #693  

We can remove the existing AKS test case and use cluster class only in this PR or do it later in a separate one.

The new template will be formatted and re-used as one of the cluster topology example that will be added to Turtles docs. A documentation issue is tracking progress on this [here](https://github.com/rancher/turtles-docs/issues/161).

Class definition includes a reference to RKE2 because **CAPI's webhooks require a bootstrap object reference** which, in this case, won't affect during provisioning as AKS is a managed service. This is a known limitation of CAPI and you can refer to https://github.com/kubernetes-sigs/cluster-api/issues/10943 for detailed information on whether this can be fixed and how.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


salasberryfin:
Updated cluster template to reference RKE2 bootstrap object.

alexander-demicev (approved):

Danil-Grigorev (approved):

furkatgofurov7 (approved):

PR #788, opened 2024-10-18
🐛 Remove requirement to specify CAPIProvider secret

### LINKED ISSUE #787: CAPIProvider: secret is not assigned by default

Create a minimal `CAPIProvider` template:
```yaml
apiVersion: turtles-capi.cattle.io/v1alpha1
kind: CAPIProvider
metadata:
  name: azure
  namespace: capz-system
spec:
  type: infrastructure
```

Default set of FG to be set. None are set by default, due to generated secret not being referenced in the underlying `InfrastructurProvider` resource, created from the supplied template.

For each provider (at the time of issue creation) the default set consists of:
```yaml
EXP_CLUSTER_RESOURCE_SET: true
CLUSTER_TOPOLOGY: true
EXP_MACHINE_POOL: true
```

Each value can be overridden with `spec.features` or `spec.variables`, but requiring these values to be always present is creating inconsistencies with e2e scenarios and real world cluster template examples. A user is not supposed to read `test/e2e` templates first, to create a provider and get a `Cluster`.

### How to reproduce it?

Always

### Rancher Turtles version

_No response_

_No response_

### Label(s) to be applied

/kind bug

Danil-Grigorev:
This change improves the process of `CAPIProvider` creation. Users don’t need to specify `spec.configSecret` at all times, in order to get default set of enabled features, such as ClusterClasses or MachinePools, + provider specific features.

Fixes #787

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/internal/sync/azure_provider_sync.go
+++ b/internal/sync/azure_provider_sync.go
@@ -67,6 +67,12 @@ func NewAzureProviderSync(cl client.Client, capiProvider *turtlesv1.CAPIProvider
 
 	capiProvider.SetSpec(spec)
 
+	if capiProvider.Spec.Variables == nil {
+		capiProvider.Spec.Variables = map[string]string{}
+	}
+
+	capiProvider.Spec.Variables["EXP_AKS_RESOURCE_HEALTH"] = "true"
+
 	return &ProviderSync{
 		DefaultSynchronizer: NewDefaultSynchronizer(cl, capiProvider, template),
 		Destination:         destination,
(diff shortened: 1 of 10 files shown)


salasberryfin:
Thanks @Danil-Grigorev. If using this "minimal" `CAPIProvider` yaml and, for example, wanting to enable Cluster Classes, is it not required to pass the `CUSTER_TOPOLOGY` environment variable to the provider?

Danil-Grigorev:
@salasberryfin Yes, there is a test case for that also https://github.com/rancher/turtles/pull/788/files#diff-4f7dd9c7662f229992f65de2f22cd082cb79ba691af9062fed41ae5e472180c7R97

alexander-demicev (approved):

salasberryfin (approved):

PR #790, opened 2024-10-21
test: disable docker/rke2 e2e scenario

### LINKED ISSUE #755: Stabilize nightly E2E tests
### Description

Nightly E2E tests have been failing for a few days now, example: https://github.com/rancher/turtles/actions/runs/11024535042. We need to investigate what's the root of the issue and fix it.

salasberryfin:

Nightly E2E tests are systematically failing and the common issue is the Docker+RKE2 test case. This may be related to some memory limits (@Danil-Grigorev noticed something similar in short E2E test executions) because I've had the tests passing locally in multiple tries.

For now, I suggest we disable this specific scenario, as **it is adding noise to our nightly tests which we no longer can rely on for detecting unexpected failures**. Let's see if disabling this improves the health of these tests. I'd say there's no harm in disabling this, as testing CAPD scenarios is not that relevant in nightly E2E.

Fixes #755

My suggestion is to try to stabilize tests and, once we accomplish this, let's focus on cleaning up test suites that are becoming less relevant. For example:
- reduce (or even remove) v1 import validation
- rethink if it is required to test the embedded capi disabled scenario (this should be covered by any of the other scenarios).

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/test/e2e/suites/import-gitops-v3/import_gitops_v3_test.go
+++ b/test/e2e/suites/import-gitops-v3/import_gitops_v3_test.go
@@ -62,7 +62,7 @@ var _ = Describe("[Docker] [Kubeadm] - [management.cattle.io/v3] Create and dele
 	})
 })
 
-var _ = Describe("[Docker] [RKE2] - [management.cattle.io/v3] Create and delete CAPI cluster functionality should work with namespace auto-import", Label(e2e.FullTestLabel), func() {
+var _ = Describe("[Docker] [RKE2] - [management.cattle.io/v3] Create and delete CAPI cluster functionality should work with namespace auto-import", Label(e2e.LocalTestLabel), func() {
 	BeforeEach(func() {
 		komega.SetClient(setupClusterResult.BootstrapClusterProxy.GetClient())
 		komega.SetContext(ctx)


test/e2e/suites/import-gitops-v3/import_gitops_v3_test.go
```diff
@@ -62,7 +62,7 @@ var _ = Describe("[Docker] [Kubeadm] - [management.cattle.io/v3] Create and dele
 	})
 })
 
-var _ = Describe("[Docker] [RKE2] - [management.cattle.io/v3] Create and delete CAPI cluster functionality should work with namespace auto-import", Label(e2e.FullTestLabel), func() {
+var _ = Describe("[Docker] [RKE2] - [management.cattle.io/v3] Create and delete CAPI cluster functionality should work with namespace auto-import", Label(e2e.DontRunLabel), func() {
```
Danil-Grigorev, line 65:
```suggestion
var _ = Describe("[Docker] [RKE2] - [management.cattle.io/v3] Create and delete CAPI cluster functionality should work with namespace auto-import", Label(e2e.LocalTestLabel), func() {
```
Should be a better option. WDYT?

salasberryfin: Fair point, I'll switch to local.

Danil-Grigorev (approved):

furkatgofurov7 (approved):

PR #796, opened 2024-10-21
feat: bump provider versions with updatecli

### LINKED ISSUE #795: Automate provider version bumps
# Description

After creating a forked version of CAPI projects (core, providers, etc.) we now need to take care of publishing mirror releases ourselves. While looking at https://github.com/rancher/turtles/issues/785, there should be a mechanism to simplify the process and automate as many steps as possible. Turtles maintains a list of URLs pointing to all pre-built providers in a configuration file [here](https://github.com/rancher/turtles/blob/main/internal/controllers/clusterctl/config.yaml). We can use [updatecli](https://github.com/rancher/turtles/blob/main/internal/controllers/clusterctl/config.yaml) to automate creation of version bump PRs.

## Options
- Easiest solution is to schedule executions of this automation and let it check what the latest release is for each of the projects. `updatecli` is smart enough to then only generate a PR if there are changes to the configuration file.
- We could also test the possibility of using a dispatch webhook so events in forked repositories can trigger actions in Turtles to run `updatecli`.

salasberryfin:

This PR adds a new `updatecli` GitHub Action that is executed daily to check for new releases of the CAPI projects we ship with Turtles. If a new release is found that does not match the one specified in `./internal/controllers/clusterctl/config.yaml`, it will update it and create a PR with all the changes.

Fixes #795 

With this workflow, the idea is that we can use the existing GitHub Actions from the forked repositories to create new releases, then in the scheduled execution of the `updatecli` action, it will create a PR to bump the version in the configuration file.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/updatecli/updatecli.d/manifest.yaml
+++ b/updatecli/updatecli.d/manifest.yaml
@@ -1,0 +1,165 @@
+name: Bump CAPI providers version
+
+# define scm for turtles
+scms:
+  turtles:
+    kind: github
+    spec:
+      user: turtles-bot
+      email: turtles@suse.de
+      owner: rancher
+      repository: turtles
+      token: '{{ requiredEnv "UPDATECLI_GITHUB_TOKEN" }}'
+      username: '{{ requiredEnv "UPDATECLI_GITHUB_ACTOR" }}'
+      branch: main
+
+# retrieve latest provider release
+sources:
+  capirelease:
+    kind: githubrelease
+    name: Get the latest core CAPI release
+    spec:
+      owner: "rancher-sandbox"
+      repository: "cluster-api"
+      token: '{{ requiredEnv "UPDATECLI_GITHUB_TOKEN" }}'
+      username: '{{ requiredEnv "UPDATECLI_GITHUB_ACTOR" }}'
+      typeFilter:
+        latest: true
+  caprke2release:
+    kind: githubrelease
+    name: Get the latest CAPI RKE2 bootstrap/control plane provider release
+    spec:
+      owner: "rancher"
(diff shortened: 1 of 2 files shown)


updatecli/updatecli.d/manifest.yaml
```diff
@@ -0,0 +1,165 @@
+name: Bump CAPI providers version
+
+# define scm for turtles
+scms:
+  turtles:
+    kind: github
+    spec:
+      user: turtles-bot
+      email: turtles@suse.de
```
furkatgofurov7, line 9: is this bot already exists?

furkatgofurov7 (commented):
Looks good, although interested to know if there was a way to test this somehow beforehand

updatecli/updatecli.d/manifest.yaml
salasberryfin: I tested this with my own fork of Turtles and using upstream CAPI repositories as source for new releases. It's not the exact same scenario but it's the closest I could test before merging. You can take a look at a sample PR `updatecli` created in my fork [here](https://github.com/salasberryfin/rancher-turtles/pull/10). I did not have to create a bot beforehand.

furkatgofurov7: Thanks @salasberryfin, it looks good to me, although PR was created using GHA bot? I thought turtles-bot would be normal GH user created by use and part of the org with right permission to push changes.

Also, I have one follow-up question: https://github.com/salasberryfin/rancher-turtles/pull/10/files#r1810287894

Danil-Grigorev (approved):
Thanks!

PR #834, opened 2024-10-31
feat: opt out of cluster owner rbac

salasberryfin:

Preparing for new Rancher release that will include a feature that allows to bypass setting a creator owner on the resource. Import controller now adds the annotation `field.cattle.io/no-creator-rbac` on any Rancher clusters it reconciles. This will apply to newly created clusters and to existing ones that are subject to reconciliation.

You can take a look at https://github.com/rancher/rancher/pull/47259, where this new option is introduced.

Fixes #

Also updates E2E and import controller tests to check that clusters are properly annotated.

**After this is merged, we'll have to wait until next Rancher minor release to consider this as done**

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/internal/controllers/helpers.go
+++ b/internal/controllers/helpers.go
@@ -53,6 +53,7 @@ const (
 	v1ClusterMigrated         = "cluster-api.cattle.io/migrated"
 
 	defaultRequeueDuration = 1 * time.Minute
+	trueAnnotationValue    = "true"
 )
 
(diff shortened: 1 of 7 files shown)


furkatgofurov7 (approved):
Overall looks good, thanks for taking care of it, LGTM. 

@JonCrowther possibly, if you could also take a look that would be great.

JonCrowther (approved):
LGTM

PR #841, opened 2024-11-05
[etcd-snapshot-restore] Aggregated ETCD snapshot

### LINKED ISSUE #764: [etcdsnapshot&restore] Create a single automatic snapshot per cluster
To reduce number of created snapshots per cluster, we can encapsulate multiple automatic snapshots for a cluster in a single resource within the resource status, instead of creating a separate one per each automatic snapshot.

This should address scalability concerns and provide a better UX working with automatic RKE2 snapshots.

Context from proposal:

> Just an idea here: potentially this number can be decreased from N*M to just N where N is the number of clusters and M is the ~number of preserved snapshot per each cluster, if the automatic snapshot will be treated as a “full cluster” state. Instead of creating a resource per each snapshot resource, there will be one snapshot resource synced to the management cluster. Automation will take care of updating name of the node and file/S3 reference, where the snapshot has taken place inside this single resource, and potentially specifying alternative up-to-date snapshots in a status, which can be used as a source for restore.

Danil-Grigorev:
From #764:

> To reduce number of created snapshots per cluster, we can encapsulate multiple automatic snapshots for a cluster in a single resource within the resource status, instead of creating a separate one per each automatic snapshot.

This allows to maintain an up-to-date per-cluster snapshot list in the management cluster, with sufficient information to request a restore.

Fixes #764 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


charts/rancher-turtles/templates/rancher-turtles-exp-etcdrestore-components.yaml
```diff
@@ -45,77 +45,51 @@ spec:
...
               machineName:
                 type: string
             required:
             - clusterName
-            - configRef
-            - location
-            - machineName
             type: object
             x-kubernetes-validations:
             - message: ETCD snapshot location can't be empty.
```
salasberryfin, line 56:
```suggestion
            - message: cluster name can't be empty.
```
Shouldn't this be referencing cluster name?

exp/etcdrestore/api/v1alpha1/etcdmachinesnapshot_types.go
```diff
@@ -39,21 +37,35 @@ const (
 	ETCDMachineSnapshotFinalizer = "etcdmachinesnapshot.turtles.cattle.io"
 )
 
-// +kubebuilder:validation:XValidation:message="ETCD snapshot location can't be empty.",rule="size(self.location)>0"
+// +kubebuilder:validation:XValidation:message="ETCD snapshot location can't be empty.",rule="size(self.clusterName)>0"
```
salasberryfin, line 40:
```suggestion
// +kubebuilder:validation:XValidation:message="cluster name can't be empty.",rule="size(self.clusterName)>0"
```

salasberryfin (commented):
Only a couple of comments. Otherwise, LGTM

furkatgofurov7 (approved):
LGTM, except the review comments by @salasberryfin earlier.

charts/rancher-turtles/templates/rancher-turtles-exp-etcdrestore-components.yaml
Danil-Grigorev: This CRD needs redesign, which will be addressed in a separate PR. I’ll update validation message there too.

PR #843, opened 2024-11-06
Remove specific versions in dev script

Danil-Grigorev:

Dev script requires specific versions for CAPI providers.

This PR proposes removing specific version requirements. Current pinned version (v1.7.7) of CAPRKE2 is not compatible with turtles, as agent install command causes continuous remediation of CP machines. This problem was fixed with latest.

Additionally, a pinned version requires regular bumps with `updatecli`, which increases complexity of the automation, focused only on maintaining dev environment.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/scripts/turtles-dev.sh
+++ b/scripts/turtles-dev.sh
@@ -48,10 +48,10 @@ export CLUSTER_TOPOLOGY=true
 
 helm install capi-operator capi-operator/cluster-api-operator \
     --create-namespace -n capi-operator-system \
-    --set infrastructure=docker:v1.7.7 \
-    --set core=cluster-api:v1.7.7 \
-    --set controlPlane=rke2:v0.8.0 \
-    --set bootstrap=rke2:v0.8.0 \
+    --set infrastructure=docker \
+    --set core=cluster-api \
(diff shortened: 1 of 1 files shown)


yiannistri (approved):

salasberryfin (approved):

scripts/turtles-dev.sh
```diff
@@ -48,10 +48,10 @@ export CLUSTER_TOPOLOGY=true
 
 helm install capi-operator capi-operator/cluster-api-operator \
     --create-namespace -n capi-operator-system \
-    --set infrastructure=docker:v1.7.7 \
-    --set core=cluster-api:v1.7.7 \
-    --set controlPlane=rke2:v0.8.0 \
-    --set bootstrap=rke2:v0.8.0 \
+    --set infrastructure=docker \
```
furkatgofurov7, line 51: the only objective I would have is: with this change, we start installing CAPI v1.8.5 and CAPRKE2 v0.8.0 where latter is compatible only with CAPI v1.7.x series

PR #860, opened 2024-11-19
Add image overrides for CAPI/CAPRKE2 images

furkatgofurov7:

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/internal/controllers/clusterctl/config.yaml
+++ b/internal/controllers/clusterctl/config.yaml
@@ -54,3 +54,12 @@ data:
     - name:         "fleet"
       url:          "https://github.com/rancher-sandbox/cluster-api-addon-provider-fleet/releases/v0.3.1/addon-components.yaml"
       type:         "AddonProvider"
+
+    # Image overrides
+    images:
+      cluster-api:
+        repository: "registry.suse.com/rancher"
+      controlplane-rke2:
+        repository: "registry.suse.com/rancher"
+      bootstrap-rke2:
+        repository: "registry.suse.com/rancher"


internal/controllers/clusterctl/config.yaml
```diff
@@ -54,3 +54,15 @@ data:
     - name:         "fleet"
       url:          "https://github.com/rancher-sandbox/cluster-api-addon-provider-fleet/releases/v0.3.1/addon-components.yaml"
       type:         "AddonProvider"
+
+    # Image overrides
+    images:
+    - name: "cluster-api"
+      repository: "https://github.com/rancher-sandbox/cluster-api"
+      tag:        "v1.7.7"
```
hardys, line 62: FYI I figured out why the `CoreProvider` override is not enough in my environment - we're following the [airgapped docs](https://turtles.docs.rancher.com/turtles/v0.13/en/getting-started/air-gapped-environment.html) and creating configmaps from the upstream manifests, we need to update the docs (and our [edge airgap resources chart](https://github.com/suse-edge/charts/tree/main/charts/rancher-turtles-airgap-resources)) to align with the forks instead

```
 # kubectl get cm -n capi-system core-cluster-api-v1.7.7 -o yaml | grep image:
            image: registry.k8s.io/cluster-api/cluster-api-controller:v1.7.7
# kubectl get deployment -n capi-system -o yaml | grep image:
          image: registry.k8s.io/cluster-api/cluster-api-controller:v1.7.7
```

hardys: Actually, it seems that CM is created even in the non-airgapped scenario so my analysis above may be incorrect, I still think we need to align the airgapped docs though?

Danil-Grigorev: Can you please check if the configMap with overrides is mounted into CAPI operator pod? If it is not, then this will not effective (meaning `clusterctl-config.yaml` configMap). May happen if capi operator is not installed as a dependency of turtles within chart settings.

You don’t need to update your generated air-gap configmaps if you only need an image to point to the registry override we are specifying here. This will be performed by regular [image overrides](https://cluster-api.sigs.k8s.io/clusterctl/configuration.html?highlight=overrid#image-overrides) functionality implemented upstream (with a [fix](https://github.com/kubernetes-sigs/cluster-api-operator/pull/616) recently, so it *should* work as long as CM is mounted) and applied to all versions of manifests, even the air-gapped ones you have.

furkatgofurov7: > Actually, it seems that CM is created even in the non-airgapped scenario so my analysis above may be incorrect, I still think we need to align the airgapped docs though?

https://github.com/rancher/turtles-docs/pull/177

furkatgofurov7:
Superseded by: https://github.com/rancher/turtles/pull/874

PR #890, opened 2024-12-02
Remove Elemental from supported providers list

anmazzotti:

The Elemental CAPI provider is going to take more incubation time. 
The team decided to remove it from the supported provider list just in case.

Note that in the [documentation](https://github.com/rancher/turtles-docs/blob/main/docs/v0.13/modules/en/pages/reference-guides/providers/certified.adoc#list-of-providers-in-experimental-mode), the Elemental provider was already removed from the table, so nothing more to do there.


kkaempf (approved):
:+1:

yiannistri (approved):

PR #908, opened 2024-12-05
Prevent infinite CAPIProvider reconciliations

### LINKED ISSUE #892: Unit tests fail with a flaky error
Unit tests fail with a flaky error - https://github.com/rancher/turtles/actions/runs/12116411500/job/33776720577

anmazzotti:

The CAPIProvider syncing secret is getting a new `resourceVersion` even on empty patch. This creates an infinite loop where the CAPIProviderReconciler is continuously invoked due to the secret version change.

This PR tries to address this, however further issues escalate after applying it. 
For example there is a conflict between the `rancher-turtles-controller-manager` and the `cluster-api-operator` when it comes to managing some resources. 
For example giving the following bootstrap, there is a conflict in how finalizers are handled:

```yaml
kind: BootstrapProvider
metadata:
  annotations:
    operator.cluster.x-k8s.io/applied-spec-hash: d637e476cb631c683729765c7f6f3fd8c04b5c314254b114aa9500b51b7ed8f1 --> Emptied/Re-filled on a race condition
  creationTimestamp: "2024-12-04T14:21:54Z" 
  finalizers:
  - foregroundDeletion
  - provider.cluster.x-k8s.io --> Deleted/Re-added on a race condition
  generation: 1
  name: rke2
  namespace: default
  ownerReferences:
  - apiVersion: turtles-capi.cattle.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: CAPIProvider
    name: rke2-bootstrap
    uid: bc2f1c0c-9f32-4bab-9859-e18098e9f82d
  resourceVersion: "246785"
  uid: 58a94292-7858-4449-afa1-82edfdbf701c
spec:
  configSecret:
    name: rke2-bootstrap
    namespace: default
  version: v0.9.0
status:
  conditions:
  - lastTransitionTime: "2024-12-04T14:22:38Z"
    reason: MinimumReplicasAvailable
    status: "True"
    type: Ready
  - lastTransitionTime: "2024-12-04T14:22:28Z"
    status: "True"
    type: PreflightCheckPassed
  - lastTransitionTime: "2024-12-04T14:22:27Z"
    status: "True"
    type: ProviderInstalled
  contract: v1beta1
  installedVersion: v0.9.0
  observedGeneration: 1

```

Fixes #892

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/internal/sync/client.go
+++ b/internal/sync/client.go
@@ -51,12 +51,11 @@ func Patch(ctx context.Context, cl client.Client, obj client.Object, options ...
 	log.Info(fmt.Sprintf("Updating %s: %s", obj.GetObjectKind().GroupVersionKind().Kind, client.ObjectKeyFromObject(obj)))
 
 	patchOptions := []client.PatchOption{
-		client.ForceOwnership,
 		client.FieldOwner(fieldOwner),
 	}
 	patchOptions = append(patchOptions, options...)
 
-	return cl.Patch(ctx, obj, client.Apply, patchOptions...)
+	return cl.Patch(ctx, obj, client.Merge, patchOptions...)
 }
 
 // PatchStatus will only patch the status subresource of the provided object.
@@ -72,8 +71,7 @@ func PatchStatus(ctx context.Context, cl client.Client, obj client.Object) error
 
 	log.Info(fmt.Sprintf("Patching status %s: %s", obj.GetObjectKind().GroupVersionKind().Kind, client.ObjectKeyFromObject(obj)))
 
-	return cl.Status().Patch(ctx, obj, client.Apply, []client.SubResourcePatchOption{
-		client.ForceOwnership,
+	return cl.Status().Patch(ctx, obj, client.Merge, []client.SubResourcePatchOption{
 		client.FieldOwner(fieldOwner),
 	}...)
 }
--- a/internal/sync/client.go
+++ b/internal/sync/client.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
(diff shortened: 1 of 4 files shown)


internal/sync/client.go
```diff
@@ -50,8 +49,7 @@ func Patch(ctx context.Context, cl client.Client, obj client.Object) error {
 
 	log.Info(fmt.Sprintf("Updating %s: %s", obj.GetObjectKind().GroupVersionKind().Kind, client.ObjectKeyFromObject(obj)))
 
-	return cl.Patch(ctx, obj, client.Apply, []client.PatchOption{
-		client.ForceOwnership,
```
anmazzotti, line 54: ForceOwnership does not work with Merge patch

internal/sync/core.go
```diff
@@ -78,8 +79,24 @@ func (s *DefaultSynchronizer) Apply(ctx context.Context, reterr *error) {
...
+			return
+		}
+	}
+
+	finalizers = append(finalizers, metav1.FinalizerDeleteDependents)
+	obj.SetFinalizers(finalizers)
+}
+
 func setOwnerReference(owner, obj client.Object) {
-	obj.SetFinalizers([]string{metav1.FinalizerDeleteDependents})
```
anmazzotti, line 82: This kept removing the `provider.cluster.x-k8s.io` finalized added by the `cluster-api-provider`

Danil-Grigorev: Isn’t this change alone enough to fix the issue?

anmazzotti: Not this alone, there is another instance where we set the finalizers, that is in `patchStatus`.

Also we are experiencing multiple conflicts, this solves the issue with the *Providers resources managed by the cluster-api-operator, however the other commit that changes `client.Apply` to `client.Merge` is to prevent all resources from having a new `resourceVersion` on empty patches. This does apply to all syncers, providers, secrets, etc.

Also the reason I kept commits separate.

Danil-Grigorev (approved):

internal/sync/core.go
Danil-Grigorev: It seems the tests are more stable with this change. I opened a separate [one](https://github.com/rancher/turtles/pull/916) to address infinite reconciliation issue with secrets, which is an issue with comparing the diff.

The only issue with this workflow is that we will cause resourceVersion bumps each time we patch CAPIProvider resource or operator resource. SSA was ment to avoid that, but it works [incorrectly](https://kubernetes.io/docs/concepts/configuration/secret/#basic-authentication-secret)  with `stringData` field. Therefore they have a diff comparison. Now the problem is on regular resources, where it is desirable to perform a diff check on the resource content before sending patch request.

Danil-Grigorev (approved):
There seems to be no regression in tests which verify RV stability on the secret resource, and SSA patch is notoriously unpredictable. I can also see one of the flakes in https://github.com/rancher/turtles/blob/a45c388c10213f31ba0b1b4c9dbc739e6dc77648/internal/sync/provider_sync_test.go#L350 no longer appearing in local runs.

LGTM

alexander-demicev (approved):
thanks!

PR #939, opened 2024-12-19
charts: Replace hardcoded namespaces

atanasdinov:

Removes the hardcoded namespace values used within the different chart hooks.

There are some resources created in the default namespace across the `pre-install-job` and `post-upgrade-job` hooks. Should those be updated as well?
- https://github.com/rancher/turtles/blob/main/charts/rancher-turtles/templates/post-upgrade-job.yaml#L54
- https://github.com/rancher/turtles/blob/main/charts/rancher-turtles/templates/pre-install-job.yaml#L7
- https://github.com/rancher/turtles/blob/main/charts/rancher-turtles/templates/pre-install-job.yaml#L61
- https://github.com/rancher/turtles/blob/main/charts/rancher-turtles/templates/pre-install-job.yaml#L83

**Checklist**:

- [X] squashed commits into logical changes
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/rancher-turtles/templates/post-delete-job.yaml
+++ b/charts/rancher-turtles/templates/post-delete-job.yaml
@@ -41,7 +41,7 @@ metadata:
 subjects:
   - kind: ServiceAccount
     name: post-delete-job
-    namespace: rancher-turtles-system
+    namespace: '{{ .Values.rancherTurtles.namespace }}'
 roleRef:
   kind: ClusterRole
   name: post-delete-job-delete-webhooks
--- a/charts/rancher-turtles/templates/post-upgrade-job.yaml
+++ b/charts/rancher-turtles/templates/post-upgrade-job.yaml
@@ -4,7 +4,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: post-upgrade-job
-  namespace: rancher-turtles-system
+  namespace: '{{ .Values.rancherTurtles.namespace }}'
   annotations:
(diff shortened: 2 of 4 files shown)


salasberryfin:
Thanks @atanasdinov, I agree this is a sensible thing to support. Maybe it would be great to also have some kind of test that validates that nothing breaks if a user specifies a custom namespace (just to be 100% sure), did you have the chance to test it yourself? There may be some "hidden" references to `rancher-turtles-system` as it's been considered "Turtles' namespace" until now, that could eventually beak something.

I think it makes sense to also create `pre-install`/`post-upgrade` jobs in the user-specified namespace. This one you're also linking https://github.com/rancher/turtles/blob/451c249449b9371dc9cc04263e14fff972c6bb16/charts/rancher-turtles/templates/pre-install-job.yaml#L7 disables the existing Rancher CAPI feature and should stay the same, though.

atanasdinov:
As far as manual testing goes, I didn't see any problems. I also tried to install the chart in a different namespace and there wasn't an issue (where both the chart namespace, as well as the one defined in the values file are the same).

The respective hook jobs are now also transferred over to the chart namespace and confirmed to be working properly. The only caveat for the pre-install ones is that the namespace has to be created first -- either manually or via `--create-namespace` during the Helm install process (I assume Rancher does it?).

Let me know what you think.

Danil-Grigorev (approved):

alexander-demicev (approved):

PR #958, opened 2024-12-31
Additional e2e fixes

anmazzotti:
This PR is to fix the e2e workflow and bring some improvements.

The turtles-e2e image is now only built once as pre-requisite job. I kept the `v0.0.1` tag because changing this has further implications down the line and should be propagated better, for example to the helm chart values and the `test/e2e/config/operator.yaml` file. 

The image is also now deleted correctly always, when all parallel jobs end.
See: https://github.com/rancher/turtles/pkgs/container/turtles-e2e

Sample run: https://github.com/rancher/turtles/actions/runs/12595032471

### DIFF
--- a/.github/workflows/run-e2e-suite.yaml
+++ b/.github/workflows/run-e2e-suite.yaml
@@ -31,7 +31,7 @@ on:
 
 permissions:
   contents: read
-  packages: write
+  packages: read
 
 env:
   NGROK_AUTHTOKEN: ${{ secrets.NGROK_AUTHTOKEN }}
@@ -54,7 +54,7 @@ env:
   GINKGO_LABEL_FILTER: full 
   GINKGO_TESTS: ${{ github.workspace }}/${{ inputs.test_suite }}
   GINKGO_NODES: 5
-  TAG: v${{ github.run_number }}.${{ github.run_attempt }}
+  TAG: v0.0.1
 
 jobs:
   run_e2e_tests:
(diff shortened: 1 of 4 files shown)


salasberryfin (approved):

furkatgofurov7:
I was checking https://github.com/rancher/turtles/pkgs/container/turtles-e2e and looks like image cleaning failed. Is it unrelated to this PR and can be ignored?

.github/workflows/run-e2e-suite.yaml
```diff
@@ -54,7 +54,7 @@ env:
   GINKGO_LABEL_FILTER: full 
   GINKGO_TESTS: ${{ github.workspace }}/${{ inputs.test_suite }}
   GINKGO_NODES: 5
-  TAG: v${{ github.run_number }}.${{ github.run_attempt }}
+  TAG: v0.0.1
```
Danil-Grigorev, line 57: Would this result in parallel e2e runs potentially using tag built in a different run?

anmazzotti: No the run_number and run_attempt is the same for all parallel runs. 
However this is now back to the static `v0.0.1`, as there are further fixes needed to make it dynamic

anmazzotti:
@furkatgofurov7 the cleanup fails because the package  has only one version left. GitHub is preventing deletion of that version (even though the error message is super misleading). 

Next time it will run the current `v0.0.1` version will become untagged, and all follow up cleanups will succeed. 
The drawback is that we will have this trailing untagged image as a leftover. 

I can experiment further and see if we can delete the entire package, instead of the specific version only, but possibly outside this PR.

Danil-Grigorev (approved):

PR #961, opened 2025-01-02
Re-enable CAPRKE2 template in e2e, updated to work with CAPI 1.9+

### LINKED ISSUE #965: Re-enable CAPRKE2 tests in test suite
Our CAPRKE2 tests are disabled for ~6 months at this point, which leaves us blind to the state of CI after each CAPI or CAPRKE2 provider bump in relation to the system in general.

We were experiencing issues related to CI runner insufficient memory, leading to nodes in the RKE2 cluster not marked as healthy. This lead to variety of issues, including inability to provision CAPRKE2 cluster to the healthy state, or install the `cattle-agent` on it to provide turtles import functionality. As the nodes are in a non-healthy state, scheduler does not have a healthy node to provision our components, leading to failures which are mostly only reproducible in CI environment.

After some investigation, it seems latest version of RKE2 provision a set of unnecessary components, contributing to the memory issue. Such as:
- `nginx` controller
- `canal` CNI in the case of a non-specified value. For CAPD it is sufficient to run a very lightweight `kindnet` CNI, deployed via other means.

Disabling this components and behavior leads to success in our e2e scenarios and issue disappearance.

Danil-Grigorev:

After ~6 months the test being disabled due to memory issues observed in CI, latest set of changes seems to resolve the issue locally, allowing the agent to be scheduled on the nodes. Attempting to re-enable the test to run in pre-submits, again.

Fixes #965

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


exp/README.md
```diff
@@ -42,12 +42,9 @@ export WORKER_MACHINE_COUNT=1
...
 export KUBERNETES_VERSION=v1.30.0
 export RKE2_VERSION=v1.31.0+rke2r1
+export RKE2_CNI=calico
 
-# Prevent replacing lb-config variables by envsubst
-export address='$address'
-export server='$server'
-
-envsubst < test/e2e/data/cluster-templates/docker-rke2.yaml | kubectl apply -f -
+envsubst '${CLUSTER_NAME} ${WORKER_MACHINE_COUNT} ${RKE2_VERSION} ${CONTROL_PLANE_MACHINE_COUNT} ${KUBERNETES_VERSION} ${RKE2_CNI}' < test/e2e/data/cluster-templates/docker-rke2.yaml | kubectl apply -f -
```
salasberryfin, line 47:
```suggestion
envsubst < test/e2e/data/cluster-templates/docker-rke2.yaml | kubectl apply -f -
```
Do these variables need to be explicitly passed here even after export?

salasberryfin (approved):
Thanks @Danil-Grigorev. It's great that we can finally have this back! I added a minor comment that can be ignored if you don't feel like it's relevant.

exp/README.md
Danil-Grigorev: This is actually what helped with the issue I mentioned today. Because any `$` notation treated as variable, a new `$backend` variable was substituted with empty string, being hard to debug. This way only known variables are substituted, thanks to @thehejik suggestion. It is better then remembering
```bash
export server=‘$server’
```
for every new change.

salasberryfin: Thanks!

test/e2e/data/cluster-templates/docker-rke2.yaml
```diff
@@ -48,13 +50,18 @@ metadata:
...
-  registrationMethod: internal-first
   rolloutStrategy:
     rollingUpdate:
       maxSurge: 1
     type: RollingUpdate
   serverConfig:
+    cni: ${RKE2_CNI}
+    kubeAPIServer:
+      extraArgs:
+      - --anonymous-auth=true
```
furkatgofurov7, line 61: Do I understand this correctly, that we authorize anonymous connection here?

furkatgofurov7 (commented):
Great, thanks for working on this!

test/e2e/data/cluster-templates/docker-rke2.yaml
Danil-Grigorev: I documented the reasoning for this during upgrading to 1.9 on CAPRKE2 provider. This is a large issue, as load balancer were still treating API sever as healthy when ETCD was down due to removal of the member and local endpoint usage: https://github.com/rancher/cluster-api-provider-rke2/issues/530

alexander-demicev (approved):

test/e2e/data/cluster-templates/docker-rke2.yaml
furkatgofurov7: Thanks for explanation, it was non-blocking question.

Issue #847, opened 2024-11-08
Handle CA certificates for turtles manager to avoid `Pending` state of CAPI cluster after importing to Rancher


thehejik:
### What would you like to be added (User Story)?

It looks more and more people is trying to install turtles into rancher but they are facing an initial issue with `Pending` clusters after provisioning and importing into rancher.

We should give them a guidance via doc and/or add a value/question into turtles helm chart/App to either mount/import rancher's known CA certificate or append `--insecure-skip-verify=true` as additional manager argument in its Deployment.

The capi manager deployment where the extra argument should be added is called `rancher-turtles-controller-manager` in `rancher-turtles-system` namespace.

### Label(s) to be applied

/kind feature

kkaempf:
release-notes maybe ?

PR #977, opened 2025-01-06
Add ability to leverage agent tls mode setting

alexander-demicev:

After merging this PR, users can leverage CA certs stored in Rancher setting to import CAPI clusters.

Fixes https://github.com/rancher/turtles/issues/847

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/internal/controllers/helpers.go
+++ b/internal/controllers/helpers.go
@@ -20,6 +20,7 @@ import (
 	"bufio"
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"errors"
 	"fmt"
 	"io"
@@ -58,7 +59,7 @@ const (
 )
 
 func getClusterRegistrationManifest(ctx context.Context, clusterName, namespace string, cl client.Client,
-	insecureSkipVerify bool,
+	caCert []byte, insecureSkipVerify bool,
 ) (string, error) {
 	log := log.FromContext(ctx)
 
@@ -85,7 +86,7 @@ func getClusterRegistrationManifest(ctx context.Context, clusterName, namespace 
 		return "", nil
 	}
 
-	manifestData, err := downloadManifest(token.Status.ManifestURL, insecureSkipVerify)
+	manifestData, err := downloadManifest(token.Status.ManifestURL, caCert, insecureSkipVerify)
 	if err != nil {
 		log.Error(err, "failed downloading import manifest")
 		return "", err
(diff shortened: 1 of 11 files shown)


internal/controllers/helpers.go
```diff
@@ -285,3 +298,49 @@ func createObject(ctx context.Context, c client.Client, obj client.Object) error
...
+		return nil, nil
+	}
+
+	agentTLSModeSetting := &managementv3.Setting{}
+
+	if err := cl.Get(ctx, client.ObjectKey{
+		Name: "agent-tls-mode",
+	}, agentTLSModeSetting); err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("agent-tls-mode setting not found, using system store")
```
anmazzotti, line 316: Nitpicking a bit. This does not align with Rancher, where the default is `strict` (since v2.9.0): https://ranchermanager.docs.rancher.com/getting-started/installation-and-upgrade/installation-references/tls-settings#agent-tls-enforcement. 

I would default to strict instead for this reason. I think it's also fair to just remove this branch and let it error out when this setting is not found, as this is not an expected scenario. The `agent-tls-mode` was added in Rancher v2.8

anmazzotti (approved):

furkatgofurov7 (dismissed):

internal/controllers/helpers.go
```diff
@@ -285,3 +298,49 @@ func createObject(ctx context.Context, c client.Client, obj client.Object) error
...
+			return nil, fmt.Errorf("error getting ca-certs setting: %w", err)
+		}
+
+		if caCertsSetting.Value == "" {
+			return nil, errors.New("ca-certs setting value is empty")
+		}
+
+		return []byte(caCertsSetting.Value), nil
+	default:
+		return nil, fmt.Errorf("invalid agent-tls-mode setting value: %s", agentTLSModeSetting.Value)
```
Danil-Grigorev, line 344: It can actually be empty as well and treated as `system-store` setting: https://github.com/rancher/rancher/blob/649e7521bec6e4f2c83a8d3b9e4f1365b29e46e8/chart/values.schema.json#L6
```suggestion
		log.Info("using system store for CA certificates")
 		return nil, nil
```

alexander-demicev: The setting should not be empty, if it's empty then we don't how to behave

Danil-Grigorev: Does it mean the comment on the provided line is wrong? `agentTLSMode must be 'strict' or 'system-store' or null (defaults to system-store)`. It looks to me that `null` or empty is treated as `system-store`.

anmazzotti (approved):

internal/controllers/helpers.go
alexander-demicev: Rancher UI doesn't allow an empty option, I assume it shouldn't be empty.

anmazzotti: The default should always be systemStore, not empty. It can be that users may nullify it manually through the Settings resource, but I would not support this scenario. 

For [reference](https://github.com/rancher/rancher/blob/de3ffa88cc75ae3da122bd36a4489663b5157ee3/pkg/settings/setting.go#L62):

```go
AgentTLSMode                        = NewSetting("agent-tls-mode", AgentTLSModeStrict).WithDefaultOnUpgrade(AgentTLSModeSystemStore)
```

anmazzotti (approved):

internal/controllers/helpers.go
Danil-Grigorev: Ok, makes sense. If users left it empty, they’ll need to upgrade or fix it, with an error returned.

Danil-Grigorev (approved):

PR #986, opened 2025-01-07
Permit using github release or preloaded binary for system-agent

### LINKED ISSUE #975: [ETCD-snapshot-restore] Run e2e tests via `short` suite
Currently ETCD Snapshot restore suite runs under `local` label, making it easy to miss a change in test results. We need to make the test run under `short` suite either as a pre-submit or nightly, so we can catch issues early.

What blocks this, is that in a `local` run we use `isolated-kind` setup, which uses `nginx` ingress and `sslip.io`. This results in a missing certificate, which causes inability to download `system-agent` binary from `rancher` endpoint.

For the e2e purposes, we may explore ignoring certificate validation while downloading `system-agent` binary to mitigate the issue.

Danil-Grigorev:

We need to allow alternative installation methods for the rancher `system-agent` binary. Users may want either:
- Pre-load system agent in the image manually via cloud-config or other means in the air-gapped installation scenario
- Use github release to download agent binary

This change allows to perform both via added annotations on the CAPI `Cluster` object.

Examples:
```yaml
apiVersion: cluster.x-k8s.io/v1beta1
kind: Cluster 
metadata:
  name: rke2-cluster
  annotations:
    cluster-api.cattle.io/upstream-system-agent: "true" # Install agent binary from github repo release
```

```yaml
apiVersion: cluster.x-k8s.io/v1beta1
kind: Cluster 
metadata:
  name: rke2-cluster
  annotations:
    cluster-api.cattle.io/local-system-agent: "/usr/local/bin/rancher-system-agent" # Use existing agent binary located at /usr/local/bin/rancher-system-agent
```

Depends on #987 

Fixes #975 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/exp/etcdrestore/webhooks/install.sh
+++ b/exp/etcdrestore/webhooks/install.sh
@@ -30,6 +30,7 @@ fi
 #   - CATTLE_AGENT_UNINSTALL_URL (default: latest GitHub release)
 #   - CATTLE_PRESERVE_WORKDIR (default: false)
 #   - CATTLE_REMOTE_ENABLED (default: true)
+#   - CATTLE_UPSTREAM_ENABLED (default: false)
 #   - CATTLE_LOCAL_ENABLED (default: false)
 #   - CATTLE_AGENT_BINARY_LOCAL (default: false)
 #   - CATTLE_AGENT_BINARY_LOCAL_LOCATION (default: )
@@ -215,7 +216,7 @@ setup_env() {
         CATTLE_REMOTE_ENABLED=$(echo "${CATTLE_REMOTE_ENABLED}" | tr '[:upper:]' '[:lower:]')
     fi
 
-    if [ "${CATTLE_LOCAL_ENABLED}" = "false" ] && [ "${CATTLE_REMOTE_ENABLED}" = "false" ]; then
+    if [ "${CATTLE_LOCAL_ENABLED}" = "false" ] && [ "${CATTLE_REMOTE_ENABLED}" = "false" ] && [ "${CATTLE_UPSTREAM_ENABLED}" = "false" ]; then
         fatal "Neither local or remote plan support was enabled"
     fi
 
--- a/exp/etcdrestore/webhooks/rke2config.go
+++ b/exp/etcdrestore/webhooks/rke2config.go
@@ -25,6 +25,7 @@ import (
 	bootstrapv1 "github.com/rancher/cluster-api-provider-rke2/bootstrap/api/v1beta1"
 	managementv3 "github.com/rancher/turtles/api/rancher/management/v3"
 
+	turtlesannotations "github.com/rancher/turtles/util/annotations"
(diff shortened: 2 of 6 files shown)


exp/etcdrestore/webhooks/install.sh
```diff
@@ -215,7 +216,7 @@ setup_env() {
         CATTLE_REMOTE_ENABLED=$(echo "${CATTLE_REMOTE_ENABLED}" | tr '[:upper:]' '[:lower:]')
     fi
 
-    if [ "${CATTLE_LOCAL_ENABLED}" = "false" ] && [ "${CATTLE_REMOTE_ENABLED}" = "false" ]; then
+    if [ "${CATTLE_LOCAL_ENABLED}" = "false" ] && [ "${CATTLE_REMOTE_ENABLED}" = "false" ] && [ "${CATTLE_UPSTREAM_ENABLED}" = "false" ]; then
```
anmazzotti, line 219: I'm a bit confused by this. 
Wouldn't be better to remove this check entirely, instead of adding `CATTLE_UPSTREAM_ENABLED` to bypass it?

The condition is still the same, there is no logic attached to the new environment variable.

anmazzotti: Also `CATTLE_LOCAL_ENABLED` and `CATTLE_REMOTE_ENABLED` are about the system-agent Plans, not how the system-agent itself is installed, so I miss the correlation with the new env.

anmazzotti: And by that I mean that if `CATTLE_LOCAL_ENABLED` and `CATTLE_REMOTE_ENABLED` are both false, then the system-agent won't be able to consume any plan, it should still be fatal.

Danil-Grigorev: > Also `CATTLE_LOCAL_ENABLED` and `CATTLE_REMOTE_ENABLED` are about the system-agent Plans, not how the system-agent itself is installed, so I miss the correlation with the new env.

This is incorrect assumption. It is only used for system agent installation.

> I'm a bit confused by this. Wouldn't be better to remove this check entirely, instead of adding `CATTLE_UPSTREAM_ENABLED` to bypass it?
> 
> The condition is still the same, there is no logic attached to the new environment variable.

I’m not an author of this script, and such change requires lot’s of bash modifications down the line, which likely affect the quality of it and lead to issues hard to debug.

> And by that I mean that if `CATTLE_LOCAL_ENABLED` and `CATTLE_REMOTE_ENABLED` are both false, then the system-agent won't be able to consume any plan, it should still be fatal.

Which is the current script logic is doing. It is best to leave it as is.

alexander-demicev (approved):

alexander-demicev:
We can think about moving the entire system-agent installation logic to CAPRKE2 to avoid relying on annotations, which is ok while the feature is experimental

anmazzotti (approved):

PR #988, opened 2025-01-08
Delete unused envs

anmazzotti:

Just a little bump to ensure we are by default testing with latest version.


furkatgofurov7 (approved):

Danil-Grigorev:
I suggest removing this variable as it has no effect on dev setup.

anmazzotti:
Good point @Danil-Grigorev 
I also removed the CAPI version that didn't seem to be in use either.

Danil-Grigorev (approved):

furkatgofurov7 (approved):

PR #897, opened 2024-12-02
Unify etcd snapshot-restore image with turtles

Danil-Grigorev:
**What this PR does / why we need it**:

We need to build etcd snapshot/restore image together with turtles, to utilize same build and release path.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #895 

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits into logical changes
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

PR #1007, opened 2025-01-16
ci: Remove incorrect make target

yiannistri:

Makefile target `docker-build-and-push-etcdrestore` was removed in https://github.com/rancher/turtles/pull/897 . Also `HELM_EXPERIMENTAL_OCI` should not be required for Helm >= 3.8.0 and the Go compiler should not be needed since we're building inside a container.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


alexander-demicev (approved):

furkatgofurov7:
LGTM, thanks!

PR #1011, opened 2025-01-17
Support deduplication when using ClusterctlConfig

furkatgofurov7:

If new provider is added through clusterctlconfig:

1. which does not exist in the current config map it will be added as new entry to configmap
2. with new URL and  if provider exists, it only updates the URL of that entry
3. without any changes to URL and if provider exists, it stays as is without any changes.

I am planning to add tests as a follow up PR, since config.go anyways did not have a dedicated test file up until now.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/internal/controllers/clusterctl/config.go
+++ b/internal/controllers/clusterctl/config.go
@@ -102,7 +102,27 @@ func ClusterConfig(ctx context.Context, c client.Client) (*ConfigRepository, err
 		clusterctlConfig.Images = map[string]ConfigImage{}
 	}
 
-	clusterctlConfig.Providers = append(clusterctlConfig.Providers, config.Spec.Providers...)
+	// Deduplicate and update providers
+	existingProviders := make(map[string]int)
+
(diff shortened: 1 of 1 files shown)


Danil-Grigorev (approved):
Can you please add unit tests for the PR? Also you mentioned some kind of issue with bootstrap or control plane provider? Is this resolved? I see no changes since last commit @furkatgofurov7 

LGTM otherwise

alexander-demicev (approved):
I removed "fixes" from the PR description so issue doesn't close. Tests can be added in a follow-up PR.

furkatgofurov7:
> Can you please add unit tests for the PR? Also you mentioned some kind of issue with bootstrap or control plane provider? Is this resolved? I see no changes since last commit @furkatgofurov7
> 
> LGTM otherwise

No issues with bootstrap after I tested it properly and it is resolved. I will open a follow-up with unit tests addition.

PR #1018, opened 2025-01-20
feat: add dockerfile cross-compilation for multi-platform images

### LINKED ISSUE #933: Investigate cross compilation
# Description

Based on https://github.com/rancher/cis-operator/blob/787cec3fc95d8f856bff5a446ebf06e122c6249d/package/Dockerfile, let's investigate if we can use cross compilation for multiarch images so we improve release times.

salasberryfin:

Use cross-compilation when building the image with `buildx`. This means that the process does not need to use emulation for the whole build which makes it considerably faster.

We can use the existing build arguments (which are globally available):
- `BUILDPLATFORM`: make the image always native to the machine.
- `TARGETOS` and `TARGETARCH`: configure the compiler process for the specific architecture.

Fixes #933 

I ran a quick test and image building is a few minutes faster after this change.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@
 # Build the manager binary
 ARG builder_image
 
-FROM ${builder_image} as builder
+FROM --platform=$BUILDPLATFORM ${builder_image} as builder
 WORKDIR /workspace
 
 # Run this with docker build --build-arg goproxy=$(go env GOPROXY) to override the goproxy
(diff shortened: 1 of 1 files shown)


yiannistri (commented):
For reference, here is the [list of multi-platform build arguments](https://docs.docker.com/build/building/variables/#multi-platform-build-arguments). The release GHA is calling `make docker-build-and-push` and uses the $(TARGET_PLATFORMS) variable which is set to `linux/amd64,linux/arm64`.

I need to test how Tilt behaves with this change.

Dockerfile
```diff
@@ -17,7 +17,8 @@
 # Build the manager binary
 ARG builder_image
 
-FROM ${builder_image} as builder
+#FROM ${builder_image} as builder
```
furkatgofurov7, line 20: do we need to keep the commented code?

Dockerfile
```diff
@@ -17,7 +17,8 @@
...
+FROM --platform=$BUILDPLATFORM ${builder_image} as builder
```
furkatgofurov7, line 21: where is `BUILDPLATFORM` arg defined?

furkatgofurov7 (commented):
Have not run it locally yet, some comments/questions inline:

Dockerfile
salasberryfin: Yep, this was committed by mistake. Thanks for noticing!

Danil-Grigorev (approved):

alexander-demicev:
@yiannistri tilt worked for me fine

PR #1032, opened 2025-01-22
chore: remove rancher-webhook feature from chart

### LINKED ISSUE #1025: Remove `features.embedded-capi` and `features.rancher-webhook`
# Description

features.embedded-capi and features.rancher-webhook can be removed and their functionality should be triggered by rancherInstalled value to support cases when Turtles is installed before Rancher. We can add a new e2e that verifies the flow.

salasberryfin:

Remove `features.rancher-webhook` and use `rancherInstalled` instead.

Fixes #1025

**This cherry-picks 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


charts/rancher-turtles/values.yaml
```diff
@@ -25,6 +20,7 @@ rancherTurtles:
       enabled: true
     agent-tls-mode:
       enabled: false
+    kubectlImage: registry.k8s.io/kubernetes/kubectl:v1.30.0
```
alexander-demicev, line 23: maybe this can be moved to the top level along with the CAPI operator one? so users set it only in one place

charts/rancher-turtles/questions.yml
```diff
@@ -56,9 +49,3 @@ questions:
     type: boolean
     label: Enable Agent TLS Mode
     group: "Rancher Turtles Features Settings"
-  - variable: rancherTurtles.features.kubectlImage
```
anmazzotti: I don't think this should be deleted, the variable is still there, just moved location, and it would still be nice to configure this when installing the chart from Rancher

anmazzotti (changes requested):

charts/rancher-turtles/questions.yml
salasberryfin: I added it back under the features section. Even though it is not a `features` value in `values.yaml`, looking at the way it's displayed in Rancher UI I think it's where it fits better for users.

anmazzotti (approved):

Danil-Grigorev (approved):

PR #1036, opened 2025-01-23
Remove docker build for clusterclass manager

anmazzotti:

@salasberryfin I noticed the `exp/clusterclass/Dockerfile` was cleaned up, so I think it makes sense to not try to build and use the docker image at this stage. `make dev-env` shows an error when trying to build the image and may be confusing for users.

That said, the clusterclass manager is correctly loaded by tilt, so it's still available in the development environment.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


salasberryfin (approved):
Thanks @anmazzotti. I'm a bit confused why I ended up leaving it like this. I agree it's better to remove it and I think it also aligns with what we do with `etcdrestore`.

yiannistri (approved):

PR #1045, opened 2025-01-24
feat: Bump CAPI and forks to latest downstream release versions

furkatgofurov7:
Prepare turtles for upcoming release by bumping dependencies to the latest downstream released versions.  

Fixes #

1. CAPRKE2 and CAPI Operator will also be bumped once new release versions are available, hence, they are not part of this change.
2. New releases of downstream providers are published in advance of this change.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


salasberryfin (approved):
Thanks @furkatgofurov7.

Some of these bumps are already part of `updatecli` . It may be interesting to think about improving the configuration so we can also have the rest automatically updated in separate PRs so we can choose what to merge. For reference https://github.com/rancher/turtles/pull/1046/

Danil-Grigorev (approved):

furkatgofurov7:
> Thanks @furkatgofurov7.
> 
> Some of these bumps are already part of `updatecli` . It may be interesting to think about improving the configuration so we can also have the rest automatically updated in separate PRs so we can choose what to merge. For reference #1046

@salasberryfin thanks for comment and I noticed also #1046 but it was missing updates to:

- new test file using CAPI version (we can easily add it to existing updatecli configuration)
- go.mod/sum files in `root`, `test` and `exp/etcdrestore` folders which should be in theory bumped by dependabot (although it does not bump all at once but in chunks/separate PRs) but at times we might need that bump at once and immediately. Do you know if we make add go.mod/sum changes (it should be able to run `go mod tidy`) as part of updatecli?

Issue #1061, opened 2025-01-29
Cluster not auto-importing after performing turtles chart update


cpinjani:
### What steps did you take and what happened?

- Install turtles v0.16.0-rc.0, default settings via Rancher Dashboard
- Import CAPI cluster (turtles deployment is patched with `--insecure-skip-verify=true` to successfully import it)
- Perform chart update and enable Agent TLS mode feature
- Cluster is removed and not seen Rancher, however CAPI cluster exists.

OR

- Install turtles v0.16.0-rc.0, customize the install settings: Agent TLS enabled, Fleet Addon disabled
- Import CAPI cluster ans wait for it to be Active
- Perform chart update and enable Fleet Addon
- Cluster is removed and not seen Rancher, however CAPI cluster exists.

Import related logs:
```
I0129 11:20:25.551454       1 import_controller_v3.go:146] "Reconciling CAPI cluster" controller="cluster" controllerGroup="cluster.x-k8s.io" controllerKind="Cluster" Cluster="default/cluster1" namespace="default" name="cluster1" reconcileID="db97a678-db3d-48c0-bf44-0243c88bdb7b"
I0129 11:20:25.551580       1 import_controller_v3.go:464] "Reconciling rancher cluster deletion" controller="cluster" controllerGroup="cluster.x-k8s.io" controllerKind="Cluster" Cluster="default/cluster1" namespace="default" name="cluster1" reconcileID="db97a678-db3d-48c0-bf44-0243c88bdb7b"
I0129 11:20:25.551591       1 import_controller_v3.go:467] "Rancher cluster is being removed, annotating CAPI cluster cluster1 with imported" controller="cluster" controllerGroup="cluster.x-k8s.io" controllerKind="Cluster" Cluster="default/cluster1" namespace="default" name="cluster1" reconcileID="db97a678-db3d-48c0-bf44-0243c88bdb7b"
I0129 11:20:25.592681       1 import_controller_v3.go:146] "Reconciling CAPI cluster" controller="cluster" controllerGroup="cluster.x-k8s.io" controllerKind="Cluster" Cluster="default/cluster1" namespace="default" name="cluster1" reconcileID="fba7460a-f271-4613-a2e5-d101c2b93606"
[...]

PR #979, opened 2025-01-06
Enable v3 migration by default

alexander-demicev:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes https://github.com/rancher/turtles/issues/978
Fixes https://github.com/rancher/turtles/issues/930

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits into logical changes
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

PR #581, opened 2024-06-26
🌱 Support for management v3 downgrade to provisioning v1 with automigration

Danil-Grigorev:
**What this PR does / why we need it**:

Add downgrade procedure as described in the ADR #579 

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits into logical changes
- [ ] includes documentation
- [x] adds unit tests
- [x] adds or updates e2e tests

PR #1063, opened 2025-01-29
fix: remove v1 clusters for migration in post-upgrade job

salasberryfin:

Now that there's a single method for importing clusters via `clusters.management.cattle.io` (v3) and migration is automatically applied for all `clusters.provisioning.cattle.io` (v1) we no longer have values in the Helm chart that control when Rancher clusters deletion is applied ([this](https://github.com/rancher/turtles/commit/1f6972edaea0c05de2406a9bd3020f8171a67576#diff-c0588fa3a386b0b96a535569bdd77004aa57407c73e52c49502cfd52a46eb23d) is where the changes are introduced).

In this case, as reported by @cpinjani [here](https://github.com/rancher/turtles/issues/1061), CAPI clusters are being un-imported after upgrade because the post-upgrade job deletes all v3 cluster resources and then the controller labels the CAPI cluster `imported=true`. For this to apply only for migration scenarios we should delete v1 clusters only (and eventually remove this action completely?).

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/rancher-turtles/templates/post-upgrade-job.yaml
+++ b/charts/rancher-turtles/templates/post-upgrade-job.yaml
@@ -23,13 +23,6 @@ rules:
   verbs:
   - list
   - delete
-- apiGroups:
-  - management.cattle.io
-  resources:
-  - clusters
-  verbs:
-  - list
-  - delete
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
@@ -65,7 +58,7 @@ spec:
           image: {{ index .Values "rancherTurtles" "kubectlImage" }}
           args:
           - delete
-          - clusters.management.cattle.io
+          - clusters.provisioning.cattle.io
           - --selector=cluster-api.cattle.io/owned
           - -A
           - --ignore-not-found=true


charts/rancher-turtles/templates/post-upgrade-job.yaml
```diff
@@ -23,13 +23,6 @@ rules:
   verbs:
   - list
   - delete
-- apiGroups:
```
furkatgofurov7, line 26: I was looking into git history to understand this better and it seems the particular piece of code was introduced in https://github.com/rancher/turtles/pull/581/commits/75d682ea81ce012ba8c00e2512c847d741ea65eb to give a possibility to opt-in for cluster downgrade scenario with managementv3-cluster-migration. 

And since we enabled v3 migration by default recently, look like it was **partially** removed in https://github.com/rancher/turtles/pull/979/commits/1f6972edaea0c05de2406a9bd3020f8171a67576#diff-c0588fa3a386b0b96a535569bdd77004aa57407c73e52c49502cfd52a46eb23dL1 and remaining parts are now cleaned-up with this patch. Does it sound an accurate summary @salasberryfin ?

furkatgofurov7 (approved):

Danil-Grigorev (approved):
Thanks, LGTM

PR #1065, opened 2025-01-30
🌱 Remove migration code and e2e tests

Danil-Grigorev:

Release v0.16 enabled migration by default, so in preparation for v0.17 the migration code can be removed, to finalize transition to managementv3 clusters.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


alexander-demicev (approved):

anmazzotti (approved):

salasberryfin (commented):
I'm okay with merging this but I'm wondering what would happen if a user upgrades from <v0.16 to >=v0.17. If they had v1 clusters that hadn't explicitly been migrated (as migration was not the default behavior before v0.16), wouldn't they end up with a broken cluster import?

I know this has to happen at some point, and we won't have control over every user behavior corner case, so I'm not again making this change now but, what if we delayed this to v0.18 and give users a two-release advise?

As I said, not a blocker, just trying to understand the implications of this change :)

furkatgofurov7:
> As I said, not a blocker, just trying to understand the implications of this change :)

+1

Danil-Grigorev:
Thanks for the review @salasberryfin, it should not be a problem, as long as our chart is upgraded sequentially and does not skipping versions. The migration flag was available for the last couple of versions, as well as `management.v3` clusters were enabled by default. So people had enough time to notice duplicate cluster imports and remove them via migration or manually.

We still need to update our docs to review/remove current warnings displayed there, so this implication can be documented.

salasberryfin (approved):

PR #1127, opened 2025-03-06
test: Ensure CAPV tests have the correct providers set up

yiannistri:

Updated CAPV tests to ensure the correct providers are set up before test execution:
- RKE2 test was not setting up CAPV provider.
- Kubeadm test was not setting up Kubeadm bootstrap and control plane providers.

The vSphere secret was also renamed to match provider name otherwise it failed to install. 

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -77,7 +77,7 @@ This will consequently:
 From the project root directory:
 
 ```bash
-GINKGO_LABEL_FILTER=local make test-e2e
+TAG=v0.0.1 GINKGO_LABEL_FILTER=local make test-e2e
 ```
 
 **Important note:** The vSphere e2e tests require a VPN connection, which makes their integration into the daily e2e CI job challenging. Therefore,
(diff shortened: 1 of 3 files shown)


test/e2e/const.go
```diff
@@ -42,6 +42,12 @@ var (
 	//go:embed data/capi-operator/azure-provider.yaml
 	AzureProvider []byte
 
+	//go:embed data/capi-operator/kubeadm-bootstrap-provider.yaml
+	KubeadmBootstrapProvider []byte
```
alexander-demicev, line 46: can these providers be installed in the same way as we do for Docker https://github.com/rancher/turtles/blob/main/test/e2e/suites/import-gitops-v3/import_gitops_v3_test.go#L43?

alexander-demicev (changes requested):

alexander-demicev:
looks good, just one small comment

test/e2e/const.go
yiannistri: Sure, I wasn't sure whether we wanted to keep things isolated or whether it's ok installing CAPD as well. I will align the tests.

alexander-demicev (approved):

PR #1160, opened 2025-03-19
Add AKS ClusterClass example

anmazzotti:

This PR converts the AKS ClusterClass used for testing, into a slightly configurable example.
Documentation PR: https://github.com/rancher/turtles-docs/pull/262
Test run: https://github.com/rancher/turtles/actions/runs/13952503039 (Some other tests fail for other reasons, but the AKS one succeeds)

Fixes #

**Checklist**:

- [x] includes documentation
- [ ] adds unit tests

### DIFF
--- a/test/e2e/data/capi-operator/azure-provider.yaml
+++ b/test/e2e/data/capi-operator/azure-provider.yaml
@@ -6,8 +6,3 @@ metadata:
   namespace: capz-system
 spec:
   type: infrastructure
-  deployment:
-    containers:
-      - name: manager
-        args:
-          "--v": "5" # CAPZ only displays meaningful logs at level 5
--- a/test/e2e/suites/import-gitops-v3/import_gitops_v3_test.go
+++ b/test/e2e/suites/import-gitops-v3/import_gitops_v3_test.go
@@ -132,13 +132,20 @@ var _ = Describe("[Azure] [AKS] Create and delete CAPI cluster from cluster clas
 			},
 		})
 
+		// Add the needed ClusterClass
+		fixedNamespace := "creategitops-azure-aks"
+		Expect(turtlesframework.CreateNamespace(ctx, bootstrapClusterProxy, fixedNamespace)).Should(Succeed())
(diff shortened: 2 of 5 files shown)


test/e2e/suites/import-gitops-v3/import_gitops_v3_test.go
```diff
@@ -132,13 +132,20 @@ var _ = Describe("[Azure] [AKS] Create and delete CAPI cluster from cluster clas
 			},
 		})
 
+		// Add the needed ClusterClass
+		fixedNamespace := "creategitops-azure-aks"
+		Expect(turtlesframework.CreateNamespace(ctx, bootstrapClusterProxy, fixedNamespace)).Should(Succeed())
```
alexander-demicev, line 137: why doesn't `go embed` work here?

anmazzotti: go embed can't reach the root folder and won't follow symlinks either

test/e2e/data/capi-operator/azure-provider.yaml
```diff
@@ -6,8 +6,3 @@ metadata:
   namespace: capz-system
 spec:
   type: infrastructure
-  deployment:
-    containers:
-      - name: manager
-        args:
```
anmazzotti, line 12: This is overriding the args, removing all the defaults, for example MachinePool feature support.
It would be best to fix this upstream, allowing for debug level configuration using an environment variable.

For now I think it's best to just remove it, I find the normal log level enough actually.

alexander-demicev: for CAPZ specifically, the default log level is not enough. Sometimes it hides errors related to for example credentials.

anmazzotti: Then again I'd rather fix this upstream, I don't think overriding the args is supported and should be used anyway, as we will diverge from upstream.

salasberryfin (approved):
Thanks @anmazzotti.

PR #187, opened 2023-10-04
✨Add vSphere e2e tests

furkatgofurov7:
**What this PR does / why we need it**:
Adds e2e tests for vSphere (CAPV) infrastructure provider

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #119 

**Special notes for your reviewer**:

**Checklist**:

- [x] squashed commits into logical changes
- [x] includes documentation
- [ ] adds unit tests
- [x] adds or updates e2e tests

PR #1163, opened 2025-03-19
:seedling: Add new management cluster environment option

### LINKED ISSUE #1162: Add support for creating bootstrap clusters in an internal network
We currently have 3 ways of creating a bootstrap cluster either for tests or local development:
- For clusters running in cloud providers, we use EKS for the management cluster (`eks`)
- For isolated clusters that share the same docker network, we use kind (`isolated-kind`) 
- For development clusters, we use kind + ngrok to expose the management cluster externally (`kind`)

We also need to be able to create bootstrap clusters that run in an internal network that expose the management cluster over an IP that is accessible from CAPI providers that run inside the network (for example CAPV). To do that, we can create a kind cluster with port mappings which forwards requests from the host.

yiannistri:

This PR adds a new option (`internal-kind`) for creating a local bootstrap cluster (using `kind` with extra port mappings)  that accepts requests from the host. This can be useful in environments where the CAPI provider is running in the local network  (eg. vSphere/CAPV). The new setup basically replicates this [guide](https://kind.sigs.k8s.io/docs/user/ingress/#option-2-extraportmapping).

Fixes #1162 

I have added a new bootstrap cluster provider instead of updating existing code we have in place for creating a `kind` cluster since the port mappings settings were [ignored](https://github.com/kubernetes-sigs/cluster-api/blob/d69751fe5d9c7f153341238bc66d6d5e723ad011/test/framework/bootstrap/kind_provider.go#L150-L154) when the docker socket extra mount was [specified](https://github.com/kubernetes-sigs/cluster-api/blob/d69751fe5d9c7f153341238bc66d6d5e723ad011/test/framework/bootstrap/kind_provider.go#L166) (useful for CAPD tests).

To test the setup of the bootstrap cluster run the following:
```
MANAGEMENT_CLUSTER_ENVIRONMENT="internal-kind" TAG=v0.0.1 GINKGO_LABEL_FILTER=vsphere make test-e2e
```

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


anmazzotti:
@yiannistri I think the extraMapping are ignored just because they are overwritten by the docker socket [code](https://github.com/kubernetes-sigs/cluster-api/blob/d69751fe5d9c7f153341238bc66d6d5e723ad011/test/framework/bootstrap/kind_provider.go#L202).

Would it be possible to fix this (it's a bug in our test suite anyway) and apply your changes to the current `kind` environment?

One other question I have is what provides `type: LoadBalancer` service? Do we have anything installed in the kind cluster for LB?

alexander-demicev (approved):

yiannistri:
@anmazzotti yes for sure, we can fix this upstream. I have added a new provider in the meantime only to get things working for both CAPD tests as well as the CAPV tests that will need the extra port mappings.

I will create an issue upstream and prepare a PR to address this.

yiannistri:
One downside of this approach (using extra port mappings) is that if the previous kind cluster is still running on the machine, the creation of a new kind cluster will fail because the 443 port will still be allocated to the previous cluster. We can ensure that no kind clusters are running when we begin a new test run, to mitigate this.

test/e2e/const.go
```diff
@@ -202,10 +206,11 @@ const (
...
 	AuthSecretName = "basic-auth-secret"
 
-	ShortTestLabel = "short"
-	FullTestLabel  = "full"
-	DontRunLabel   = "dontrun"
-	LocalTestLabel = "local"
+	ShortTestLabel   = "short"
+	FullTestLabel    = "full"
+	DontRunLabel     = "dontrun"
+	LocalTestLabel   = "local"
```
furkatgofurov7, line 212: I recall this was introduced by me in https://github.com/rancher/turtles/pull/187/files#diff-f41624bd6051a35a1694431afebcf5012b89a1cb9ac856464228f797065d34b1R125 while adding CAPV e2e tests but looking at the current code, that label and some more [logic](https://github.com/rancher/turtles/pull/187/files#diff-804cacacd09be55e4060407719307c3a82878d91b7d457fa71ad73ac767eb9a8R237-R239) looks redundant to me. Do we want to cleanup those as part of the same PR?

test/testenv/gitea.go
```diff
@@ -318,6 +334,7 @@ func PreGiteaInstallHook(giteaInput *DeployGiteaInput) {
 	case e2e.ManagementClusterEnvironmentIsolatedKind:
 		giteaInput.ServiceType = corev1.ServiceTypeNodePort
 	case e2e.ManagementClusterEnvironmentKind:
+	case e2e.ManagementClusterEnvironmentInternalKind:
```
furkatgofurov7, line 337: I will leave it to you to accept or leave it as is, but I thought below is bit more readable :)

```suggestion
	case e2e.ManagementClusterEnvironmentKind, e2e.ManagementClusterEnvironmentInternalKind:
```

furkatgofurov7 (commented):
LGTM thanks for the PR!  One question/suggestion inline:

test/e2e/const.go
yiannistri: Well spotted, I've now removed that label and associated code.

alexander-demicev (approved):
thanks!

furkatgofurov7 (approved):
Thanks!

PR #1170, opened 2025-03-20
Embed example clusterclasses in e2e suite

alexander-demicev:

Embed example clusterclasses in e2e suite, this should the code more readable

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/examples/examples.go
+++ b/examples/examples.go
@@ -1,0 +1,30 @@
+/*
+Copyright © 2025 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


anmazzotti (approved):

examples/examples.go
Danil-Grigorev, line 1: If I get it right, examples is an external package and importing it will cause issues for users trying to import e2e suite externally.

alexander-demicev: It will require adding turtles to go mod too

alexander-demicev:
@anmazzotti but we can still set the separate clusterclass namespace with embedded files

anmazzotti (approved):

examples/examples.go
Danil-Grigorev: That’s probably fine. On the second look I see that it is only used in a test case, which in general are not imported on it’s own. We can add GitOps scenarios on top of this change.

Danil-Grigorev (approved):

PR #1171, opened 2025-03-20
Use Calico CNI and Azure CCM for Azure e2e

### LINKED ISSUE #1157: Investigate GitOps lifecycle
From original issue:

- [ ] Investigate GitOps lifecycle:
  - What happens on Cluster deletion? Is that going to be stuck, for example if the InfraTemplates or other resources are also deleted at the same time? (See: [commit](https://github.com/rancher-sandbox/rancher-turtles-fleet-example/commit/ce56455e24e0983954739f166c47f957e313203b))
  - ClusterClasses versioning and updates. Is creating a new ClusterClass and referencing it to an existing Cluster going to work? (See: [doc](https://cluster-api.sigs.k8s.io/tasks/experimental-features/cluster-class/change-clusterclass#effects-on-the-clusters))
  - What happens on Cluster deletion? Is that going to be stuck, for example if the InfraTemplates or other resources are also deleted at the same time? (See: [commit](https://github.com/rancher-sandbox/rancher-turtles-fleet-example/commit/ce56455e24e0983954739f166c47f957e313203b))
  - How to manage ClusterClasses lifecycle. Is creating a new ClusterClass and referencing it to an existing Cluster going to work? (See: [doc](https://cluster-api.sigs.k8s.io/tasks/experimental-features/cluster-class/change-clusterclass#effects-on-the-clusters))
  - How to manage ClusterClasses lifecycle
  - How to integrate Cluster API Add-on Provider for Fleet (CAAPF) to customize managed Clusters (ClusterResourceSet alternative) manage Clusters lifecycle

Danil-Grigorev:

This PR introduces changes enabled by latest version of `CAAPF` as means to aid provisioning and reduce boilerplate code across e2e and provide Fleet ready (tested) examples for the target audience reading `Turtles` documentation.

In particular, this change adds `Calico` as a CNI solution for `Azure` clusters, which is also a part of recommended `quick-start` process from the [docs](https://cluster-api.sigs.k8s.io/user/quick-start#deploy-a-cni-solution).

It also gives `CCM` for Azure, installable as a helm chart, which reduces templating and maintenance for this component in comparison to a full-sized `CRS`.

All functionality is tested in an e2e test employing a cross-namespaced `ClusterClass` reference.

Fixes #1157 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


anmazzotti:
@Danil-Grigorev beware of https://github.com/rancher/turtles/pull/1170 that this is conflicting with.
I do like this approach better btw @alexander-demicev as it does test the cross namespace functionality.

anmazzotti:
@Danil-Grigorev another change you may add is to remove the [hardcoded](https://github.com/rancher/turtles/blob/main/test/e2e/data/cluster-templates/azure-rke2-topology.yaml#L47) version in the Azure RKE2 topology.
This can now be set to `${KUBERNETES_VERSION}` for tests, as thanks to your changes we install `latest` Azure cloud provider.

test/e2e/helpers.go
```diff
@@ -55,16 +55,12 @@ type Setup struct {
...
-	if namespaceName == "" {
-		namespaceName = fmt.Sprintf("%s-%s", specName, util.RandomString(6))
-	}
 	namespace, cancelWatches := framework.CreateNamespaceAndWatchEvents(ctx, framework.CreateNamespaceAndWatchEventsInput{
 		Creator:             clusterProxy.GetClient(),
 		ClientSet:           clusterProxy.GetClientSet(),
-		Name:                namespaceName,
+		Name:                fmt.Sprintf("%s-%s", specName, util.RandomString(6)),
 		LogFolder:           filepath.Join(artifactFolder, "clusters", clusterProxy.GetName()),
 		IgnoreAlreadyExists: true,
```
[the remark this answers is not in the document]
Danil-Grigorev: This can be also updated upstream. Usage of generate name for namespace is replicating this manual behavior, and API server (since some time ago) will retry name generation on name conflict. So ignore is no longer needed.

Just a trivial change and a thought.

Updated in this PR.

alexander-demicev (approved):

anmazzotti (approved):
Looks good to me. 
There seem to be a potential race condition in CAAPF when the ClusterGroup namespace is also deleted.

Still I think this PR can be merged.
Thank you for the improvements!

test/e2e/data/cluster-templates/azure-rke2-topology.yaml
```diff
@@ -20,7 +20,9 @@ apiVersion: cluster.x-k8s.io/v1beta1
 kind: Cluster
 metadata:
   labels:
+    cluster-api.cattle.io/rancher-auto-import: "true"
```
salasberryfin, line 23: I feel like this could conflict with the current test spec, which labels the namespace for auto-import. Specifically, the Azure RKE2 test passes `LabelNamespace=true` https://github.com/rancher/turtles/blob/db38a7c0ee2a002fa70f98758b5d4b05c3175e9b/test/e2e/suites/import-gitops-v3/import_gitops_v3_test.go#L229 which means the spec will label the namespace during test execution https://github.com/rancher/turtles/blob/db38a7c0ee2a002fa70f98758b5d4b05c3175e9b/test/e2e/specs/import_gitops_mgmtv3.go#L227 Will this be an issue? Also, this does not apply to this specific test case but, if we use this approach for other scenarios, I think it may break the re-import validation.

salasberryfin (commented):
LGTM overall but I added a comment to double-check how we label namespaces/clusters for import.

test/e2e/data/cluster-templates/azure-rke2-topology.yaml
anmazzotti: FYI this template is not directly exposed to documentation (only the Classes are reused).
In `turtles-docs` there is a duplication of it, so it's fine if this does not have the import label, feel free to modify anything.

That said the label is redundant, but should not influence tests.

furkatgofurov7 (approved):

PR #1203, opened 2025-03-25
Use HelmApp for installing Calico on example Class

anmazzotti:
This PR makes consistent use of HelmApp to install Calico when using the example Classes.
This is to avoid confusion with the `fleet.yaml` alternative.

Test run: https://github.com/rancher/turtles/actions/runs/14083478348/job/39441876036

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


examples/applications/cni/calico/fleet.yaml
```diff
@@ -1,27 +0,0 @@
-helm:
-  releaseName: projectcalico
-  repo: https://docs.tigera.io/calico/charts
-  chart: tigera-operator
-  templateValues:
-    installation: |-
-      cni:
-        type: Calico
-        ipam:
-          type: HostLocal
```
Danil-Grigorev, line 10: It feels like this may be the source of the issues with original chart. +1 on being removed.

test/e2e/config/operator.yaml
```diff
@@ -58,7 +58,7 @@ variables:
   HELM_EXTRA_VALUES_FOLDER: "/tmp"
 
   # Rancher Configuration
-  RANCHER_VERSION: "v2.10.2"
+  RANCHER_VERSION: "v2.11.0-rc7"
   RANCHER_ALPHA_VERSION: "v2.11.0-alpha8"
```
Danil-Grigorev, line 62: BTW - we need to revise how e2e test is installing rancher. It is currently locked to alpha version here.

examples/applications/cni/calico/fleet.yaml
anmazzotti: Indeed documentation says `HostLocal` is for GKE, otherwise it should default to `Calico`, so I omitted it.

test/e2e/suites/import-gitops-v3/import_gitops_v3_test.go
```diff
@@ -215,11 +213,6 @@ var _ = Describe("[Azure] [RKE2] - [management.cattle.io/v3] Create and delete C
 			Namespace:    topologyNamespace,
```
Danil-Grigorev, line 213:
```suggestion
			TargetNamespace:    topologyNamespace,
```

Since it has to be applied to local cluster and used from CC namespace.

Danil-Grigorev (approved):

furkatgofurov7 (approved):

PR #1240, opened 2025-04-02
Allow CAPZ cluster replicas change using example template

Danil-Grigorev:

- Remove variables affecting cluster class templating process and preventing further cluster scaling.
- Replace imageVersion with automatic jsonPatch based on builtin cluster version.
- Use default registration method to allow later change to control-plane-endpoint

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/examples/clusterclasses/azure/clusterclass-rke2-example.yaml
+++ b/examples/clusterclasses/azure/clusterclass-rke2-example.yaml
@@ -84,14 +84,6 @@ spec:
           description: "The image name"
           type: string
           default: "capi-ubun2-2404"          
-    - name: imageVersion
-      required: true
-      schema:
-        openAPIV3Schema:
-          description: "The version of the image."
-          type: string
-          default: "1.31.1"
-          example: "latest"
     - name: vmSize
       required: true
       schema:
@@ -124,7 +116,6 @@ spec:
               path: "/spec/template/spec/identityRef/name"
               valueFrom:
                 variable: azureClusterIdentityName
-            # Builtins
             - op: add
               path: "/spec/template/spec/networkSpec/vnet/name"
               valueFrom:
@@ -165,10 +156,10 @@ spec:
               path: "/spec/template/spec/image/computeGallery/name"
               valueFrom:
                 variable: imageName
-            - op: add
-              path: "/spec/template/spec/image/computeGallery/version"
+            - op: replace
+              path: /spec/template/spec/image/computeGallery/version
               valueFrom:
-                variable: imageVersion
+                template: '{{ trimPrefix "v" (trimSuffix "+rke2r1" .builtin.cluster.topology.version) }}'
             - op: add
               path: "/spec/template/spec/vmSize"
               valueFrom:
@@ -198,6 +189,8 @@ spec:
         - name: control-plane-subnet
           role: control-plane
         - name: node-subnet
+          natGateway:
+            name: node-natgateway
           role: node
 ---
 apiVersion: controlplane.cluster.x-k8s.io/v1beta1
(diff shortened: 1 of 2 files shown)


examples/clusterclasses/azure/clusterclass-rke2-example.yaml
```diff
@@ -207,7 +194,6 @@ metadata:
 spec:
   template:
     spec:
-      registrationMethod: internal-first
```
anmazzotti, line 210: Does this need to be fixed upstream first in order to work? I think that by using control-plane-endpoint it won't work right now.

examples/clusterclasses/azure/clusterclass-rke2-example.yaml
```diff
@@ -165,10 +152,10 @@ spec:
               path: "/spec/template/spec/image/computeGallery/name"
               valueFrom:
                 variable: imageName
-            - op: add
-              path: "/spec/template/spec/image/computeGallery/version"
+            - op: replace
+              path: /spec/template/spec/image/computeGallery/version
               valueFrom:
-                variable: imageVersion
+                template: '{{ trimPrefix "v" (trimSuffix "+rke2r1" .builtin.cluster.topology.version) }}'
```
alexander-demicev, line 158: 👍

anmazzotti: This is nice, but we can't really guarantee it works. 
The user should actually make check that the community image has that version. 
It should work in most cases, so I agree with merging it.

examples/clusterclasses/azure/clusterclass-rke2-example.yaml
Danil-Grigorev: With this change template can be changed to migrate to cp endpoint later, while current config is final and would require cluster reprovisioning.

examples/clusterclasses/azure/clusterclass-rke2-example.yaml
```diff
@@ -124,11 +116,6 @@ spec:
               path: "/spec/template/spec/identityRef/name"
               valueFrom:
                 variable: azureClusterIdentityName
-            # Builtins
-            - op: add
-              path: "/spec/template/spec/networkSpec/vnet/name"
```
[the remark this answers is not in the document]
Danil-Grigorev: Those values are immutable, and are set to empty by default on creation. I'm still following up on what's the issue, but this patch doesn't/wouldn't work.

examples/clusterclasses/azure/clusterclass-rke2-example.yaml
alexander-demicev: There can also be `+rke2r2` but it's rare and users can modify the example according to their needs

examples/clusterclasses/azure/clusterclass-rke2-example.yaml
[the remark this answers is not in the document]
Danil-Grigorev: It appears the issue was due to `natGateway` field elsewhere. I reverted it.

examples/clusterclasses/azure/clusterclass-rke2-example.yaml
Danil-Grigorev: Users can add optional variable on top to override this behavior. Patches are processed sequentially in the list

examples/clusterclasses/azure/clusterclass-rke2-example.yaml
anmazzotti: There is no defaulting for the registrationMethod however. 
If I follow the RKE2 Control plane controller it feels this will error out until the user goes ahead and manually patches the method to be `internal-first`. 
It won't work with anything else, will fail to provision as far I know.

alexander-demicev (approved):

anmazzotti (approved):

PR #1266, opened 2025-04-09
Ensure Cluster is Ready in e2e tests

anmazzotti:

This PR adds an additional check to verify the CAPI Clusters have the `Ready` condition.
This is to ensure they are fully provisioned.

It can be otherwise that the test will succeed only if the first control plane node is initialized, but all the others fail to join.

Did a first run without editing the Azure RKE2 Class, and it does fail as expected:
https://github.com/rancher/turtles/actions/runs/14360104898/job/40259416000

```
  [FAILED] Timed out after 2100.001s.
  CAPI Cluster should be Ready
  Expected success, but got an error:
      <*errors.errorString | 0xc0008d0aa0>: 
      Cluster is not Ready
      {
          s: "Cluster is not Ready",
      }
```

Running a second one (expected to succeed) after I added back the `registrationMethod: internal-first` to the Class: https://github.com/rancher/turtles/actions/runs/14362168776

The observed behavior when `registrationMethod` is not defined, is that all nodes fail to join the Cluster, with the rke2-agent failing with:

```
failed to get CA certs: Get \"https://my-control-plane-endpoint:9345/cacerts\": context deadline exceeded
```

A proper fix depends on https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/5525

Using  `registrationMethod: internal-first`  solves the provisioning problem, but will later fail on machine rollouts, so it's a temporary solution.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/test/e2e/specs/import_gitops_mgmtv3.go
+++ b/test/e2e/specs/import_gitops_mgmtv3.go
@@ -337,6 +337,28 @@ func CreateMgmtV3UsingGitOpsSpec(ctx context.Context, inputGetter func() CreateM
 		By("Running checks on Rancher cluster")
 		validateRancherCluster()
 
+		By("Waiting for the CAPI Cluster to be Ready")
+		Eventually(func() error {
+			if err := input.BootstrapClusterProxy.GetClient().Get(ctx, client.ObjectKeyFromObject(capiCluster), capiCluster); err != nil {
+				return fmt.Errorf("getting Cluster: %w", err)
+			}
+
+			readyCondition := conditions.Get(capiCluster, clusterv1.ReadyCondition)
+			if readyCondition == nil {
+				return fmt.Errorf("Cluster Ready condition is not found")
+			}
+
+			switch readyCondition.Status {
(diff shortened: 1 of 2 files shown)


test/e2e/specs/import_gitops_mgmtv3.go
```diff
@@ -337,6 +337,28 @@ func CreateMgmtV3UsingGitOpsSpec(ctx context.Context, inputGetter func() CreateM
 		By("Running checks on Rancher cluster")
 		validateRancherCluster()
 
+		By("Waiting for the CAPI Cluster to be Ready")
```
valaparthvi, line 340: Will this check also be proof enough that all the machines are in `Active` state?

anmazzotti: A Cluster should be Ready when the provisioning of all desired machines is complete, so yes.

valaparthvi (approved):

alexander-demicev (approved):

PR #1277, opened 2025-04-14
Add legacy examples folder

alexander-demicev:

Some providers are already moved to the new clusterclass setup but the examples should still be available in older versions of Turtles. I will create a follow-up docs PR later.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


examples/legacy/docker-rke2.yaml
```diff
@@ -0,0 +1,245 @@
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: Cluster
```
mjura, line 2: I think ClusterClass resource should be applied first, otherwise we will get WARNING that Cluster is trying to use non-existing ClusterClass

examples/legacy/vsphere-kubeadm.yaml
```diff
@@ -0,0 +1,1490 @@
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: Cluster
```
mjura, line 2: Again user will see WARNING that Cluster resource has reference to non-existing VSphereCluster

alexander-demicev:
@anmazzotti These are legacy to support older docs versions

anmazzotti:
Right, sorry, I thought they were the same as the new ones. 
Thank you.

anmazzotti (approved):

mjura (approved):

PR #1290, opened 2025-04-16
feat: add capa+cabpk clusterclass example and test

salasberryfin:
**What this PR does / why we need it**:

Add CAPA/CABPK ClusterClass example and cover cluster provisioning from this class in E2E tests.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1278

**Special notes for your reviewer**:

Sample nightly E2E run: https://github.com/rancher/turtles/actions/runs/14578369518

**Checklist**:

- [x] squashed commits into logical changes
- [ ] includes documentation
- [ ] adds unit tests
- [x] adds or updates e2e tests

PR #1300, opened 2025-04-21
 Use fleet for aws kubeadm e2e

### LINKED ISSUE #1144: Utilize Fleet for EC2 AWS Kubeadm example
We need to replace usage of `CRS` in EC2 AWS examples with Fleet bundles. This would allow to cover lifecycle of applications on the cluster, simplify maintainability of the manifests by reduced size and using Helm chart, and allow users to replace configurations on demand (Calico -> Cilium/Flannel).

salasberryfin:

~~**This PR depends on https://github.com/rancher/turtles/pull/1290**~~

Now that the CAPA+CABPK cluster class is available as an example and tested in the nightly E2E, this PR moves to using Fleet for deploying the required configuration in the cluster:
- CNI: Calico
- AWS EBS CSI Driver

Cloud Controller Manager is installed directly via Fleet Bundle as the existing Helm chart has been unmaintained for a long time. We can use this alternative method for now and perhaps contribute upstream so the chart is usable for our use case and we can get the configuration we need.

Fixes #1144 

The following E2E run validates these changes: https://github.com/rancher/turtles/actions/runs/14597461324

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


examples/clusterclasses/aws/clusterclass-kubeadm-example.yaml
```diff
@@ -160,20 +125,33 @@ spec:
...
+                protocol: tcp
+                toPort: 179
+              - description: IP-in-IP
+                fromPort: 0
+                protocol: ipinip
+                toPort: 0
+              - description: Calico Typha
+                fromPort: 5473
+                protocol: tcp
+                toPort: 5473
```
Danil-Grigorev, line 146: What is the purpose of this change? Provided Calico CNI configuration uses no encapsulation here, which is recommended by upstream documentation. Shouldn’t this be dropped?

examples/applications/ccm/aws/fleet-bundle.yaml
```diff
@@ -0,0 +1,190 @@
...
+              - key: node.cloudprovider.kubernetes.io/uninitialized
+                value: "true"
+                effect: NoSchedule
+              - key: node-role.kubernetes.io/control-plane
+                effect: NoSchedule
+            affinity:
+              nodeAffinity:
+                requiredDuringSchedulingIgnoredDuringExecution:
+                  nodeSelectorTerms:
+                    - matchExpressions:
```
alexander-demicev, line 36: is this compatible with RKE2? if not we can use these selectors
```
nodeSelectorTerms:
- matchExpressions:
  - key: node-role.kubernetes.io/control-plane
    operator: Exists
- matchExpressions:
  - key: node-role.kubernetes.io/master
    operator: Exists
```

examples/clusterclasses/aws/clusterclass-kubeadm-example.yaml
salasberryfin: These are the rules we had to add to make Calico work with AWS/Kubeadm provisioning. The logs were reporting that these needed to be allowed in the security group.

alexander-demicev: BGP and IP-in-IP are the default rules assigned anyway, and an additional one is needed for Typha to function properly, as it's used by Calico when installed with the helm chart

examples/applications/ccm/aws/fleet-bundle.yaml
salasberryfin: This is a separate file that's used for Kubeadm only (for now). I'm planning on taking on RKE2 after this is merged and we can review how to unify both cloud controller manager configurations into a single file.

alexander-demicev (approved):

anmazzotti (approved):

PR #1321, opened 2025-04-25
test: increase cpu for aws/rke2 test

salasberryfin:

Looks like AWS/RKE2 tests are intermittently failing and the workload seems to have a very high CPU usage. In some of executions where an error occurs, the cluster is provisioned successfully, the Rancher agent is deployed and validated and the cluster is fully functional. Then, before finishing all checks, it just stops responding. This would also explain why this error happens randomly and some tests just pass.

To remediate this, we can go up in the size of the EC2 instances used. To avoid increasing costs for AWS/Kubeadm, where we don't encounter this issue, a new variable is added to set the machine type for AWS/RKE2 only.

Fixes #

Let's run a full nightly E2E to validate that this is in fact solving the issue.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/test/e2e/config/operator.yaml
+++ b/test/e2e/config/operator.yaml
@@ -51,6 +51,8 @@ variables:
   KUBERNETES_MANAGEMENT_AWS_REGION: "eu-west-2"
   AWS_CONTROL_PLANE_MACHINE_TYPE: "t3.large"
   AWS_NODE_MACHINE_TYPE: "t3.large"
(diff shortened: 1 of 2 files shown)


alexander-demicev:
The issue is most likely not related to CPU but to DNS

anmazzotti (approved):

furkatgofurov7 (approved):
This is just using same machine type but with bigger specs for rke2 tests only.

PR #1322, opened 2025-04-25
Use rancherInstalled value for FleetAddon ConfigMap sync

anmazzotti:
Only use CAAPF ConfigMap sync if Rancher is installed.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/rancher-turtles/templates/addon-provider-fleet.yaml
+++ b/charts/rancher-turtles/templates/addon-provider-fleet.yaml
@@ -31,12 +31,14 @@ data:
     spec:
       config:
         featureGates:
+        {{- if index .Values "rancherTurtles" "rancherInstalled" }}
           configMap:
             ref:
               kind: ConfigMap
               apiVersion: v1
               name: rancher-config
               namespace: cattle-system
+        {{- end }}
           experimentalOciStorage: true
           experimentalHelmOps: true
       clusterClass:


alexander-demicev (dismissed):

charts/rancher-turtles/templates/addon-provider-fleet.yaml
```diff
@@ -31,12 +31,14 @@ data:
     spec:
```
Danil-Grigorev, line 31: Technically we may later introduce `install` section, gated behind `rancherInstalled` value or similar, which will take care of fleet installation in such clusters. It may need some `APIserver` or `CA` logic to configure fleet.

Otherwise in a non rancher scenario `featureGates` section is ignored, and it is up to users to set them up.

Also interesting what happens if turtles is upgraded to `rancherInstalled` true, or if `rancherInstalled` is just a flag to not expect rancher *yet*, but it would come up shortly after. But it is just some thoughts.

LGTM

furkatgofurov7 (approved):

charts/rancher-turtles/templates/addon-provider-fleet.yaml
hardys: @Danil-Grigorev FYI in the edge wrapper chart we set `rancherInstalled` to false only because both turtles and rancher are installed via the RKE2 helm controller (so we can't control the ordering), so in this scenario rancher is just not installed _yet_ - it would be great to discuss this use-case and how we may best handle it, AFAICS we'll have to revert this change in the meantime?

Issue #1268, opened 2025-04-09
Deprecate usage of --insecure-skip-verify=true across test environments


anmazzotti:
In e2e and dev-env (and quickstart script), we use `--set rancherTurtles.managerArguments[0]=--insecure-skip-verify=true"}` when installing Turtles.

This has the very nice effect that Cluster import always works without problems.
However we may want to surface problems in most cases and this is not close to anything production like.

A valid alternative should be to use the new `agent-tls-mode` feature gate, which should also work with self signed Rancher CA when `agent-tls-mode` is `Strict` (default)

This should be verified and if so the insecure-skip-verify option should be removed from all the environments.
Possibly this could deserve some specific documentation page to clarify how exactly the Rancher `agent-tls-mode` is supported by Turtles and what effects does it have (for example when importing/installing the cattle-cluster-agent).

PR #1326, opened 2025-04-28
:seedling: Promote agent-tls-mode feature to beta

### LINKED ISSUE #1318: Promote agent tls mode to Beta
Promote agent tls mode to beta by enabling it by default in helm values and opening a docs PR

yiannistri:

Promoted agent-tls-mode to beta so it is now enabled by default. Note that the tests do not validate the behaviour when the mode is set to `strict` as that would also require a custom CA to be used which complicates test setup.

Fixes #1318 

Related docs PR: https://github.com/rancher/turtles-docs/pull/294

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/rancher-turtles/values.yaml
+++ b/charts/rancher-turtles/values.yaml
@@ -39,10 +39,10 @@ rancherTurtles:
       etcdBackupRestore:
         # enabled: Turn on (true) or off (false).
         enabled: false
-    # agent-tls-mode: Alpha feature for agent TLS.
+    # agent-tls-mode: Beta feature for agent TLS.
     agent-tls-mode:
       # enabled: Turn on or off.
-      enabled: false
+      enabled: true
     # clusterclass-operations: Alpha feature. Manages cluster class ops. Not ready for testing yet.
     clusterclass-operations:
       # enabled: Turn on or off.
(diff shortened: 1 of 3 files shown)


anmazzotti:
This is great, but I still miss the documentation regarding this feature (or any other feature to be honest).
Possibly related to https://github.com/rancher/turtles/issues/1268
Not sure we should scope creep, but I also don't like the idea of turning this out of Alpha without any user documentation at all. @alexander-demicev what do you think?

alexander-demicev:
@anmazzotti we can merge this PR and create a new one for docs, it will get released at the end of the month, we still have time

anmazzotti:
> Would it be possible to update `make dev-env` so it uses system-store? This will make development easier

@alexander-demicev I think the setting we want is `Strict`, which should be default instead.
This should allow importing without the --insecure-skip-verify exploit. 

Still we have no solution for HelmApps, that indeed should be `SystemStore` driven, but there is no option to configure it as far I know. Possibly a feature request for Fleet.

anmazzotti (dismissed):

furkatgofurov7 (approved):
Documentation update is missing (at least not referenced yet), otherwise this LGTM thanks!

alexander-demicev:
Before merging, could you verify that upgrading from the previous version won't change the value to true?

mjura (approved):
LGTM

PR #1327, opened 2025-04-28
Use predictable resourceGroup for AKS nodes

anmazzotti:
Just a smaller quality of life changes. 
By default Azure creates a randomized resource group for AKS nodes, for example: `MC_my-resource-group-w77s7_westeurope`. 

By reusing the already defined `resourceGroup` variable, we can better scope this group name. 
Additionally, this change should now also cover the `resourceGroup` in the [cleanup job](https://github.com/rancher/turtles/blob/main/.github/workflows/e2e-cleanup.yaml#L42), which is not the case currently due to the `MC_` prefix.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/examples/clusterclasses/azure/clusterclass-aks-example.yaml
+++ b/examples/clusterclasses/azure/clusterclass-aks-example.yaml
@@ -118,6 +118,11 @@ spec:
               valueFrom:
                 variable: resourceGroup
             - op: add
+              path: "/spec/template/spec/nodeResourceGroupName"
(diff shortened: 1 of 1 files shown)


mjura (approved):
I am fine with this change, but I think we should give a possibility to use another variable `resourceGroupNodes` if not set then use `{{ .resourceGroup }}-nodes`

PR #1355, opened 2025-05-05
Enable long run in pre-submits via the merge-queue

Danil-Grigorev:

This change would automate the nightly run for every PR, instead of requiring the reviewer to attach one automatically. This check can be made mandatory to prevent merges with failing long e2e tests later, if needed.

Such change would eliminate situations when several independent PRs are merged simultaneously, and all have attached successful nightly runs, but in combination they cause nightlies to fail.

Merge queues perform checks in a sequential order, with all PRs currently in the queue rebased on top of each other. Therefore, it follows the established concurrency pattern, but in order to eliminate conflicts between regular nightly runs and merge queue ones, the `concurrency: ci_e2e_tests` is lifted, and each cluster uses a unique name now.

In addition to that, it uses a specific [temporary branch](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#triggering-merge-group-checks-with-third-party-ci-providers) created in the `rancher/turtles` repo. Therefore, a merge queue run has access to repository secrets as any other nightly run.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/e2e-long.yaml
+++ b/.github/workflows/e2e-long.yaml
@@ -1,11 +1,16 @@
-name: Run nightly e2e tests
+name: Run long e2e tests
 
 on:
   schedule:
     - cron: "0 0 * * *"
   workflow_dispatch:
+  merge_group:
 
-concurrency: ci_e2e_tests
+concurrency:
+  group: ci-${{ github.ref }}
+
+env:
+  GINKGO_LABEL_FILTER: ${{ github.event.pull_request.labels && (contains(join(github.event.pull_request.labels.*.name, ','), 'test/full') && 'full' || '') || 'full' }}
 
 jobs:
   publish_e2e_image:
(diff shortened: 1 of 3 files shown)


mjura (dismissed):

anmazzotti (changes requested):
On a second thought, I think now this is a good idea.
However we have a couple of tests failing:
- [AWS] [EC2 Kubeadm] (needs to be fixed)
- [GCP] [GKE] (often flaky)

I think these needs to be fixed or skipped, or this workflow to be made optional.

Also a small nitpick, the workflow can also be renamed to `Run long e2e tests` instead, since it's no longer strictly `nightly`

alexander-demicev:
after this change, can someone accidentally start a second nightly run manually?

salasberryfin:
I think this is a nice addition, as short tests are not enough to validate functionality after certain code changes but I'm a bit concern about the security implications of this. We'll need to be extra careful with what's pushed to a PR (branch), as it will translate into creating "real-life" paid resources (which can potentially be publicly accessible if not configured properly?).

We'll also need to accept that fast development (i.e. quick fixes) will be affected by this.

Other than that, I'd vote for having this merged.

salasberryfin (approved):

Danil-Grigorev:
I’ll address cancel-in-progress and concurrency ambiguity in a follow-up PR.

PR #1362, opened 2025-05-09
add docker rke2 cluster example

salasberryfin:

Add Docker/RKE2 pre-configured example to `examples/clusters` that'll be used in a demo video for the docs. This is taken from the existing cluster template in `test/e2e/data/cluster-templates`, just replacing the environment variables.

Fixes #

As requested by @anmazzotti, the config map definition is removed from the cluster definition and added as a separate "application" template under `examples/applications`. This will make it easier to reference in the docs, too.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


furkatgofurov7 (approved):

examples/clusters/docker/rke2/cluster.yaml
```diff
@@ -0,0 +1,100 @@
...
+
+    backend rke2-servers
+      option httpchk GET /v1-rke2/readyz
+      http-check expect status 403
+      {{range $server, $backend := .BackendServers }}
+      server {{ $server }} {{ $backend.Address }}:9345 check check-ssl verify none
+      {{- end}}
+kind: ConfigMap
+metadata:
+  name: docker-rke2-lb-config
```
anmazzotti, line 98: Since this configmap does not require templating, may it be moved to examples/applications instead?
I think this is also going to make the [docs](https://github.com/rancher/turtles-docs/blob/main/docs/next/modules/en/pages/user/clusterclass.adoc?plain=1#L653) simpler, since it can just be applied.

anmazzotti (changes requested):

mjura (approved):

examples/clusters/docker/rke2/cluster.yaml
salasberryfin: I will open a separate PR as this is out of the scope of this issue. Then we can put this PR on hold until the other is merged.

mjura (approved):

PR #1479, opened 2025-06-19
Add new example for CAPG/Kubeadm

### LINKED ISSUE #1478: Add self-managed GCP cluster
# Description

As of today, only GCP managed clusters (GKE) are covered in both `./examples` and E2E tests. We should also add an additional use case for GCP self-managed, for now using Kubeadm.

salasberryfin:

This PR adds a new `ClusterClass` example based on CAPG+Kubeadm. It also includes a new test case that covers the new example.

Two applications need to be deployed for the cluster to be functional:
- CNI: reusing the existing Calico `HelmApp`.
- [GCP Cloud Controller Manager](https://github.com/kubernetes/cloud-provider-gcp) does not provide a Helm chart, so a `Bundle` is used instead.

Fixes #1478 

Following up on this, I'll add this use case to the documentation.

Reference E2E execution: https://github.com/rancher/turtles/actions/runs/15827179643

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


examples/clusterclasses/gcp/kubeadm/clusterclass-kubeadm-example.yaml
```diff
@@ -0,0 +1,256 @@
...
+      required: true
+      schema:
+        openAPIV3Schema:
+          type: string
+    - name: workerImageId
+      required: true
+      schema:
+        openAPIV3Schema:
+          type: string
+    - name: controlPlaneMachineType
```
anmazzotti, line 69: I think there no need to distinguish `controlPlaneMachineType` and `workerMachineType`. Just one variable would fit, and it can be [overridden](https://cluster-api.sigs.k8s.io/tasks/experimental-features/cluster-class/write-clusterclass#machinedeployment-and-machinepool-variable-overrides) on machinedeployments.

I think the same applies for `controlPlaneImageId` and `workerImageId`

anmazzotti (approved):
Nitpicking on the variables definition, I would appreciate to see that simplified if possible.
Rest looks all good!
If you could also run this branch on the e2e tests, it would be perfect.
Thank you.

examples/clusterclasses/gcp/kubeadm/clusterclass-kubeadm-example.yaml
anmazzotti: We don't run the e2e on merge yet. Definitely too unstable for that.

You can push the same branch upstream, no need to open a new PR, just remember to delete it manually after please.

anmazzotti (dismissed):

salasberryfin:
Running the E2E from GitHub helped realize that there was a new secret missing `GCP_IMAGE_ID`. It's now added and a new workflow is running :crossed_fingers:

alexander-demicev:
Looks great, I will approve once e2e suite passes

furkatgofurov7 (approved):
Great work, thanks (looking fwd to documentation PR)!

salasberryfin:
This is now passing the nightly E2E: https://github.com/rancher/turtles/actions/runs/15827179643

anmazzotti (approved):

PR #1522, opened 2025-07-08
Remove Tilt from dev-env deployment

anmazzotti:
`make dev-env` differs in the way the containers are deployed from e2e environment, since we are using Tilt instead.
This difference needs to be addressed every time we modify containers, like adding new mounts. 

It would be better to simply install the chart and rely on it. This PR also makes the assumption the Tilt setup is something that no one uses normally, therefore it's simpler to just clean it up.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


alexander-demicev:
I'm using Tilt regularly, I love the hot reload feature

anmazzotti:
> I'm using Tilt regularly, I love the hot reload feature

Ok then I may as well just fix it.

PR #1549, opened 2025-07-16
Cleanup vsphere self-hosted runner

### LINKED ISSUE #1540: Add vSphere cleanup steps to self-hosted runner
Enhance vsphere wf https://github.com/rancher/turtles/blob/main/.github/workflows/run-vsphere-tests.yaml

Minimal cleanup step would be:
```
docker system prune --all --volumes --force
```
But this cleans up also `kind` network which should be recreated by:
```
docker network create \
        --driver=bridge \
        --subnet=172.19.0.0/16 \
        --gateway=172.19.0.1 \
        --opt "com.docker.network.bridge.enable_ip_masquerade"="true" \
        --opt "com.docker.network.driver.mtu"="1500" \
        kind
```

thehejik:
Cleans up docker artefacts after performing e2e tests / without this we would be hitting `no space left` soon.

Fixes #1540

* I decided to use `run:` block for the cleanup as the self-hosted runner is only temporary and should be replaced by hosted runner where the cleanup won't be needed.
* Cleans up all docker resources: containers, images, docker build cache, volumes and networks.
* Re-creates kind network needed by kind cluster.

**Checklist**:

- [x] testrun https://github.com/rancher/turtles/actions/runs/16320506448

### DIFF
--- a/.github/workflows/run-vsphere-tests.yaml
+++ b/.github/workflows/run-vsphere-tests.yaml
@@ -7,8 +7,8 @@ on:
   workflow_dispatch:
 
 env:
-  VSPHERE_TLS_THUMBPRINT: ${{ secrets.VSPHERE_TLS_THUMBPRINT }} 
-  VSPHERE_SERVER: ${{ secrets.VSPHERE_SERVER }} 
+  VSPHERE_TLS_THUMBPRINT: ${{ secrets.VSPHERE_TLS_THUMBPRINT }}
+  VSPHERE_SERVER: ${{ secrets.VSPHERE_SERVER }}
   VSPHERE_DATACENTER: ${{ secrets.VSPHERE_DATACENTER }}
   VSPHERE_DATASTORE: ${{ secrets.VSPHERE_DATASTORE }}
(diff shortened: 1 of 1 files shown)


furkatgofurov7 (commented):
@thehejik can you sign your commit?

.github/workflows/run-vsphere-tests.yaml
```diff
@@ -77,3 +77,18 @@ jobs:
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
           SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
+      - name: Docker cleanup
+        if: always()
+        run: |
+          CONTAINERS=$(docker ps --all --quiet)
```
alexander-demicev, line 83: can it somehow cause problems for e2e suites running in QA environment?

thehejik: QA is using another runner so this doesn't have any impact.

PR #1582, opened 2025-07-29
feat: Add post-upgrade hook that deletes all CAPI Operator provider resources

### LINKED ISSUE #1548: Remove CAPI Operator resources on turtles chart upgrade

yiannistri:

This PR adds a post-upgrade hook that removes all CAPI Operator provider resources from the cluster.

Fixes #1548 

To test this, run `RELEASE_TAG=v0.22.0 make release` then upgrade an existing Turtles installation by running 
```
helm upgrade rancher-turtles -n rancher-turtles-system --reset-then-reuse-values ./out/package/rancher-turtles-0.22.0.tgz --debug
```
to install the new chart version.

The `kubectl` image could not be used as it is distroless. 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/rancher-turtles/templates/post-upgrade-job.yaml
+++ b/charts/rancher-turtles/templates/post-upgrade-job.yaml
@@ -1,0 +1,109 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: post-upgrade-job
+  namespace: '{{ .Values.rancherTurtles.namespace }}'
+  annotations:
+    "helm.sh/hook": post-upgrade
+    "helm.sh/hook-weight": "1"
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: post-upgrade-job-delete-capi-operator-resources
+  annotations:
+    "helm.sh/hook": post-upgrade
+    "helm.sh/hook-weight": "1"
+rules:
+- apiGroups:
+  - operator.cluster.x-k8s.io
+  resources:
+  - addonproviders
+  - bootstrapproviders
+  - controlplaneproviders
+  - coreproviders
+  - infrastructureproviders
+  - ipamproviders
+  - runtimeextensionproviders
+  verbs:
+  - get
+  - watch
+  - list
+  - delete
+  - patch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: post-upgrade-job-capi-operator-resources-cleanup
+  annotations:
+    "helm.sh/hook": post-upgrade
+    "helm.sh/hook-weight": "1"
+subjects:
+  - kind: ServiceAccount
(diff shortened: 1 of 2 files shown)


salasberryfin (dismissed):
Thanks @yiannistri.

alexander-demicev (commented):
do we also need to remove finalisers?

yiannistri:
@alexander-demicev you are right, we should first remove the finalisers before deleting the CAPI Operator providers. However, we cannot do that in bulk, for all providers across all namespaces, using a simple `kubectl` command. Instead we should iterate through them one by one and patch them to remove the finaliser. Should I attempt to write a bash script that runs on the `kubectl` image that we use for other cleanups? Would that be an acceptable solution, if it works?

yiannistri:
Moving this back to draft in order to also handle removal of finalisers.

alexander-demicev (approved):

charts/rancher-turtles/templates/post-upgrade-job.yaml
```diff
@@ -0,0 +1,109 @@
...
+  annotations:
+    "helm.sh/hook": post-upgrade
+    "helm.sh/hook-weight": "1"
+data:
+  cleanup.sh: |
+    #!/usr/bin/env bash
+
+    set -euo pipefail
+
+    kubectl get addonproviders.operator.cluster.x-k8s.io --all-namespaces --no-headers --ignore-not-found | awk '{print $1 " " $2}' | xargs -r -n2 bash -c 'kubectl patch addonproviders.operator.cluster.x-k8s.io "$1" -n "$0" --type merge -p "{\"metadata\":{\"finalizers\":null}}"'
```
valaparthvi, line 65: You could add macro functions that accept the resource type to make this simpler, I think.

```sh

remove_finalizers_and_delete() {
  resource_type=$1
  kubectl get $resource_type --all-namespaces --no-headers --ignore-not-found | awk '{print $1 " " $2}' | xargs -r -n2 bash -c 'kubectl patch $resource_type "$1" -n "$0" --type merge -p "{\"metadata\":{\"finalizers\":null}}"'
  kubectl delete $resource_type
}

# List of resource types to process
resource_types=(
    "addonproviders.operator.cluster.x-k8s.io"
    "bootstrapproviders.operator.cluster.x-k8s.io"
    "controlplaneproviders.operator.cluster.x-k8s.io"
    "coreproviders.operator.cluster.x-k8s.io"
    "infrastructureproviders.operator.cluster.x-k8s.io"
    "ipamproviders.operator.cluster.x-k8s.io"
    "runtimeextensionproviders.operator.cluster.x-k8s.io"
)

# Loop through each resource type, call the finalizer removal macro, then the delete macro
for type in "${resource_types[@]}"; do
    remove_finalizers_and_delete "$type"
done
```

I haven't tested the script though.

valaparthvi (approved):

PR #1601, opened 2025-08-07
Implement CAPIProvider EnableAutomaticUpdate toggle

anmazzotti:

This PR does mainly two things: 
- Allows the upstream provider reconciler to patch the CAPIProviderSpec. It's more straightforward to simply rely on the upstream business logic for all the generic provider fields.
- Introduces a new feature `EnableAutomaticUpdate` on CAPIProviders. This feature is needed to clearly distinguish whether the user wants the providers to be automatically updated, or not. Previously this was done by leaving the CAPIProvider.Spec.Version blank. The behavior is going to change in the following way:

1. Users installing the default providers from the chart will not be affected (core, rke2, fleet) and will continue to experience automatic updates on Turtles updates (on ClusterctlConfig updates). 
2. Users who pinned the `Version` on their installed providers will not be affected. The new toggle is off by default, they have to opt-in explicitly.
3. Users who didn't pin the `Version` (previous automatic update toggle), will see their providers updating to latest (which should already be the case) and the `Version` then pinned. In order to restore automatic updates functionality, they have to explicitly opt-in with the new `EnableAutomaticUpdate` toggle. 

Documentation: https://github.com/rancher/turtles-docs/pull/351

Fixes #

The current implementation of CAPIProvider reconciler does not patch the CAPIProviderSpec. This has the consequence of never updating the `operator.cluster.x-k8s.io/applied-spec-hash` annotation, which will lead to the cache constantly being invalid. This PR originated from this issue and it fixes it.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/api/v1alpha1/capiprovider_types.go
+++ b/api/v1alpha1/capiprovider_types.go
@@ -59,6 +59,10 @@ type CAPIProviderSpec struct {
 	// +kubebuilder:example={CLUSTER_TOPOLOGY:"true",EXP_CLUSTER_RESOURCE_SET:"true",EXP_MACHINE_POOL: "true"}
 	Variables map[string]string `json:"variables,omitempty"`
 
+	// EnableAutomaticUpdate can be used to automatically update the CAPIProvider to a newest version.
+	// +optional
+	EnableAutomaticUpdate bool `json:"enableAutomaticUpdate,omitempty"`
+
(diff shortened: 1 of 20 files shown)


mjura (approved):
Changes which are removing some bits of code are always the best and welcome :D

alexander-demicev (approved):

mjura (approved):
LGTM

PR #1652, opened 2025-08-25
ADR 0016: CAPI version pinning strategy

alexander-demicev:

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/docs/adr/0016-capi-version-pinning.md
+++ b/docs/adr/0016-capi-version-pinning.md
@@ -1,0 +1,61 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [CAPI version pinning strategy (community vs prime)](#capi-version-pinning-strategy-community-vs-prime)
+  - [Context](#context)
+  - [Options](#options)
+  - [Decision](#decision)
+  - [Open questions](#open-questions)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# CAPI version pinning strategy (community vs prime)
+
+- Status: proposed
+- Date: 2025-08-25
+- Authors: [@alexander-demicev]
+- Deciders: [TBD]
+
+## Context
+
+We need a clear strategy for pinning the Core CAPI version and CAPI provider versions across community and prime builds. Today, Turtles embeds a `clusterctl` configuration as `config.yaml` (see `internal/controllers/clusterctl/config.yaml`, embedded via `internal/controllers/clusterctl/config.go`). The question is where those pins should live and how we differentiate community vs prime delivery.
+
+## Options
+
+1) Keep existing `config.yaml`, switch by build tag
+- Keep an embedded `config.yaml` but provide a Go build tag to select the correct version:
+  - Prime version: pin Core CAPI and providers to our forks.
+  - Community version: pin only Core CAPI to upstream.
+- Pros: possbily smallest change out of all options, easy to configure.
+- Cons: difficult to explain and document how it's combined with ClusterctlConfig.
+
+2) No provider pinning in Turtles, use separate providers Helm chart with [ClusterctlConfig](https://turtles.docs.rancher.com/turtles/stable/en/reference/clusterctlconfig.html)
+- Do not hardcode provider pins in Turtles. Create a separate Helm chart for providers and do the pinning there via `ClusterctlConfig` resource. For prime users, make this chart mandatory to use CAPI providers. Community users are on their own to manage provider versions.
+- Pros: clean separation, providers can be updated independently of Turtles code.
+- Cons: introduces a strongly required dependency for provider pinning functionality, still need to figure out core CAPI pinning for community.
+
+3) Create two Turtles charts (community and prime), move pinning entirely to these charts
+- Do not embed `config.yaml` in Turtles. Provide two Helm charts:
+  - Community chart with a `ClusterctlConfig` where core CAPI is pinned to upstream.
+  - Prime chart with a `ClusterctlConfig` pinned to Rancher forks.
+- Pros: even cleaner separation, providers can be updated independently of Turtles code.
+- Cons: more packaging/release complexity, users must adopt charts for installation.
+
+4) Create one Turtles chart (community && prime) and one Turtles Certified Providers chart (prime only). 
+- Core CAPI provider is installed by the Turtles chart and always pinned for all users. 
+- Core CAPI provider uses the upstream ghcr.io registry by default, however this can be configured as a chart value, so when Rancher installs the Turtles chart, it can override the image to registry.suse.com (and configure the chart for Prime in other ways via values)
+- Turtles Certified Providers chart can be only published for prime users. It can load all manifests via oci, it will have all versions pinned. Community users can pin their own versions using CAPIProvider. 
+- Pros: It's one build, both for Turtles binary and for the Turtles chart. No shenanigans, no hardcoded embedded configuration values. Everything is configured at the chart level, fully transparent to the users.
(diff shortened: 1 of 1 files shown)


docs/adr/0016-capi-version-pinning.md
```diff
@@ -0,0 +1,51 @@
...
+- Pros: clean separation, providers can be updated independently of Turtles code.
+- Cons: introduces a strongly required dependency for provider pinning functionality, still need to figure out core CAPI pinning for community.
+
+3) Create two Turtles charts (community and prime), move pinning entirely to these charts
+- Do not embed `config.yaml` in Turtles. Provide two Helm charts:
+  - Community chart with a `ClusterctlConfig` where core CAPI is pinned to upstream.
+  - Prime chart with a `ClusterctlConfig` pinned to Rancher forks.
+- Pros: even cleaner separation, providers can be updated independently of Turtles code.
+- Cons: more packaging/release complexity, users must adopt charts for installation.
+
```
anmazzotti, line 43:
```suggestion

4) Create one Turtles chart (community && prime) and one Turtles Certified Providers chart (prime only). 
- Core CAPI provider is installed by the Turtles chart and always pinned for all users. 
- Core CAPI provider uses the upstream ghcr.io registry, however this can be configured as a chart value, so when Rancher installs the Turtles chart, it can override the image to registry.suse.com (and configure the chart for Prime in other ways via values)
- Turtles Certified Providers chart can be only published for prime users. It can load all manifests via oci, it will have all versions pinned. Community users can pin their own versions using CAPIProvider. 
- Pros: It's one build, both for Turtles binary and for the Turtles chart. No shenanigans, no hardcoded embedded configuration values. Everything is configured at the chart level, fully transparent to the users.
- Cons: None, this is the best idea ever.
```

anmazzotti: I see, so the idea is that the **version** comes from Turtles, the providers chart is not meant to bring updated versions of the providers. But then are users supposed to install the providers chart only once? Or are we going to have a separate lifecycle for this separate chart? Like release only when needed?

I did assume the providers chart would follow Turtles release. This does imply however that the user also needs to update this chart after updating Rancher. We can still make this automatic however, the Rancher controller can use a new Setting to determine whether to install or not the providers chart (and keep it up to date together with Turtles).

anmazzotti:
Just to add another option, the KDM meta-data has been mentioned before and it could be another way of pinning providers: https://releases.rancher.com/kontainer-driver-metadata/release-v2.11/data.json

docs/adr/0016-capi-version-pinning.md
alexander-demicev: We can keep the providers chart release separate and only bump it when needed. It doesn’t have to be tied to Turtles.
For installation, using `helm install` is a simpler path for this chart compared to introducing a dedicated setting. We can cover the details in a separate ADR, this one should stay focused on where the CAPI and provider version pinning will happen.

alexander-demicev: @furkatgofurov7 How would UX look like in this case? Do we keep `ClusterctlConfig` for pinning? I'm trying to keep the user interaction as simple as possible. Are we ok with complicating the release process and having to maintain 2 release workflows for Turtles chart long term?

furkatgofurov7: I had a feeling we can't have the best of both worlds (user experience and cleanest separation at once), hence I was in favor of the latter.

docs/adr/0016-capi-version-pinning.md
```diff
@@ -0,0 +1,51 @@
...
+We need a clear strategy for pinning the Core CAPI version and CAPI provider versions across community and prime builds. Today, Turtles embeds a `clusterctl` configuration as `config.yaml` (see `internal/controllers/clusterctl/config.yaml`, embedded via `internal/controllers/clusterctl/config.go`). The question is where those pins should live and how we differentiate community vs prime delivery.
+
+## Options
+
+1) Keep existing `config.yaml`, switch by build tag
+- Keep an embedded `config.yaml` but provide a Go build tag to select the correct version:
+  - Prime version: pin Core CAPI and providers to our forks.
+  - Community version: pin only Core CAPI to upstream.
+- Pros: possbily smallest change out of all options, easy to configure.
+- Cons: difficult to explain and document how it's combined with ClusterctlConfig.
```
salasberryfin, line 30: This, I assume, is a problem if users want to interact with different versions, as in choosing a provider version different to the one pinned. Is that the case? I don't see much difficulty otherwise, since pinning is done on the time of release by the Turtles team.

It may not be perfect nor the most scalable solution, but I think this is a good compromise between simplicity and functionality, unless we want to give an improved UX for customizing provider versions. However, since the prime differentiator here is that we pin tested versions, I don't know if this scenario is in fact that relevant. What do you think?

docs/adr/0016-capi-version-pinning.md
salasberryfin: Now that Turtles is becoming a part of Rancher, having a completely separate chart would mean that we also need to publish this chart to the Rancher repos, as in updating `rancher/charts`, etc. To me that sounds like too much extra work for a minuscule chart change and, other than a more clear organization, I don't see many benefits.

docs/adr/0016-capi-version-pinning.md
```diff
@@ -0,0 +1,51 @@
...
+
+2) No provider pinning in Turtles, use separate providers Helm chart with [ClusterctlConfig](https://turtles.docs.rancher.com/turtles/stable/en/reference/clusterctlconfig.html)
+- Do not hardcode provider pins in Turtles. Create a separate Helm chart for providers and do the pinning there via `ClusterctlConfig` resource. For prime users, make this chart mandatory to use CAPI providers. Community users are on their own to manage provider versions.
```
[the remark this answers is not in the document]
salasberryfin: This sounds like the best approach to me if we consider that the existing embedded `config.yaml` is not good enough.

alexander-demicev:
@furkatgofurov7 @salasberryfin @anmazzotti I pushed some updates and decision section, PTAL and resolve discussions if you're ok with that.

docs/adr/0016-capi-version-pinning.md
```diff
@@ -0,0 +1,61 @@
...
+- Cons: Introduces complexity with multiple charts, requiring users to install and manage additional components beyond a simple CAPIProvider CR. On Rancher updates, providers should update automatically based on new pinned versions, but with separate charts, this may require manual intervention or additional steps.
+
+## Decision
+
+- Use Go build tags to pin CAPI providers for community and Prime builds.
+- If the Go build tag approach for pinning providers doesn't work well, switch to pinning with Helm charts. Since the provider chart is already merged, it won't be hard to add this functionality.
+- To improve the air-gapped experience with Helm chart changes in Turtles, we will add a ConfigMap with bundled core CAPI components to the Turtles chart. This would make Rancher air-gapped installs easier by removing the need for any OCI usage for core CAPI, as all manifests would be bundled with Rancher through Turtles. Users would still need to handle image mirroring, as they do now.
```
salasberryfin, line 55: I don't fully understand what the implications of these components being embedded into Turtles chart are. Does this mean that all core CAPI CRDs are stored in a ConfigMap when installing the chart the same way CAPI operator stores the manifest's components? If this only affects core CAPI, it means that for any other type of provider users will need to handle image pulling.

salasberryfin (commented):
Overall looks good to me. I just added a comment to understand one of the points. Not blocking at all.

docs/adr/0016-capi-version-pinning.md
alexander-demicev: We will provide a cache configmap so Turtles doesn't have to create it's own. In this case airgapped installation won't have problem with github access and users won't need to bump to use OCI for core.

docs/adr/0016-capi-version-pinning.md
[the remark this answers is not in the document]
alexander-demicev: @salasberryfin can this be resolved?

salasberryfin (approved):

PR #1818, opened 2025-10-13
fix: Change Turtles namespace to `cattle-turtles-system`

yiannistri:

Updated default Turtles namespace to `cattle-turtles-system` in order for Rancher UI to treat its resources as system resources rather than user resources. Once this PR is merged we also need to review/merge https://github.com/rancher/rancher/pull/52294 as well.
 
Part of https://github.com/rancher/rancher/issues/52286

**Special notes for your reviewer**: N/A

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/test/e2e/data/test-providers/unknown-provider.yaml
+++ b/test/e2e/data/test-providers/unknown-provider.yaml
@@ -10,6 +10,6 @@ metadata:
   name: vcluster
   namespace: unknown-provider
 spec:
-  enableAutomaticUpdate: true # Since this provider is uknown, enabling automatic updates should have no effect.
+  enableAutomaticUpdate: true # Since this provider is unknown, enabling automatic updates should have no effect.
   version: v0.2.1
   type: infrastructure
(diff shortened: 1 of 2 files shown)


yiannistri:
> Thanks @yiannistri.
> 
> I'm thinking about how this affects the migration script for `CAPIProviders`
> 
> https://github.com/rancher/turtles/blob/17c01fc926cbad1548ad362071291183af597d5d/scripts/migrate-providers-ownership.sh#L43
> I'm assuming it's fine if we leave the reference to `rancher-turtles-system` as using the script means that users are on a previous version. Is that correct?

Yes, this variable specifically needs to be updated to the namespace that will be used to deploy the `rancher-turtles-providers` chart, which is why I've set that to `cattle-turtles-system`. If that's a mistake let me know.
The `TURTLES_CHART_NAMESPACE` variable however was left as `rancher-turtles-system` because that is the (default) namespace that we used to deploy providers in previous versions < v0.25.0.

salasberryfin (approved):

Issue #1816, opened 2025-10-13
Add support for installation of custom system-chart for turtles


cpinjani:
Add support for installation of custom system-chart for turtles, in order to test dev/nightly version from 2.13 onward.

Similar reference implementation of hosted providers:
- https://confluence.suse.com/spaces/EN/pages/1271562687/Testing+using+HEAD
- https://github.com/rancher/rancher/pull/41389

PR #1836, opened 2025-10-20
feat: Install Turtles as system chart in dev-env

anmazzotti:
This PR is related to https://github.com/rancher/turtles/issues/1816

Draft state as it's a proof of concept still.

Changes:

1. Fork and update the Rancher charts repo to add a dev build of Turtles
2. Add Gitea to the dev environment to serve Rancher system charts

How to test:
```bash
export RANCHER_HOSTNAME=my.test.example 
export NGROK_AUTHTOKEN=foo 
export NGROK_API_KEY=bar
make dev-env
```

### DIFF
--- a/scripts/turtles-dev.sh
+++ b/scripts/turtles-dev.sh
@@ -22,14 +22,20 @@ if [ -z "$RANCHER_HOSTNAME" ]; then
     exit 1
 fi
 
-RANCHER_CHANNEL=${RANCHER_CHANNEL:-latest}
+RANCHER_CHANNEL=${RANCHER_CHANNEL:-alpha}
 RANCHER_PASSWORD=${RANCHER_PASSWORD:-rancheradmin}
-RANCHER_VERSION=${RANCHER_VERSION:-v2.12.1}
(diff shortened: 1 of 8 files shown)


main.go
```diff
@@ -181,7 +181,7 @@ func main() {
 	setupReconcilers(ctx, mgr)
 
 	// +kubebuilder:scaffold:builder
-	setupLog.Info("starting manager", "version", version.Get().String())
+	setupLog.Info("starting Testy MCtest manager", "version", version.Get().String())
```
alexander-demicev: is this here intentionally?

scripts/turtles-dev.sh
```diff
@@ -127,15 +162,15 @@ install_local_providers_chart() {
...
+# echo "Installing local Rancher Turtles Providers..."
+# install_local_providers_chart
 
-if [ "$USE_TILT_DEV" == "true" ]; then
-    echo "Using Tilt for development..."
-    tilt up
-fi
+# if [ "$USE_TILT_DEV" == "true" ]; then
+#     echo "Using Tilt for development..."
+#     tilt up
```
alexander-demicev: Does that work fine with tilt? doesn't system chart controller override the deployment?

PR #1850, opened 2025-10-30
fix: Set `securityContext` field to Turtles controller and hooks manifests

yiannistri:

This PR sets the `securityContext` field for all containers that are bundled in the Turtles chart (controller and Helm hooks).  This change will allow Turtles to be installed and run on hardened clusters. 

The removal of the conditional ns lookup was added because if a Turtles installation or upgrade fails and the namespace is still present in the cluster, then the next time Turtles get installed or upgrade, Helm will generate a diff that will cause the `cattle-capi-system` to be deleted.

Part of fixing https://github.com/rancher/rancher/issues/52501

When testing this, it is not enough to be able to install the chart, any Helm hooks should also run without errors caused after setting a `securityContext`. 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/rancher-turtles/templates/core-provider.yaml
+++ b/charts/rancher-turtles/templates/core-provider.yaml
@@ -1,14 +1,9 @@
 {{- if index .Values "cluster-api-operator" "cluster-api" "enabled" }}
-{{- $namespace := index .Values "cluster-api-operator" "cluster-api" "core" "namespace" }}
-{{- if not (lookup "v1" "Namespace" "" $namespace) }}
 ---
 apiVersion: v1
 kind: Namespace
 metadata:
-  annotations:
-    "helm.sh/hook-weight": "1"
   name: {{ index .Values "cluster-api-operator" "cluster-api" "core" "namespace" }}
-{{- end }}
 ---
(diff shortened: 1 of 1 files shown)


mjura (dismissed):

furkatgofurov7 (approved):

anmazzotti (approved):

thehejik:
There is some problem with upgrading turtles on hardened rke2.

On attempt to upgrade to very same version (turtles built from this PR) the upgrade process fails on:
image

It doesn't happen on non-hardened cluster.

I'm using this psa.yaml configuration https://ranchermanager.docs.rancher.com/reference-guides/rancher-security/psa-restricted-exemptions

thehejik:
> @thehejik new namespaces need to be added to the list

@alexander-demicev Are you sure? By adding the namespace into `psa.yaml`, we'll just skip the restricted PSA enforcement for that namespace. By doing so this PR doesn't make any sense IMO.

thehejik (approved):
With the latest commit all three scenarios - install, upgrade and uninstall are working as expected on hardened RKE2 cluster even without the namespace in psa.yaml. I could also provision a downstream v2prov rke2 cluster on vmware.

PR #1912, opened 2025-11-24
fix: Prefix image repository with image registry for provider images

### LINKED ISSUE #1904: Levelage systemDefaultRegistry setting when creating clusterctl config
We need to leverage `systemDefaultRegistry` setting when creating clusterctl config, if this setting is specified Turtles has to use it's value for substituting image registries, if the setting is empty use current defaults.

yiannistri:
This PR adds a new feature flag (which is enabled by default) that controls whether Turtles will use the Rancher default registry when pulling provider images.

Fixes #1904 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/test_chart.yaml
+++ b/.github/workflows/test_chart.yaml
@@ -83,6 +83,12 @@ jobs:
           kubectl --namespace cattle-system wait --for=create deployments/rancher-webhook --timeout=300s
           kubectl --namespace cattle-system wait --for=condition=Available deployment/rancher-webhook --timeout=300s
 
+      - name: Set default registry
+        run: |
+          kubectl patch setting.management.cattle.io -n cattle-system system-default-registry \
+            --type merge \
+            --patch '{"value":"registry.suse.com"}'
+
       - name: Run chart-testing (install)
         run: helm install rancher-turtles out/charts/rancher-turtles/ -n rancher-turtles-system --set namespace=rancher-turtles-system --create-namespace --wait --debug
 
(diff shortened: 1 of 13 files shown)


charts/rancher-turtles/templates/core-provider-configmap.yaml
```diff
@@ -15142,7 +15142,7 @@ data:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.uid
-            image: registry.rancher.com/rancher/cluster-api-controller:v1.10.6
+            image: {{ template "system_default_registry" . }}rancher/cluster-api-controller:v1.10.6
```
alexander-demicev: This might be overridden by the `clusterctlconfig` CR

yiannistri: You are right, we need a different approach for this. I will rework this PR.

api/v1alpha1/clusterctl_config_types.go
```diff
@@ -36,6 +36,10 @@ type ClusterctlConfigSpec struct {
 	// Provider overrides
 	// +optional
 	Providers ProviderList `json:"providers,omitempty"`
+
+	// UseRancherDefaultRegistry indicates whether to use the Rancher default registry for pulling images.
+	// +optional
+	UseRancherDefaultRegistry bool `json:"useRancherDefaultRegistry,omitempty"`
```
alexander-demicev: Should `systemDefaultRegistry` be used by default always? We could even skip including it in the spec and just read the setting directly in the controller code. From the setting description:

```
Private registry to be used for all Rancher System Container Images. If no value is specified, the default registry for the container runtime is used. For Docker and containerd, the default is `docker.io`.
```

What I think we can do is:
- If the setting value is empty, let Turtles use whatever registry is currently specified in the clusterctl config for community/prime provider images.
- If the setting value is not empty, Turtles should use it to substitute the image name.

thehejik:
> @thehejik regarding capz, there will be 2 pods running in the namespace, one is the azureserviceoperator, but there should also be the cluster-api-azure-controller pod running

Right, I overlooked the second one, table above updated - it is present on stgregistry, thanks

thehejik:
caapf image override has been fixed in commit e0e2992 but the image is missing (report above updated):
```
caapf		stgregistry.suse.com/rancher/cluster-api-addon-provider-fleet:v0.12.0		missing
```

anmazzotti (approved):

PR #1951, opened 2025-12-17
docs: Add step for releasing providers chart

### LINKED ISSUE #1718: Setup release process for Turtles Providers chart
Turtles Providers chart was added recently and integrated into our CI but there is no release process. Short summary of what needs to be done:
- Add new release workflow, the chart will have a separate release cycle, it doesn't depend on Turtles tag/releases. The chart should be published as OCI artifact in prime registry, that would require an EIO ticket.
- Document the release process for this chart in `docs/release.md`

yiannistri:

Add providers chart release step.

Fixes #1718 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/docs/release-v2.md
+++ b/docs/release-v2.md
@@ -115,7 +115,7 @@ The providers chart should also be released whenever there is a new version of T
 1. Ensure that the providers chart has the correct [version/appVersion](https://github.com/rancher/turtles/blob/4a30465fb7c5503a2963bab1b041939d6e08a323/charts/rancher-turtles-providers/Chart.yaml#L6-L7) set for the upcoming release. The version should match the Turtles version.
 
 2. Invoke the 'Manual Trigger for Auto Bump' CI workflow from the Prime charts repository and pass the following parameters:
-- Use workflow from: Branch: dev-v2.13
+- Use workflow from: Branch: dev-v2.13 (replace `v2.13` with the desired Rancher minor version)
 - Chart name: rancher-turtles-providers
 - Version override: auto
 


docs/release-v2.md
```diff
@@ -107,6 +108,19 @@ This part of the release is also automated via a GitHub Actions workflow, that n
...
 Once this workflow has finished, a new PR should have been created in the `rancher/rancher` repository that updates the selected branch with the new Turtles version. You need to review and merge this PR. When this PR gets merged, you will have completed the process of releasing a new version of Turtles and including it in an upcoming version of Rancher.
 
+### Release providers chart
+
+The providers chart should also be released whenever there is a new version of Turtles. To release the providers chart:
+
+1. Ensure that the providers chart has the correct [version/appVersion](https://github.com/rancher/turtles/blob/4a30465fb7c5503a2963bab1b041939d6e08a323/charts/rancher-turtles-providers/Chart.yaml#L6-L7) set for the upcoming release. The version should match the Turtles version.
+
+2. Invoke the 'Manual Trigger for Auto Bump' CI workflow from the Prime charts repository and pass the following parameters:
+- Use workflow from: Branch: dev-v2.13
```
salasberryfin:
```suggestion
- Use workflow from: Branch: dev-v2.13 (replace `v2.13` with the desired Rancher minor version)
```

salasberryfin (dismissed):
Thanks @yiannistri. Nit-picking a suggestion but feel free to ignore it, it is not a blocker.

docs/release-v2.md
```diff
@@ -107,6 +108,19 @@ This part of the release is also automated via a GitHub Actions workflow, that n
 
 Once this workflow has finished, a new PR should have been created in the `rancher/rancher` repository that updates the selected branch with the new Turtles version. You need to review and merge this PR. When this PR gets merged, you will have completed the process of releasing a new version of Turtles and including it in an upcoming version of Rancher.
 
+### Release providers chart
+
+The providers chart should also be released whenever there is a new version of Turtles. To release the providers chart:
```
anmazzotti: Should the chart be released on a new Turtles version?
I recall the decision of having hardcoded provider versions was to explicitly avoid this scenario, and release the provider chart only when providers are added/removed I think. I would still miss *when* to make a new release and which lifecycle should it follow.

salasberryfin: We decided to have matching Turtles and Providers chart versions for simplicity. That's why it is expected to be released for each Turtles release.

anmazzotti: Does that mean we can simply configure the provider versions in the providers chart then and have no hardcoded configs and separate builds?

salasberryfin: I'm not sure we'd be able to do that. We still need to differentiate core CAPI, which is embedded in Turtles and cannot be pinned in the providers chart and that means we need to stick with creating a separate image anyway, don't we?

furkatgofurov7 (approved):

PR #1953, opened 2025-12-19
Switch from Turtles to embedded CAPI

### LINKED ISSUE #1827: Switch from Turtles to embedded CAPI
# Description

As a user on `rancher:v2.13.x`, I'd like to enable `embedded-cluster-api` back and disable Turtles.

## Definition of Done

A new test validates that, after installing `rancher:v2.13.x`, Turtles becomes the only source of truth for CAPI CRDs and controllers. Eventually, this configuration is changed and `embedded-cluster-api` feature is enabled. Existing clusters and provisioning should be unaffected.

**Ideally** we should also check that users can go back Turtles once again.

valaparthvi:
kind/enhancement
area/testing

This PR adds a new suite to test Turtles to embedded CAPI switch and back.

>A new test validates that, after installing rancher:v2.13.x, Turtles becomes the only source of truth for CAPI CRDs and controllers. Eventually, this configuration is changed and embedded-cluster-api feature is enabled. Existing clusters and provisioning should be unaffected.
>
>Ideally we should also check that users can go back Turtles once again.

1. We begin with Rancher 2.13 and check if turtles is in charge, checking turtles helm app and `cluster-api-controller-manager` deployment, and provisioning a cluster.

2. Then turtles is disabled and embedded-cluster-api is enabled. We now check if turtles helm app and `cattle-turtles-system` namespace is deleted, ensuring no resources related to it are present. We also check if embedded-capi has been installed in `cattle-provisioning-capi-system` namespace. We also ensure the cluster is unaffected.

3. Then turtles is enabled again, disabling embedded-cluster-api. We now check everything we did in step 1, including the deletion of `cattle-provisioing-capi-system` namespace and ensure the cluster is unaffected.

Fixes #1827

_Please suggest if any additional checks can be added. The issue description mentions checking CAPI CRDs, but I am not sure how to do that._

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests
- [x] adds or updates e2e tests - https://github.com/rancher/turtles/actions/runs/20366987143

### DIFF
--- a/test/e2e/suites/turtles-switch/turtles_switch_test.go
+++ b/test/e2e/suites/turtles-switch/turtles_switch_test.go
@@ -1,0 +1,255 @@
+//go:build e2e
+// +build e2e
+
+/*
+Copyright © 2023 - 2025 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package turtles_switch
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/drone/envsubst/v2"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/rancher/turtles/test/e2e"
+	"github.com/rancher/turtles/test/e2e/specs"
+	turtlesframework "github.com/rancher/turtles/test/framework"
+
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	capiframework "sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/controller-runtime/pkg/envtest/komega"
+)
+
+var _ = Describe("Switch from Turtles to embedded CAPI", Ordered, Label(e2e.ShortTestLabel), func() {
+	var turtlesHelmApp, capiHelmApp turtlesframework.GetHelmAppInput
+	const (
+		oldCAPINamespace = "cattle-provisioning-capi-system"
+		turtlesNamespace = e2e.NewRancherTurtlesNamespace
+
+		clusterName = "cluster-docker-rke2-switch"
+	)
+	BeforeAll(func() {
+		komega.SetClient(bootstrapClusterProxy.GetClient())
+		komega.SetContext(ctx)
+		e2eConfig = e2e.LoadE2EConfig()
+		turtlesHelmApp = turtlesframework.GetHelmAppInput{
+			GetLister:        bootstrapClusterProxy.GetClient(),
+			HelmAppName:      "rancher-turtles",
+			HelmAppNamespace: turtlesNamespace,
+		}
+
+		capiHelmApp = turtlesframework.GetHelmAppInput{
+			GetLister:        bootstrapClusterProxy.GetClient(),
+			HelmAppName:      "rancher-provisioning-capi",
+			HelmAppNamespace: oldCAPINamespace,
+		}
+	})
+
+	It("Should use Turtles as the source of truth", func() {
+		By("Verifying turtles helm app is installed", func() {
+			app, err := turtlesframework.GetHelmApp(ctx, turtlesHelmApp)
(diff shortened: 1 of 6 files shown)


test/e2e/suites/turtles-switch/turtles_switch_test.go
```diff
@@ -0,0 +1,232 @@
...
+	turtlesframework "github.com/rancher/turtles/test/framework"
+
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	capiframework "sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/controller-runtime/pkg/envtest/komega"
+)
+
+var _ = Describe("Switch from Turtles to embedded CAPI", Ordered, Label(e2e.ShortTestLabel), func() {
```
salasberryfin: I'd vote in favor of running this on `FullTestLabel` only. I think it covers a very specific user scenario which is probably not worth testing in basic CI checks and increasing execution time considerably.

test/e2e/data/test-switch/enable-embedded-cluster-api-feature.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: turtles
+spec:
+  value: false
+---
+apiVersion: management.cattle.io/v3
+kind: Feature
+metadata:
+  name: embedded-cluster-api
+spec:
+  value: true
```
salasberryfin: I'm thinking we may be able to configure this with a single file and use `envsubst.Eval()` to modify the values. What do you think? This way we can simplify it and also make it easily reusable.

My idea is that we have a single yaml, i.e. `turtles-embedded-capi-features.yaml` like:
```yaml
apiVersion: management.cattle.io/v3
kind: Feature
metadata:
  name: turtles
spec:
  value: ${ENABLE_TURTLES_FEATURE}
---
apiVersion: management.cattle.io/v3
kind: Feature
metadata:
  name: embedded-cluster-api
spec:
  value: ${ENABLE_EMBEDDED_CAPI_FEATURE}
```
Then when we apply this file we can use `envsubst.Eval()` to substitue these variables. We do something similar [here](https://github.com/rancher/turtles/blob/67e8412fbd90f1c4975a90bd24faf144278e3981/test/testenv/gitea.go#L228):
```go
ingress, err = envsubst.Eval(string(input.CustomIngressConfig), func(s string) string {
    if s == "GITEA_INGRESS_CLASS_NAME" {
		return "nginx"
	}
	return os.Getenv(s)
})
```

test/framework/const.go
```diff
@@ -26,5 +26,5 @@ const (
 	// MagicDNS is the dns name to use in isolated mode
 	MagicDNS = "sslip.io"
 	// DefaulRancherTurtlesNamespace is the name of the default namespace for Rancher Turtles.
-	DefaultRancherTurtlesNamespace = "rancher-turtles-system"
+	DefaultRancherTurtlesNamespace = "cattle-turtles-system"
```
salasberryfin: This is a good catch. In fact I think this constant is not even used. Looks like the values in `./test/e2e/const.go` are used instead https://github.com/rancher/turtles/blob/main/test/e2e/const.go#L150-L150. I think we can just delete this one?

salasberryfin (commented):
Thanks @valaparthvi.

Overall looks good to me. Just added some suggestions and nit-picks. Let me know if you think they don't make sense.

valaparthvi:
Latest run: https://github.com/rancher/turtles/actions/runs/20500606120

salasberryfin (dismissed):
Thanks @valaparthvi! LGTM. Perhaps we can also do a final check and create a branch to run the long E2E manually from it.

valaparthvi:
Long run: https://github.com/rancher/turtles/actions/runs/20519170972

test/e2e/suites/turtles-switch/turtles_switch_test.go
```diff
@@ -0,0 +1,255 @@
...
+
+	specs.CreateUsingGitOpsSpec(ctx, func() specs.CreateUsingGitOpsSpecInput {
+		By("Provisioning workload cluster")
+
+		const topologyNamespace = "switch-docker-kubeadm"
+
+		return specs.CreateUsingGitOpsSpecInput{
+			E2EConfig:                      e2eConfig,
+			BootstrapClusterProxy:          bootstrapClusterProxy,
+			ClusterTemplate:                e2e.CAPIDockerKubeadmTopology,
```
anmazzotti: any reason this is testing kubeadm and not rke2 cluster? I'd rather have the latter covered, but not a blocker for the PR

anmazzotti (dismissed):
Great work. Thank you!

test/e2e/suites/turtles-switch/turtles_switch_test.go
valaparthvi: No particular reason, I can surely do RKE2.

PR #1965, opened 2025-12-30
test: core capi bump on turtles upgrade

### LINKED ISSUE #1819: Test CAPI upgrade when installing from `ConfigMap`
# Description

Core CAPI [is now installed](https://github.com/rancher/turtles/pull/1805) from embedded `ConfigMap`, and there's no remote fetch unless the user explicitly configures the chart with a remote URL. Turtles creates a single `ConfigMap` resource that contains the manifest and a pre-configured `matchLabels` [selector](https://github.com/rancher/turtles/blob/59be443575cbef5eef99582adfc3fc5ed79dad02/charts/rancher-turtles/values.yaml#L95-L100) is used to filter the correct `ConfigMap`. ~However, as a user I may want to create multiple `ConfigMap` with different versions of core CAPI and we need to verify that this configuration does not break the installation process.~

## Definition of Done

At least two new tests should be added that cover this setting:

1. Going from Turtles `vx.x.1` with core CAPI `vx.x.1` -> to Turtles `vx.x.2` with core CAPI `vx.x.2`: during upgrade, the controller should automatically move from the older version of core CAPI to the new one. The `ConfigMap` should contain the manifest with the most recent version of CAPI and the `selector` should match the correct resource. ~If two `ConfigMap` that satisfy the `matchLabels` exist, it should pick the latest version.~ Only one `ConfigMap` can exist because if two resources satisfy the `matchLabels`, the controller will just fail to know which is the correct one.
1. ~While on Turtles `vx.x.1`, if user creates a new `ConfigMap` that matches the `selector`, it should not break the installation of core CAPI.~

salasberryfin:
kind/enhancement
area/testing

This PR adds coverage for testing bumping the version of core CAPI when upgrading Turtles. The scenario is the following:
- **From**: Turtles built from current source with pinned version of core CAPI. Currently `v1.10.6`.
- **To**: custom Turtles built from source with a newer version of core CAPI. Currently `v1.10.7`.

Turtles is upgraded as a system chart by upgrading Rancher as whole and using a new version of Turtles chart which points to a new image. The chart embeds the manifest for the updated version of core CAPI (for which the fetch script is used) and the image is built with a clusterctl configuration that points to the new version of core CAPI and to the upstream registry, as there will be no forked versions of a non-pinned version of the controller.

Testing this requires having a custom procedure for building this alternate image and chart:
- A new test clusterctl configuration yaml file is added to `./test/e2e/data`. This is parsed and used for building the custom image.
- The core provider's manifest is embedded in the chart before building it and removed afterwards.

This new test is added to the existing `chart-upgrade` suite as it aligns with its overall scope.

#### Long E2E test run: https://github.com/rancher/turtles/actions/runs/20662890136

Fixes #1819 

While working on these changes I noticed that there were steps in the E2E process that were duplicated in existing GitHub workflows:
- Long E2E tests first run the image build and publish to Docker Hub, and then run the E2E via `make test-e2e`. However, `make test-e2e` itself calls the image building action.
- Charts were also built twice because of this inconsistency.

The PR tries to address these issues but further optimization (and overall cleanup) should be applied in our current `Makefile` and automations.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/test/testenv/eks.go
+++ b/test/testenv/eks.go
@@ -62,6 +62,7 @@ func CreateEKSBootstrapClusterAndValidateImages(ctx context.Context, input Creat
 		turtlesframework.RunCommand(ctx, turtlesframework.RunCommandInput{
 			Command: "docker",
 			Args: []string{
+				"manifest",
 				"inspect",
 				image.Name,
 			},


alexander-demicev (dismissed):

salasberryfin:
This should be ready to be merged once the long E2E is confirmed to pass after the most recent changes: https://github.com/rancher/turtles/actions/runs/20657367190

salasberryfin:
E2E long are now passing: https://github.com/rancher/turtles/actions/runs/20662890136

PR #1967, opened 2025-12-31
ci: prevent flaky imported annotation removal

anmazzotti:
This PR fixes a flaky error that occasionally pops up during execution.
There are times where the Cluster is updated concurrently and a conflict happens:

```
  [FAILED] Failed to remove 'imported' annotation from CAPI cluster
  Unexpected error:
      <*errors.StatusError | 0xc000281220>: 
      Operation cannot be fulfilled on clusters.cluster.x-k8s.io "kubeadm-test-lxi6hc": the object has been modified; please apply your changes to the latest version and try again
      {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "Operation cannot be fulfilled on clusters.cluster.x-k8s.io \"kubeadm-test-lxi6hc\": the object has been modified; please apply your changes to the latest version and try again",
              Reason: "Conflict",
              Details: {
                  Name: "kubeadm-test-lxi6hc",
                  Group: "cluster.x-k8s.io",
                  Kind: "clusters",
                  UID: "",
                  Causes: nil,
                  RetryAfterSeconds: 0,
              },
              Code: 409,
          },
      }
  occurred
  In [It] at: /home/andrea/repos/rancher-turtles/test/e2e/specs/import_gitops.go:367 @ 12/31/25 10:52:19.915

```

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/test/e2e/specs/import_gitops.go
+++ b/test/e2e/specs/import_gitops.go
@@ -359,12 +359,13 @@ func CreateUsingGitOpsSpec(ctx context.Context, inputGetter func() CreateUsingGi
 			Eventually(komega.Get(rancherCluster), deleteClusterWait...).Should(MatchError(ContainSubstring("not found")), "Rancher cluster should be unimported (deleted)")
 
 			By("Removing 'imported' annotation from CAPI cluster")
-			Eventually(komega.Get(capiCluster), input.E2EConfig.GetIntervals(input.BootstrapClusterProxy.GetName(), "wait-rancher")...).Should(Succeed())
-			annotations := capiCluster.GetAnnotations()
-			delete(annotations, "imported")
-			capiCluster.SetAnnotations(annotations)
-			err = input.BootstrapClusterProxy.GetClient().Update(ctx, capiCluster)
-			Expect(err).NotTo(HaveOccurred(), "Failed to remove 'imported' annotation from CAPI cluster")
+			Eventually(func() error {
+				Eventually(komega.Get(capiCluster), input.E2EConfig.GetIntervals(input.BootstrapClusterProxy.GetName(), "wait-rancher")...).Should(Succeed())
+				annotations := capiCluster.GetAnnotations()
+				delete(annotations, "imported")
+				capiCluster.SetAnnotations(annotations)
+				return input.BootstrapClusterProxy.GetClient().Update(ctx, capiCluster)
+			}).ShouldNot(HaveOccurred(), "Failed to remove 'imported' annotation from CAPI cluster")
 
 			By("Validating annotation is removed from CAPI cluster")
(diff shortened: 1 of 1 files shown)


test/e2e/specs/import_gitops.go
```diff
@@ -359,12 +359,13 @@ func CreateUsingGitOpsSpec(ctx context.Context, inputGetter func() CreateUsingGi
...
-			capiCluster.SetAnnotations(annotations)
-			err = input.BootstrapClusterProxy.GetClient().Update(ctx, capiCluster)
-			Expect(err).NotTo(HaveOccurred(), "Failed to remove 'imported' annotation from CAPI cluster")
+			Eventually(func() error {
+				Eventually(komega.Get(capiCluster), input.E2EConfig.GetIntervals(input.BootstrapClusterProxy.GetName(), "wait-rancher")...).Should(Succeed())
+				annotations := capiCluster.GetAnnotations()
+				delete(annotations, "imported")
+				capiCluster.SetAnnotations(annotations)
+				return input.BootstrapClusterProxy.GetClient().Update(ctx, capiCluster)
+			}).ShouldNot(HaveOccurred(), "Failed to remove 'imported' annotation from CAPI cluster")
```
salasberryfin:
```suggestion
			Eventually(komega.Get(capiCluster), input.E2EConfig.GetIntervals(input.BootstrapClusterProxy.GetName(), "wait-rancher")...).Should(Succeed())
			patch := []byte(`[{"op": "remove", "path": "/metadata/annotations/imported"}]`)
			Eventually(func() error {
				return input.BootstrapClusterProxy.GetClient().Patch(ctx, capiCluster, client.RawPatch(types.JSONPatchType, patch))
			}).ShouldNot(HaveOccurred(), "Failed to remove 'imported' annotation from CAPI cluster")
```
What if we replace the `Update` with a JSON `Patch`? I think it should prevent any chance of a race condition here (I haven't tried this code, though, it's just an idea). I was thinking that perhaps this is a bit cleaner than simply retrying whenever there's a collision. Does that make sense?

salasberryfin (commented):
I think using `Patch` instead of `Update` could be a cleaner fix to this. Does this make sense?

test/e2e/specs/import_gitops.go
alexander-demicev: alternatively, the timeout for eventually block can be increased in case update fails

salasberryfin (approved):

PR #2003, opened 2026-01-16
Bump Cluster API to v1.11.4

furkatgofurov7:

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/rancher-turtles-providers/values.yaml
+++ b/charts/rancher-turtles-providers/values.yaml
@@ -326,31 +326,31 @@ providers:
 images:
   addonFleet:
     repository: rancher/cluster-api-addon-provider-fleet
-    tag: v0.12.0
+    tag: v0.13.0
   bootstrapKubeadm:
     repository: rancher/kubeadm-bootstrap-controller
-    tag: v1.10.6
+    tag: v1.11.4
   bootstrapRKE2:
     repository: rancher/cluster-api-provider-rke2-bootstrap
-    tag: v0.21.1
+    tag: v0.22.0
   controlplaneKubeadm:
     repository: rancher/kubeadm-control-plane-controller
-    tag: v1.10.6
+    tag: v1.11.4
   controlplaneRKE2:
     repository: rancher/cluster-api-provider-rke2-controlplane
-    tag: v0.21.1
+    tag: v0.22.0
   infrastructureAWS:
     repository: rancher/cluster-api-aws-controller
-    tag: v2.9.1
+    tag: v2.10.0
   infrastructureAzure:
     repository: rancher/cluster-api-azure-controller
-    tag: v1.21.0
+    tag: v1.22.0
     azureServiceOperator:
       repository: rancher/azureserviceoperator
-      tag: v2.11.0
+      tag: v2.13.0
   infrastructureGCP:
     repository: rancher/cluster-api-gcp-controller
-    tag: v1.10.0  
+    tag: v1.11.0  
   infrastructureVSphere:
     repository: rancher/cluster-api-vsphere-controller
-    tag: v1.13.1
+    tag: v1.14.0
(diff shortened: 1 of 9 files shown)


examples/clusterclasses/aws/eks/clusterclass-eks-example.yaml
```diff
@@ -1,16 +1,16 @@
-apiVersion: cluster.x-k8s.io/v1beta1
+apiVersion: cluster.x-k8s.io/v1beta2
 kind: ClusterClass
 metadata:
   name: aws-eks-example
 spec:
   controlPlane:
     ref:
```
salasberryfin:
```suggestion
    templateRef:
```
All fields that match a `*Template` type are now `templateRef` and I think we'll need to apply this change to all our examples (for each of the affected types, i.e. `controlPlane`, `machineInfrastructure`, `infrastructure`).

.github/scripts/fetch-core-capi.sh
```diff
@@ -35,7 +35,7 @@ kubectl krew install operator/clusterctl-operator
 kubectl operator version
 
 # use CAPI Operator plugin to generate ConfigMap with core CAPI components
-kubectl operator preload --core cluster-api:${CORE_CAPI_NAMESPACE} -u ${CAPI_RELEASE_URL} >>${OUTPUT_DIR}/${OUTPUT_FILE}
```
salasberryfin: We can probably safely remove this as it is the default value of `CAPI_RELEASE_URL`

Makefile
```diff
@@ -37,7 +37,7 @@ GO_CONTAINER_IMAGE ?= docker.io/library/golang:$(GO_VERSION)
 REPO ?= rancher/turtles
 
 CAPI_VERSION ?= $(shell grep "sigs.k8s.io/cluster-api" go.mod | head -1 |awk '{print $$NF}')
-CAPI_VERSION_TEST_BUMP ?= v1.10.7
+CAPI_VERSION_TEST_BUMP ?= v1.11.4
```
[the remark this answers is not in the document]
furkatgofurov7: I just pinned it to latest since we are upgrading to latest ATM :) However, once v1.11.5 is out we can just bump it but for now it is no-op.

.github/scripts/fetch-core-capi.sh
furkatgofurov7: I was facing some issues like "/components.yaml" is not found or similar along those lines and it was trying to fetch the upstream CAPI release instead of rancher-sandbox so I had to hardcode it. However, now it is working fine, confused a bit but at the same time thinking it could have been a temporary local issue.

salasberryfin (commented):
I think we can probably disable any RKE2-related E2E tests since they'll be failing until the provider version is bumped to a v1.11 compatible release (not yet ready 😓).

go.mod
```diff
@@ -10,43 +10,36 @@ require (
...
+	k8s.io/apimachinery v0.34.1
+	k8s.io/client-go v0.34.1
+	k8s.io/component-base v0.34.1
 	k8s.io/klog/v2 v2.130.1
-	k8s.io/utils v0.0.0-20241104100929-3ea5e8cea738
-	sigs.k8s.io/cluster-api v1.10.6
-	sigs.k8s.io/cluster-api-operator v0.23.1
-	sigs.k8s.io/controller-runtime v0.20.4
+	k8s.io/utils v0.0.0-20250604170112-4c0f3b243397
+	sigs.k8s.io/cluster-api v1.11.4
```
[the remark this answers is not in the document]
furkatgofurov7: @salasberryfin we have `CAPI_VERSION_TEST_BUMP` var defined in Makefile that I previously set to v1.11.4 and used the same CAPI version throughout since it was the only option. Now v1.11.5 being available, are you suggesting to bump that var only so that we test chart upgrade v1.11.4->v1.11.5 OR are you suggesting to use v1.11.5 everywhere ?

Issue #2016, opened 2026-01-21
[SURE-11104]


kkaempf:
### Issue description:
An uninstall of the rancher HelmChart leaves many resources and namespaces like cattle-capi-system ,cattle-fleet-local-system and cattle-turtles-system in the cluster. This is more than a minor inconvenience, as if you try to reinstall the Rancher chart. the rancher-turtles chart will fail on install because of the following issue:
~~~
Error: UPGRADE FAILED: failed to create resource: configmaps "core-cluster-api-v1.10.6" is forbidden: unable to create new content in namespace cattle-capi-system because it is being terminated 
~~~
And it will create pods on a loop until the node is full.

If you try to delete the charts manually, and force delete the namespaces, they will be stuck in terminating even if you delete the finalizers. 

After deleting the rancher-turtles chart, you can see that some resources are being kept:

These resources were kept due to the resource policy:
~~~
[CustomResourceDefinition] capiproviders.turtles-capi.cattle.io
[CustomResourceDefinition] clusterctlconfigs.turtles-capi.cattle.io 
~~~

### Business impact:
Not able to reinstall rancher

### Repro steps:

Install Rancher v2.13.0
Uninstall Rancher
Reinstall Rancher

### Workaround:
Is a workaround available and implemented? yes
What is the workaround: Painfully deleting every resource manually.

### Actual behavior:
There are many lingering resources on an uninstall

### Expected behavior:
The uninstall leaves the cluster clean

PR #2027, opened 2026-01-27
Make delete hooks do a full cleanup

alexander-demicev:

This PR fixes an issue where the config map with CAPI resources was not removed after chart deletion and allows deletion of the CAPI Provider and ClusterctlConfig CRDs. Having these resources present causes issues when re-installing the chart. Neither of these resources contains any data that needs to be backed up if accidentally deleted, all provider CRDs will remain untouched, as they are today.

Fixes https://github.com/rancher/turtles/issues/2016

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/rancher-turtles/templates/post-delete-job.yaml
+++ b/charts/rancher-turtles/templates/post-delete-job.yaml
@@ -30,6 +30,13 @@ rules:
   - deployments
   verbs:
   - delete
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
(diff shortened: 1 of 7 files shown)


alexander-demicev:
@salasberryfin yes, thanks. We also caught a case where providers stay deployed after turtles feature is switched off. Having upstream providers and not having Turtles installed is a unsupported setup.

salasberryfin (approved):

PR #2177, opened 2026-03-03
Use ClusterClass in e2e for GKE provisioning

### LINKED ISSUE #2084: Use ClusterClass for GKE provisioning
# Description

CAPG [v1.11.0](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/releases/tag/v1.11.0) adds experimental support for GKE cluster provisioning using ClusterClass.

Currently, GKE is the only type of cluster provisioning that is not using a topology yet in testing. A new example class should be added, inline with the existing templates, and the cluster manifest in the test data directory should be modified to use this new class.

furkatgofurov7:
A new example clusterclass was added, in line with the existing templates, and the cluster manifest in the test data directory was modified to use this new clusterclass.

GKE automatically upgrades the control plane during provisioning, causing
the Available condition to have a non-empty message about the upgrade in
progress. This is unavoidable GKE behavior and doesn't impact functionality
(the cluster successfully imports into Rancher).
Add SkipClusterAvailableWait flag to CreateUsingGitOpsSpecInput to allow
tests to skip the VerifyClusterAvailable check. Set this flag to true for
the GKE test case since the cluster is functional despite the upgrade state.

Fixes #2084 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/test/e2e/specs/import_gitops.go
+++ b/test/e2e/specs/import_gitops.go
@@ -89,8 +89,12 @@ type CreateUsingGitOpsSpecInput struct {
 	// with stable versions of Turtles, for example during the chart upgrade test.
 	SkipLatestFeatureChecks bool
 
-	// SkipClusterAvailableWait can be used to skip the VerifyClusterAvailable check. This is useful for managed
-	// clusters like GKE that may be in the middle of automatic upgrades during the test.
+	// SkipClusterAvailableWait can be used to skip the VerifyClusterAvailable check.
+	// CAPI's VerifyClusterAvailable asserts that the Available condition has an empty Message,
+	// but managed clusters like GKE may have a non-empty message during automatic upgrades
+	// (e.g. "TopologyReconciled: Cluster is upgrading to v1.x.x"). Since CAPG's releaseChannel
+	// enum (rapid|regular|stable|extended) does not support disabling auto-upgrades, this skip
+	// is necessary for GKE clusters.
 	SkipClusterAvailableWait bool
 }
 
--- a/test/e2e/suites/import-gitops/import_gitops_test.go
+++ b/test/e2e/suites/import-gitops/import_gitops_test.go
@@ -456,13 +456,13 @@ var _ = Describe("[GCP] [GKE] Create and delete CAPI cluster functionality shoul
 			ClusterTemplate:           e2e.CAPIGCPGKETopology,
 			ClusterName:               "cluster-gke",
 			ControlPlaneMachineCount:  ptr.To(1),
-			WorkerMachineCount:        ptr.To(3),
+			WorkerMachineCount:        ptr.To(3), // GKE regional clusters (us-west1 has 3 zones) require machine pool replicas to be a multiple of the zone count (1 node per zone × 3 zones = 3 replicas minimum).
 			LabelNamespace:            true,
 			RancherServerURL:          hostName,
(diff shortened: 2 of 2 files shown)


test/e2e/suites/import-gitops/import_gitops_test.go
```diff
@@ -440,23 +440,37 @@ var _ = Describe("[GCP] [Kubeadm] Create and delete CAPI cluster functionality s
...
 	specs.CreateUsingGitOpsSpec(ctx, func() specs.CreateUsingGitOpsSpecInput {
 		return specs.CreateUsingGitOpsSpecInput{
 			E2EConfig:                 e2e.LoadE2EConfig(),
 			BootstrapClusterProxy:     bootstrapClusterProxy,
-			ClusterTemplate:           e2e.CAPIGCPGKE,
+			ClusterTemplate:           e2e.CAPIGCPGKETopology,
 			ClusterName:               "cluster-gke",
 			ControlPlaneMachineCount:  ptr.To(1),
-			WorkerMachineCount:        ptr.To(1),
+			WorkerMachineCount:        ptr.To(3),
```
anmazzotti: Any reason this was bumped to 3?

salasberryfin (commented):
Thanks @furkatgofurov7.

~Can I ask you to create a branch upstream to verify full end-to-end tests are passing?~ 
**Update**: sorry @furkatgofurov7. I just noticed there are successful E2E executions from your upstream branch.

furkatgofurov7:
@anmazzotti @salasberryfin folks I have long e2e tests passing with current state of PR: https://github.com/rancher/turtles/actions/runs/22696054809

I can address your comments in the same PR or in a follow-up also

test/e2e/suites/import-gitops/import_gitops_test.go
furkatgofurov7: since we changed it on other tests as well, shall we stick to the same approach as in current PR?

salasberryfin:
Thanks @furkatgofurov7. Sorry I missed that you had already executed E2E successfully and only noticed after my message was already published,

test/e2e/specs/import_gitops.go
```diff
@@ -88,6 +88,10 @@ type CreateUsingGitOpsSpecInput struct {
 	// SkipLatestFeatureChecks can be used to skip tests that have not been released yet and can not be tested
 	// with stable versions of Turtles, for example during the chart upgrade test.
 	SkipLatestFeatureChecks bool
+
+	// SkipClusterAvailableWait can be used to skip the VerifyClusterAvailable check. This is useful for managed
+	// clusters like GKE that may be in the middle of automatic upgrades during the test.
```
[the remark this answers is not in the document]
salasberryfin: `none` is not explicitly exposed in `releaseChannel` but it's the default value if the field is not set:
https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/c8b4e1818e5a783126e23ddefe7edc62f17bd940/cloud/services/container/clusters/reconcile.go#L420

However, not setting a release channel won't prevent the cluster from being updated, it just will manage what and when to update differently: https://docs.cloud.google.com/kubernetes-engine/docs/concepts/release-channels?authuser=2#comparison-table-no-channel.

I think using a single replica could also help with reducing the update time. To be honest, it's very impractical that CAPI's testing framework does not allow to set a custom timeout on this check 😓

PR #2281, opened 2026-04-07
e2e: verify EKS etcd size

### LINKED ISSUE #2106: ci: monitor etcd size in e2e tests
This issue is about implementing a way to monitor etcd size when running tests.

etcd can grow to the point of breaking a cluster when there are conflicts in patching resources, like for example controllers misbehaving

in some instances tests may be silently affected by this problem, as the cluster would survive long enough to complete the test

Even if not fully automated with a test, it would be useful to just dump etcd info at the end of the test, to determine whether the database size was 500MB (probably fine) or 2.5GB (there is a problem).

data can be printed out **and** dumped into artifacts from the management cluster using etcdctl, like:

```bash
ETCDCTL_API=3 etcdctl --endpoints=https://127.0.0.1:2379 --cacert=/etc/kubernetes/pki/etcd/ca.crt --cert=/etc/kubernetes/pki/etcd/peer.crt --key=/etc/kubernetes/pki/etcd/peer.key endpoint status --write-out=table
```

the amount of versions per key would also help determining what is constantly being updated.

anmazzotti:

This PR adds EKS etcd size testing.

Fixes #2106 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


Makefile
```diff
@@ -222,6 +222,7 @@ CACHE_COMMANDS = "--cache-from type=local,src=$(CACHE_DIR) --cache-to type=local
 
 # ETCD Verification
 ETCD_VERIFY_SCRIPT_PATH := $(ROOT_DIR)/scripts/collect-etcd-data.sh
+VERIFY_EKS_ETCD_SCRIPT_PATH := $(ROOT_DIR)/scripts/verify-eks-etcd-size.sh
```
furkatgofurov7, line 225: to align with other already defined var:
```suggestion
EKS_ETCD_VERIFY_SCRIPT_PATH := $(ROOT_DIR)/scripts/verify-eks-etcd-size.sh
```

scripts/verify-eks-etcd-size.sh
```diff
@@ -0,0 +1,34 @@
...
+echo "Calculating etcd size using kubeconfig: $KUBECONFIG_PATH"
+
+size=`kubectl --kubeconfig $KUBECONFIG_PATH get --raw /metrics | grep ^apiserver_storage_size_bytes | awk  '{print $2}' | awk -F"E" 'BEGIN{OFMT="%10.0f"} {print $1 * (10 ^ $2) / 1000000}' | xargs`
+
+echo "Calculated etcd size from metrics is $size MB"
+
+size_limit=500
+if (( size > size_limit )); then
+  printf "Error: ETCD database size exceeding limit of $size_limit MB. Found: $size MB\n" >&2
+  exit 1
```
furkatgofurov7, line 31: Do we always fail the tests if db size >=500MB? could there be a case even though it exceed the limit, tests might still pass and it that case we can just dump the logs and not fail whole test?

anmazzotti: Yes the test has to fail to highlight there's a problem with etcd.
This test is the last one performed in the import gitops spec. Logs are always collected in any case, they are part of AfterEach.

Also note that since all tests for this suite run on the same management cluster, having one failing etcd size it's not really indicative that the problem is with a specific provider. There must be some investigation looking at the artifacts on why this happened. 

We are just taking etcd snapshots of the same management cluster at different moments in time (when a cluster has been fully provisioned and verified)

PR #2296, opened 2026-04-13
check immutable checksum after curl download

salasberryfin:

For every tool downloaded using curl, validate the downloaded artifact's sha256 against an immutable, pre-calculated version of the hash.

This also removes unused `updatecli` actions from the Makefile which were using `curl` but are not used as part of the actual `updatecli` workflow which relies on the project's GitHub action.

Fixes #

Original security issue: https://github.com/rancher/rancher-security/issues/2038

Files that are reported in the issue but are not shipped or not considered relevant:
- `ensure-pr-labels.yaml` and `ensure-issues-labels-desc.yaml`: these just curl the GitHub API to check for (lack of) labels. No file or executable involved.
- `fetch-core-capi-airgapped.yaml`: this is in fact aimed at pulling the latest version of Cluster API (from Rancher's forked version of the repository) and creating a new PR which will need to be manually reviewed and approved before merging. No execution happens after curl and we have full control over the releases of the source repository.
- `scripts/turtles-dev.sh` and `scripts/turtles-quickstart.sh`: quick start scripts for engineers and users. Only for development purposes.
- `scripts/go-install`: this is only called from `Makefile` actions and always with a static pinned version, never `latest`.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/hack/ensure-kubectl.sh
+++ b/hack/ensure-kubectl.sh
@@ -26,7 +26,29 @@ fi
 source "$(dirname "${BASH_SOURCE[0]}")/utils.sh"
 
 GOPATH_BIN="$(go env GOPATH)/bin/"
+
+# kubectl version and immutable checksum for validation
 MINIMUM_KUBECTL_VERSION=v1.34.1
+# renovate-local: kubectl-linux-amd64=v1.34.1
+KUBECTL_SUM_linux_amd64="7721f265e18709862655affba5343e85e1980639395d5754473dafaadcaa69e3" # immutable sha256 for v1.34.1 linux amd64
+# renovate-local: kubectl-linux-arm64=v1.34.1
+KUBECTL_SUM_linux_arm64="420e6110e3ba7ee5a3927b5af868d18df17aae36b720529ffa4e9e945aa95450" # immutable sha256 for v1.34.1 linux arm64
+# renovate-local: kubectl-darwin-amd64=v1.34.1
+KUBECTL_SUM_darwin_amd64="bb211f2b31f2b3bc60562b44cc1e3b712a16a98e9072968ba255beb04cefcfdf" # immutable sha256 for v1.34.1 darwin amd64
+# renovate-local: kubectl-darwin-arm64=v1.34.1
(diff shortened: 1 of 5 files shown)


hack/ensure-kubectl.sh
```diff
@@ -27,6 +27,8 @@ source "$(dirname "${BASH_SOURCE[0]}")/utils.sh"
 
 GOPATH_BIN="$(go env GOPATH)/bin/"
 MINIMUM_KUBECTL_VERSION=v1.34.1
+MINIMUM_KUBECTL_VERSION_EXPECTED_SHA256=7721f265e18709862655affba5343e85e1980639395d5754473dafaadcaa69e3 # immutable sha256 for v1.34.1
```
yiannistri: This is the right SHA256 value for linux/amd64 but do we need to cater for macos too?

yiannistri (dismissed):
Thanks, all checksums look good. Unsure whether we should also support macos artifacts or whether `hack/ensure-kubectl.sh` is only used by Linux CI machines.

hack/ensure-kubectl.sh
salasberryfin: Yes, I think I will update this to dynamically support arm/amd

hack/ensure-kubectl.sh
anmazzotti, line 1: @salasberryfin I still see in this file `krew` is installed via `curl` and not checked, and `krew` itself is then used to install latest crust-gather and clusterctl-operator plugins.

salasberryfin: Thanks for catching this.

hack/ensure-kubectl.sh
salasberryfin: It should now dynamically validate against linux/darwin arm/amd depending on the machine's os and architecture.

hack/ensure-kubectl.sh
anmazzotti: fyi krew is still installing latest version of the plugins, is that something we also need to pin or not?

salasberryfin: @anmazzotti I'm not sure `krew` supports installing specific versions.

In any case, since this script is mostly used for development and is not technically shipped or used to build any artifacts, I think it can be considered safe enough in its current form.

anmazzotti (approved):

yiannistri (approved):

PR #2300, opened 2026-04-14
ci: use publish-image action for releasing turtles images

mbologna:
kind/security

- Replace the release + merge matrix jobs with a single publish-images job
- Use `rancher/ecm-distro-tools/actions/publish-image` pinned to a specific SHA commit
- Pin `read-vault-secrets` to a specific SHA commit
- Skip staging push for release candidates (tags containing -rc)
- Add `push-image` and `push-prime-image` Makefile targets required by the action
- Add `REPO`, `IMAGE`, `IMAGE_NAME`, `PUSH_BUILD_TAGS` variables to Makefile
- Update tag filter from limited version range to `v*` to match all releases
- Switch from `ubuntu-latest` to `8cpu-linux-x64` runner for better performance

See also: https://github.com/rancher/rancher-security/issues/2087

**Checklist**:

- [X] squashed commits into logical changes

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -3,11 +3,7 @@ name: Turtles release
 on:
   push:
     tags:
-      # This is just a temporary filter while we maintain two release workflows.
-      # TODO: change to "v*" when this becomes the only one.
-      - "v0.2[5-9].*"
-      - "v0.[3-9]*"
-      - "v[1-9]*"
+      - "v*"
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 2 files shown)


furkatgofurov7 (commented):
This is a big refactor of the release GH workflow that might have been broken. How do we make sure the workflow will work after this change? (I expect at least the workflow being run in a fork repo with some of these or at least a subset of changes, if possible, because you can't test everything in a fork, which I can understand)

furkatgofurov7:
This can be closed as it is re-implemented in #2354

PR #2335, opened 2026-04-27
Improve Image Pull Secret Support

HarrisonWAffel:

This PR updates the turtles chart to use the existing `imagePullSecrets` value in `values.yaml` throughout all of the workloads deployed by the chart. It also updates `core-provider.yaml` to pass the configured image pull secrets to the core CAPI provider deployed by turtles. 

This approach works as the turtles CAPIProvider embeds the [upstream ProviderSpec](https://github.com/kubernetes-sigs/cluster-api-operator/blob/main/api/v1alpha2/provider_types.go#L41), which in turn [supports image pull secrets](https://github.com/kubernetes-sigs/cluster-api-operator/blob/main/api/v1alpha2/provider_types.go#L235). This ensures that turtles can be automatically deployed onto a Rancher instance which only has access to an authenticated private registry. 

I was able to test these changes against a local development build of Rancher and confirmed that both turtles and the core CAPI controllers pulled their images successfully. If there are other areas of turtles that also need to be updated with these pull secrets, let me know. 

Relates to: https://github.com/rancher/rancher/issues/54806

**Checklist**:

- [ ] includes documentation

### DIFF
--- a/charts/rancher-turtles-providers/templates/addon-fleet.yaml
+++ b/charts/rancher-turtles-providers/templates/addon-fleet.yaml
@@ -104,6 +104,13 @@ spec:
     name: {{ index .Values "providers" "addonFleet" "configSecret" "name" }}
     namespace: {{ index .Values "providers" "addonFleet" "configSecret" "namespace" }}
 {{- end }}
+{{- if .Values.imagePullSecrets }}
+  deployment:
+    imagePullSecrets:
+{{- range .Values.imagePullSecrets }}
+      - name: {{ . }}
+{{- end }}
+{{- end }}
 {{- if index .Values "providers" "addonFleet" "fetchConfig" }}
   fetchConfig:
     {{- if index .Values "providers" "addonFleet" "fetchConfig" "url" }}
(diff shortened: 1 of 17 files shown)


anmazzotti:
@HarrisonWAffel thank you for the changes. Should this cover also the providers in `charts/rancher-turtles-providers`? We didn't map the imagePullSecrets value there, so it needs to be exposed.

HarrisonWAffel:
@anmazzotti 

Sure, I can expose those there too. But I want to note that, unlike the `rancher-turtles-system`, Rancher will not automatically mirror the global pull secrets into those provider namespaces (`capa-system`, `capz-system`, etc.). This was a limitation called out and agreed to in the RFD, so we might also want to make some doc's changes to call that out?

anmazzotti:
Thanks, I was not aware. 
Just mentioning as user it's something I would expect. Also fine if added later eventually.

anmazzotti (approved):

salasberryfin (approved):
Thanks @HarrisonWAffel

salasberryfin (commented):
Can you amend and sign the commits for CI to allow merging?

HarrisonWAffel:
@salasberryfin should be good now!

PR #2354, opened 2026-05-05
feat: use publish image action in release workflow

furkatgofurov7:
Replaces the manual Docker login + docker buildx build steps in the release job with [rancher/ecm-distro-tools/actions/publish-image](https://github.com/rancher/ecm-distro-tools/tree/master/actions/publish-image) - the standard action used across Rancher projects for publishing community and prime images. The main goal of this PR is to introduce as minimal changes as possible to the existing workflow structure while introducing a new action and keeping the resulting artifacts identical.

Things that got changed:
- replaced QEMU setup, registry logins, and inline docker buildx build steps with two publish-image action calls (community + prime)
- moved Cosign/slsactl installation to after imagetools creation, and  fixing a bug where signing ran per-arch before instead of on the final multi-arch manifest now
- added `PRIME_REGISTRY` vault secret to the release job (required by publish-image for --sign-container-identity when signing per-arch prime images)
- replaced `docker-build-and-push-*` targets with new `push-image` / `push-prime-image` called by the action;
- changed `TARGET_PLATFORMS :=` to `?=` in Makefile so the action can pass per-arch platform values via environment.

Fixes #

Since this is a release workflow it is challenging to test it properly, however, this is what I could do:
- All code tested in my personal fork, where I have 2 commits: 1st is the main changes (what we have in this PR) and the second is what I had to rework to make it run successfully in my fork (not included in this PR): https://github.com/rancher/turtles/compare/main...furkatgofurov7:turtles:main
- verified that the full workflow [run](https://github.com/furkatgofurov7/turtles/actions/runs/25367574900) and [fake release](https://github.com/furkatgofurov7/turtles/releases/tag/v0.99.5-test.) was created fine 
- verified that Multi-arch manifest structure matches lastest release artifacts `rancher/turtles:v0.26.1` (linux/amd64 + linux/arm64 + attestations)
- verified Cosign signature and SLSA provenance attestation on the multi-arch manifest

- [ ] includes documentation
- [ ] adds unit tests

Takes inspiration from: https://github.com/rancher/turtles/pull/2300

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -32,12 +32,6 @@ jobs:
       with:
         fetch-depth: 0
 
-    - name: Setup QEMU
-      uses: docker/setup-qemu-action@ce360397dd3f832beb865e1373c09c0e9f86d70a # v4.0.0
-      with:
-        image: tonistiigi/binfmt:qemu-v8.1.5
-        cache-image: false
-
     - name: Read Vault secrets
       uses: rancher-eio/read-vault-secrets@0da85151ad1f19ed7986c41587e45aac1ace74b6
       with:
@@ -47,37 +41,33 @@ jobs:
           secret/data/github/repo/${{ github.repository }}/rancher-prime-stg-registry/credentials registry | STAGE_REGISTRY ;
           secret/data/github/repo/${{ github.repository }}/rancher-prime-stg-registry/credentials username | STAGE_REGISTRY_USERNAME ;
           secret/data/github/repo/${{ github.repository }}/rancher-prime-stg-registry/credentials password | STAGE_REGISTRY_PASSWORD ;
+          secret/data/github/repo/${{ github.repository }}/rancher-prime-registry/credentials registry | PRIME_REGISTRY ;
 
-    - name: Log into Docker Hub registry
-      uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
+    - name: Push community image to DockerHub
+      uses: rancher/ecm-distro-tools/actions/publish-image@5d6e63896fea85836b2fca6dfb9236bcefe2aaa5
       with:
-        username: ${{ env.DOCKER_USERNAME }}
-        password: ${{ env.DOCKER_PASSWORD }}
+        image: turtles
+        tag: ${{ env.TAG }}-${{ matrix.tag-suffix }}
+        platforms: ${{ matrix.platform }}
+        push-to-prime: false
+        public-repo: rancher
+        public-username: ${{ env.DOCKER_USERNAME }}
+        public-password: ${{ env.DOCKER_PASSWORD }}
+        make-target: push-image
 
-    - name: Log into Staging registry
-      uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
+    - name: Push prime image to staging registry
+      uses: rancher/ecm-distro-tools/actions/publish-image@5d6e63896fea85836b2fca6dfb9236bcefe2aaa5
       with:
-        username: ${{ env.STAGE_REGISTRY_USERNAME }}
-        password: ${{ env.STAGE_REGISTRY_PASSWORD }}
-        registry: ${{ env.STAGE_REGISTRY }}
-
-    - name: Build and push community image
-      shell: bash
-      env:
-        REGISTRY: docker.io
-        ORG: rancher
-      run: |
-        IID_FILE=$(mktemp)
-        make docker-build-and-push-community TAG=${{ env.TAG }}-${{ matrix.tag-suffix }} REGISTRY=${{ env.REGISTRY }} ORG=${{ env.ORG }} IID_FILE=${IID_FILE} TARGET_PLATFORMS=${{ matrix.platform }}
-
-    - name: Build and push prime image
-      shell: bash
-      env:
-        REGISTRY: ${{ env.STAGE_REGISTRY }}
-        ORG: rancher
-      run: |
-        IID_FILE=$(mktemp)
-        make docker-build-and-push-prime TAG=${{ env.TAG }}-${{ matrix.tag-suffix }} REGISTRY=${{ env.REGISTRY }} ORG=${{ env.ORG }} IID_FILE=${IID_FILE} TARGET_PLATFORMS=${{ matrix.platform }}
+        image: turtles
+        tag: ${{ env.TAG }}-${{ matrix.tag-suffix }}
+        platforms: ${{ matrix.platform }}
+        push-to-public: false
+        prime-repo: rancher
+        identity-registry: ${{ env.PRIME_REGISTRY }}
+        prime-registry: ${{ env.STAGE_REGISTRY }}
+        prime-username: ${{ env.STAGE_REGISTRY_USERNAME }}
+        prime-password: ${{ env.STAGE_REGISTRY_PASSWORD }}
+        prime-make-target: push-prime-image
 
   merge:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


mbologna:
Supersedes https://github.com/rancher/turtles/pull/2300

mbologna:
See also https://github.com/rancher/rancher-security/issues/2087

mbologna (dismissed):

.github/workflows/release.yaml
```diff
@@ -47,37 +41,33 @@ jobs:
           secret/data/github/repo/${{ github.repository }}/rancher-prime-stg-registry/credentials registry | STAGE_REGISTRY ;
           secret/data/github/repo/${{ github.repository }}/rancher-prime-stg-registry/credentials username | STAGE_REGISTRY_USERNAME ;
           secret/data/github/repo/${{ github.repository }}/rancher-prime-stg-registry/credentials password | STAGE_REGISTRY_PASSWORD ;
+          secret/data/github/repo/${{ github.repository }}/rancher-prime-registry/credentials registry | PRIME_REGISTRY ;
 
-    - name: Log into Docker Hub registry
-      uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
+    - name: Push community image to DockerHub
+      uses: rancher/ecm-distro-tools/actions/publish-image@5d6e63896fea85836b2fca6dfb9236bcefe2aaa5
```
yiannistri: Not a blocker, but perhaps we could pin to a more recent commit, as this one is from April 10th.

.github/workflows/release.yaml
```diff
@@ -47,37 +41,33 @@ jobs:
...
-      env:
-        REGISTRY: ${{ env.STAGE_REGISTRY }}
-        ORG: rancher
-      run: |
-        IID_FILE=$(mktemp)
-        make docker-build-and-push-prime TAG=${{ env.TAG }}-${{ matrix.tag-suffix }} REGISTRY=${{ env.REGISTRY }} ORG=${{ env.ORG }} IID_FILE=${IID_FILE} TARGET_PLATFORMS=${{ matrix.platform }}
+        image: turtles
+        tag: ${{ env.TAG }}-${{ matrix.tag-suffix }}
+        platforms: ${{ matrix.platform }}
+        push-to-public: false
```
anmazzotti, line 65: a bit confused on why I don't see `push-to-prime: true` here. 
is this implied by `push-to-public: false` or by `public-repo` vs `prime-repo` usage?
can it be set explicitly on both steps? like:

community:
```
push-to-prime: false
push-to-public: true
public-repo: rancher
```

prime:
```
push-to-prime: true
push-to-public: false
prime-repo: rancher
```

Also to avoid changes in default behavior in the future.

.github/workflows/release.yaml
```diff
@@ -154,14 +136,22 @@ jobs:
...
 
+    - name: Install Cosign
+      if: ${{ matrix.image-type == 'prime' }}
+      uses: sigstore/cosign-installer@cad07c2e89fa2edd6e2d7bab4c1aa38e53f76003 # v4.1.1
+
+    - name: Install slsactl
+      if: ${{ matrix.image-type == 'prime' }}
+      uses: rancherlabs/slsactl/actions/install-slsactl@d70524abe2bec09b488ab946abcb4bef9c1331bf # v0.1.31
+
     - name: Sign multi-platform image
```
anmazzotti, line 149: Is this still needed? I think I read something similar in the `rancher/ecm-distro-tools/actions/publish-image`

.github/workflows/release.yaml
```diff
@@ -154,14 +136,22 @@ jobs:
...
       if: ${{ matrix.image-type == 'prime' }}
       run: |
         cosign sign \
           --oidc-provider=github-actions \
           --yes \
-          --sign-container-identity="${{ env.PRIME_REGISTRY }}/rancher/${IMAGE}" \
+          --sign-container-identity="${{ env.PRIME_REGISTRY }}/rancher/turtles" \
           "${MULTI_PLATFORM_IMAGE}"
 
     - name: Attest provenance
```
anmazzotti, line 159: Same as above, this also seem to be part of the new action

Makefile
```diff
@@ -389,17 +389,31 @@ docker-build-and-push: buildx-machine docker-pull-prerequisites ## Run docker-bu
...
-.PHONY: docker-build-and-push-community
-docker-build-and-push-community:
-	$(MAKE) docker-build-and-push TARGET_BUILD=community
-
 docker-list-all:
 	@echo $(CONTROLLER_IMG):${TAG}
 
+# The following targets are called by the publish-image action in the release workflow.
+# See .github/workflows/release.yml for details.
+# The docker-build-and-push* targets above were used by the old release workflow.
```
yiannistri: We can probably get rid of this comment, since we're removing them from the Makefile.

Makefile
```diff
@@ -389,17 +389,31 @@ docker-build-and-push: buildx-machine docker-pull-prerequisites ## Run docker-bu
 			--build-arg go_build_tags=$(TARGET_BUILD) \
 			--build-arg ldflags="$(LDFLAGS)" . -t $(CONTROLLER_IMG):$(TAG)
 
-.PHONY: docker-build-and-push-prime
-docker-build-and-push-prime:
-	$(MAKE) docker-build-and-push TARGET_BUILD=prime
```
anmazzotti, line 394: Can `docker-build-and-push` target and `buildx-machine` be cleaned up now or is there still any usage for them?

yiannistri (dismissed):
Thanks, just a couple of non-blocking comments.

.github/workflows/release.yaml
furkatgofurov7: @anmazzotti, that is a good question. 

`rancher/ecm-distro-tools/actions/publish-image`  action signs the per-arch image using the digest from its own build's IID_FILE (tag = vX.Y.Z-linux-amd64): https://github.com/rancher/ecm-distro-tools/blob/5d6e63896fea85836b2fca6dfb9236bcefe2aaa5/actions/publish-image/action.yaml#L243-L252

While our merge step signs the multi-arch manifest index assembled by imagetools create a different digest that doesn't exist yet when the action runs:

```
IMAGE_DIGEST=$(docker inspect --format='{{index .RepoDigests 0}}' ${URL} | sed 's/.*@//')
MULTI_PLATFORM_IMAGE="${{ env.STAGE_REGISTRY }}/rancher/${IMAGE}@${IMAGE_DIGEST}"
cosign sign ... "${MULTI_PLATFORM_IMAGE}"
```

To let the action fully own sign/attest, we'd need to drop the matrix + imagetools create pattern and build both arches in a single job. 

TL:DR; is with what you are suggesting we will lose per-arch tags (i.e., rancher/turtles:v0.26.1-linux-amd64 / rancher/turtles:v0.26.1-linux-arm64), while as is we keep the same and old behaviour. 

TBH, I was not sure if we actually have consumers of the per-arch tags. If not, we could drop the matrix build, simplify the workflow significantly, and let the action own everything, including sign/attest. I kept them to preserve existing behaviour, but if nobody uses them, it's unnecessary complexity.

.github/workflows/release.yaml
furkatgofurov7: same as above, new action attests per-arch digest, while our merge job attests multi-arch digest.

Makefile
furkatgofurov7: They are used in https://github.com/rancher/turtles/blob/main/.github/workflows/nightly-chart-and-image-publish.yaml#L32-L34 and the latter is a dependency of `docker-build-and-push` and also `docker-build`

.github/workflows/release.yaml
furkatgofurov7: pinning it to https://github.com/rancher/ecm-distro-tools/commit/067de5ae5517b661b4d2f10e22cc5a21aa0ac915 (from May 1st)

.github/workflows/release.yaml
anmazzotti: thank you for the explanation!

anmazzotti (approved):

Makefile
furkatgofurov7: Removed it

furkatgofurov7:
Tested again from fork with second commit included: https://github.com/furkatgofurov7/turtles/actions/runs/25381102185 (🟢)

yiannistri (approved):

mbologna (approved):

salasberryfin (approved):

furkatgofurov7:
Thanks everyone for reviews!
I am merging this now (as early as possible) and will cut an upcoming RC release using it to see it in action and follow up with extra fixes if needed (hopefully not 🤞🏼).

PR #2416, opened 2026-05-19
make release workflow comply with immutability

### LINKED ISSUE #2359: chart-release workflow does not comply with immutable releases
**Problem:**
Currently, in [chart-release workflow](https://github.com/rancher/turtles/blob/main/.github/workflows/chart-release.yml#L61-L88) `cr upload` creates and publishes the GitHub release immediately. Any steps that run after it, specifically the recovery step: `Verify chart asset present`, which does `gh release upload --clobber`  attempts to modify an already published release, violating GitHub's immutable releases requirement.

**Additional note:**
The recovery step was introduced (by me 😢) in #2141 to handle [the case](https://github.com/rancher/turtles/pull/2138) where cr upload succeeds in creating the release but the chart asset ends up missing. Whatever solution is implemented must continue to address that scenario.

**Thoughts on possible fix:**
We could replace the current `cr upload` + `recovery pattern` with a proper draft and then publish flow so that the release is only published after all assets are confirmed present, making both immutable releases compliance and #2138 concerns addressed in one go. Rougly the fix could look like:

- Create the release as a draft first
- Upload the chart asset explicitly to the draft
- Update the helm index
- Publish the release as the very last step

furkatgofurov7:
### Problem
Currently, `cr upload` creates and immediately publishes the GitHub release in a single step. This means the release is public and mutable from the moment it was created.

This causes two problems:

1. **Chart asset missing on re-run** ([example](https://github.com/rancher/turtles/actions/runs/22062100408/job/63745117786)): if `cr upload` failed mid-way (e.g. a 504 from GitHub), re-running the job would skip creating the release (via `--skip-existing`) but also skip uploading the `.tgz` asset, leaving the release with no chart attached.
2. **Immutability violation**: the `Verify chart asset present` recovery step introduced to fix the above used `gh release upload --clobber` to attach the missing asset to an already-published release. Once we enable immutable releases, modifying a published release will no longer be possible, so this recovery step would break.

### What this PR does

Adds `--draft` to `cr upload` so the release is created in draft state. This means:

- The release is **mutable throughout** the entire job: `cr upload`, asset verification, recovery upload, and `cr index` all operate on a draft release
- The **recovery step is safe to keep** as-is because `gh release upload --clobber` on a draft should be perfectly valid
- The release is only made visible to users after a **manual undraft** from the GitHub UI

This directly fixes the immutability violation while continuing to address the original chart-missing scenario we had to fix with the recovery step.

### Path to immutable releases

This PR is **Phase 1** of a gradual migration:

**Phase 1 (this PR):** Draft-first flow, manual undraft. Gives us confidence that the suggested process works reliably without the risk of being stuck with a broken immutable release.

**Phase 2 (follow-up, once Phase 1 is validated over a few releases):** Uncomment the `Publish release` step already included in the workflow. This automates the undraft as the very last step, only after all assets are confirmed present, fully following [GitHub's recommended immutable release workflow](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases)

**Phase 3 (no code change needed, a follow-up once Phase 2 is validated):** Enable immutable releases in the
repository settings under **Settings -> Releases -> "Enable release immutability"**.

I will create a follow-up issue to track Phase 2 and 3 which should not block this work and anyways it will take time to get there.

Fixes: #2359

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/chart-release.yml
+++ b/.github/workflows/chart-release.yml
@@ -58,8 +58,10 @@ jobs:
           rm -rf .cr-index
           mkdir -p .cr-index
 
-      - name: Run chart-releaser upload
-        run: cr upload --skip-existing -c "$(git rev-parse HEAD)" --generate-release-notes --release-name-template "${{ inputs.tag }}" --make-release-latest=false
+      - name: Run chart-releaser upload (draft release)
+        # Phase 1: create release as draft so all assets can be verified before publishing.
+        # The release must be manually undrafted from the GitHub UI - be careful not to mark it as latest.
+        run: cr upload --skip-existing -c "$(git rev-parse HEAD)" --generate-release-notes --release-name-template "${{ inputs.tag }}" --make-release-latest=false --draft
 
       - name: Verify chart asset present (recover if missing)
         run: |
@@ -69,11 +71,6 @@ jobs:
             exit 1
           fi
           ASSET_NAME=$(basename "${CHART_FILE}")
-
-          if ! gh release view "${{ inputs.tag }}" > /dev/null 2>&1; then
-            echo "ERROR: Release '${{ inputs.tag }}' not found."
-            exit 1
-          fi
 
           FOUND=$(gh release view "${{ inputs.tag }}" \
             --json assets \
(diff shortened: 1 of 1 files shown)


anmazzotti (approved):
I was expecting the unnecessary cr tool to be removed.
As I pointed out before, `gh release create` and `gh release upload` are all that we need. External tools should be removed as they will probably be flagged by security as well eventually. `gh` is the recommended one.

I think the documentation should also be updated, no one is really going to read the `be careful not to mark it as latest.` comment in here, not sure what the audience of that comment is.

Non blockers though, can be addressed later.

yiannistri (approved):

PR #2448, opened 2026-05-27
ci: Add new workflow for releasing turtles providers chart

### LINKED ISSUE #2236: Improve `rancher-turtles-providers` version management
# Description

With the current release process, before a new release is published the `version` and `appVersion` in `charts/rancher-turtles-providers/Chart.yaml` must be manually updated to match the coming release.

This configuration allows us to use the existing `rancher/prime-charts` [automation](https://github.com/rancher/prime-charts/actions/workflows/auto-bump-manual-trigger.yaml) for creating new versions of prime-only charts.

However, it poses the problem of forcing us to change the state of the repository after validation on a release candidate is performed, which effectively means that we end up creating a stable release that doesn't correspond to the same version of the code (different commit hash).

For example:
- `rancher-turtles` and `rancher-turtles-providers` `v0.26.0-rc.1` is published for validation.
- Validation is successful and we're now ready to create a stable `v0.26.0`.
- Before we un-RC, we must update `rancher-turtles-providers` `Chart.yaml` to set `v0.26.0`. At this point we add commits on top of the validated revision.
- Without the above change we cannot release a new `rancher-turtles-providers` chart.

This can get worse if, during the time between RC creation and un-RC, new changes are merged to the release branch. In this case the branch has diverged from the original state of the QA-validated RC.

## Definition of Done

We should be able to un-RC a version of Turtles maintaining the same code revision.

yiannistri:

This PR introduces a new release workflow for the providers chart, to bring it inline with the turtles chart release process. The main issue being addressed is that it no longer requires an new commit to set the version number for the providers chart, whenever a release happens. Instead, the workflow runs as part of a release (invoked manually at release time) to create a PR that bumps the providers chart version with the correct version number. 

Prerequisites for this to work:
- permissions granted to create a PR in rancher/prime-charts (currently in progress)
- the ability to point to a commit SHA of a release branch in order to identify the version of the providers chart to be released/bumped (needs input from release team)

Fixes #2236 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/scripts/release-against-prime-charts.sh
+++ b/.github/scripts/release-against-prime-charts.sh
@@ -1,0 +1,160 @@
+#!/usr/bin/env bash
+#
+# Bumps Turtles providers version in a locally checked out rancher/prime-charts repository
+#
+# Usage:
+#   ./release-against-prime-charts.sh <path to prime-charts repo> <path to turtles repo> <prev turtles providers release> <new turtles providers release> [bump major]
+#
+# Example:
+# ./release-against-prime-charts.sh "${GITHUB_WORKSPACE}/prime-charts" "${GITHUB_WORKSPACE}/turtles" "v0.27.0-rc.0" "v0.27.0" false
+
+CHARTS_DIR=$1
+TURTLES_DIR=$2
+PREV_TURTLES_PROVIDERS_VERSION="$3"   # e.g. 0.23.0-rc.0
+NEW_TURTLES_PROVIDERS_VERSION="$4"    # e.g. 0.23.0
+BUMP_MAJOR="${5:-false}"   # default false if not given
+
+usage() {
+    cat <<EOF
+Usage:
+  $0 <path to prime-charts repo> <path to turtles repo> <prev turtles providers release> <new turtles providers release> [bump_major]
+
+Arguments:
+  <path to prime-charts repo>       Path to locally checked out prime-charts repo
+  <path to turtles repo>            Path to locally checked out turtles repo
+                                    (checked out at the ref being released; its HEAD commit is recorded)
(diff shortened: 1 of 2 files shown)


.github/scripts/release-against-prime-charts.sh
```diff
@@ -0,0 +1,160 @@
...
+    echo "Bumping chart patch: $PREV_CHART_VERSION to $(bump_patch "$PREV_CHART_VERSION")"
+    NEW_CHART_VERSION=$(bump_patch "$PREV_CHART_VERSION")
+    COMMIT_MSG="Bump rancher-turtles-providers to $NEW_TURTLES_PROVIDERS_VERSION (chart version patch bump)"
+else
+    echo "Keeping chart version unchanged: $PREV_CHART_VERSION"
+    NEW_CHART_VERSION=$PREV_CHART_VERSION
+    COMMIT_MSG="Bump rancher-turtles-providers to $NEW_TURTLES_PROVIDERS_VERSION (no chart version bump)"
+fi
+
+# Determine the commit SHA for the release branch
```
anmazzotti, line 135: Can't we pass the `${{ github.ref_name }}` value instead? Job is always triggered on tags

anmazzotti: Sorry, I confused this with the NEW_CHART_VERSION. Nevermind

yiannistri:
The current tooling for releasing charts does not allow us to use a commit SHA to indicate which version of the rancher-turtles-providers chart should be released **without** also specifying the version in Chart.yaml. And we cannot create a GH release of the providers chart in any other repo other than rancher/turtles because GH releases are linked to tags.

PR #2462, opened 2026-05-31
fix: return error from DefaultSynchronizer.Get instead of silently dropping it

immanuwell:
What this PR does / why we need it:

`DefaultSynchronizer.Get` logs non-NotFound errors but always returns `nil` - errors get silently dropped. The sync pipeline continues with stale destination state, no requeue triggered.

`interface.go` collects `syncer.Get(ctx)` errors expecting them to propagate - they never do.

Fix is one line: return the error. NotFound still returns nil (intentional, unchanged).

Reproduces on any transient non-404 API error: rate limiting (429), timeouts, auth errors (403), server errors (500).

Which issue(s) this PR fixes:
Fixes #

Special notes for your reviewer:

Checklist:
- [ ] includes documentation
- [ ] adds unit tests


anmazzotti (approved):

furkatgofurov7 (approved):
Thanks for PR!

PR #2463, opened 2026-06-02
feat: translate Rancher AWS Cloud Credentials to `AWSClusterStaticIdentity`

salasberryfin:

- :green_circle:  Successfull long end-to-end test execution: https://github.com/rancher/turtles/actions/runs/27346906885

The goal of this PR is to introduce logic that translates Rancher AWS Cloud Credentials into CAPA `AWSClusterStaticIdentity` (and underlying `Secret`) for enabling CAPI Plug and Play for AWS.

Translation is (for now) implemented for AWS and `AWSClusterStaticIdentity` only. Users can continue to use regular CAPI authentication mechanisms normally. This should only affect users of Rancher v2prov and Rancher Cloud Credentials.

For now, **this is behind a feature gate because it cannot be enabled until the second part of this feature (validation webhook) is implemented adding a security layer** that will prevent users that do not own a given Rancher Cloud Credential from using the translated AWS identity.

Some implementation details:
- A new controller will watch Rancher Cloud Credential `Secrets` in the `cattle-global-data` namespace. It will only trigger on `Secrets` that represent AWS credentials (only the AWS Translator exists).
- Even if there are Rancher AWS Cloud Credentials, the controller will perform a validation to check if the CAPA `CAPIProvider` object is installed and `Status.Phase=Ready`. Otherwise, translation won't happen.
- If the above conditions are met, a new `AWSClusterStaticIdentity` (global resource) and a `Secret` (with the corresponding AWS keys) will be created in `capa-system`.
- The original Rancher Cloud Credential `Secret` is annotated with a reference to the new `AWSClusterStaticIdentity`. This also serves as a way to identify which credentials have already been translated.
- A new finalizer is set on the Rancher Cloud Credential secret to cascade down the deletion of translated objects, namely `AWSClusterStaticIdentity` and `Secret`, before entirely removing the Cloud Credential.
- By default, the translated objects will use the same name as the Rancher Cloud Credential `Secret` object, which is automatically generated `cc-*`.
- The existing E2E test that provisions a AWS RKE2 cluster is updated to use a translated Rancher Cloud Credential. A new `Secret` that represents a Cloud Credential is created and Turtles is deployed with the translation feature gate enabled. The `AWSClusterStaticIdentity` is now embedded in the `AWSClusterTemplate` so that it uses it to authenticate against AWS. All other AWS-related tests continue to work with CAPI-specific authentication mechanisms (remain unchanged).
- A new set of Go tests is added to validate the reconciling logic: including standard translation, not erroring on missing CAPA `CAPIProvider`, translating older Cloud Credentials when CAPA is installed, not translating credentials other than AWS, etc.
- To prevent object name collisions, a reference (annotation) to the original Rancher Cloud Credential `Secret` UID is added to the translated resources (in AWS `Secret` and `AWSClusterStaticIdentity`. This allows the controller to **only** create/update these resources when there are no name collisions. If there's a `Secret` or `AWSClusterStaticIdentity` that exists with the same name as the generated credential, translation will fail and no resources will be created. The cleanup logic is also updated to **only** remove objects which are verified to be owned by a Rancher Cloud Credential that was translated.
- Translated `AWSClusterStaticIdentity` will only allow provisioning clusters to `fleet-default` namespace.

Fixes #

We decided to have a first implementation that limits user configuration. CAPA `CAPIProvider` is expected to be installed in the default system namespace `capa-system` and with the default name `capa`. Both of these values are defaults in the `rancher-turtles-providers` Helm chart. If there's interest for more granular user configurations, it was agreed that we would study adding it in the future.

**Controller logic is agnostic and should be easily scalable to add more providers by implementing the `CredentialTranslator` interface.**

Additionally, Kubebuilder validation on `providerSpec.configSecret` prevents from populating either `configSecret` or `configSecret.namespace`. However this namespace is populated with a default (the provider's ns) and any update to the object after the default is set will fail because it is caught by the CEL rule.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -53,6 +53,7 @@ rules:
   - infrastructure.cluster.x-k8s.io
   resources:
   - '*'
+  - awsclusterstaticidentities
   verbs:
   - create
   - delete
--- a/examples/clusterclasses/aws/rke2/clusterclass-ec2-rke2-example.yaml
+++ b/examples/clusterclasses/aws/rke2/clusterclass-ec2-rke2-example.yaml
@@ -73,7 +73,6 @@ spec:
         openAPIV3Schema:
           description: The AWSClusterStaticIdentity resource name referencing the credentials to create the Cluster.
           type: string
-          default: cluster-identity
   patches:
     - name: awsClusterTemplate
       definitions:
@@ -99,7 +98,7 @@ spec:
             matchResources:
               infrastructureCluster: true
           jsonPatches:
-            - op: add
+            - op: replace
               path: /spec/template/spec/identityRef/name
               valueFrom:
                 variable: awsClusterIdentityName
@@ -167,7 +166,7 @@ spec:
     spec:
       identityRef:
         kind: AWSClusterStaticIdentity
-        name: cluster-identity
+        name: default
       bastion:
         enabled: false
       controlPlaneLoadBalancer:
--- a/internal/controllers/credentialtranslation_controller.go
+++ b/internal/controllers/credentialtranslation_controller.go
@@ -1,0 +1,369 @@
+/*
+Copyright © 2023 - 2026 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	turtlesv1 "github.com/rancher/turtles/api/v1alpha1"
+	turtlesannotations "github.com/rancher/turtles/util/annotations"
+)
+
+const (
+	// awsDriverAnnotationValue is the value of the driver annotation that identifies AWS credentials.
+	awsDriverAnnotationValue = "aws"
+
+	// awsRancherCCFinalizer is the finalizer added to Rancher Cloud Credentials to ensure
+	// cleanup of the derived AWSClusterStaticIdentity when the credential is deleted.
+	awsRancherCCFinalizer = "cloudcredential.cattle.io/aws-identity-finalizer"
+
+	// capaProviderSpecName is CAPA's CAPIProvider `spec.name`.
+	capaProviderSpecName = "aws"
+
+	// capaProviderNamespace is the default namespace where CAPA is installed and where the credentials secret will be created.
+	// defaults to `capa-system`.
+	capaProviderNamespace = "capa-system"
+
+	// rancherAWSAccessKeyField is the field name in Rancher Cloud Credential secrets for the AWS access key ID.
+	rancherAWSAccessKeyField = "amazonec2credentialConfig-accessKey"
+
+	// rancherAWSSecretKeyField is the field name in Rancher Cloud Credential secrets for the AWS secret key.
+	//nolint:gosec  // This is not a hardcoded secret, just a key name.
+	rancherAWSSecretKeyField = "amazonec2credentialConfig-secretKey"
+)
+
+// RancherCredentialReconciler reconciles Rancher Cloud Credentials in the cattle-global-data
+// namespace into CAPA-specific AWSClusterStaticIdentity resources. This enables users to reuse
+// Rancher Cloud Credentials when provisioning AWS clusters with CAPI/CAPA.
+type RancherCredentialReconciler struct {
+	Client client.Client
+
+	// CAPASystemNamespace is the namespace where the CAPA controller is installed and where
+	// the credentials secret will be created. Defaults to "capa-system".
+	CAPASystemNamespace string
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *RancherCredentialReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, options controller.Options) error {
+	if r.CAPASystemNamespace == "" {
+		r.CAPASystemNamespace = capaProviderNamespace
+	}
+
+	log := log.FromContext(ctx)
+
+	log.Info("Watching AWS Cloud Credential secrets")
+
+	secretPredicate := predicate.NewPredicateFuncs(func(obj client.Object) bool {
+		return obj.GetNamespace() == rancherCredentialsNamespace &&
+			obj.GetAnnotations()[driverNameAnnotation] == awsDriverAnnotationValue
+	})
+
+	capiProviderPredicate := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			providerObj, ok := e.Object.(*turtlesv1.CAPIProvider)
+			if !ok {
+				return false
+			}
+
+			return providerObj.Spec.Type == turtlesv1.Infrastructure &&
+				providerObj.Spec.Name == capaProviderSpecName
+		},
+
+		UpdateFunc: func(_ event.UpdateEvent) bool {
+			return false
+		},
+
+		DeleteFunc: func(_ event.DeleteEvent) bool {
+			return false
+		},
+
+		GenericFunc: func(_ event.GenericEvent) bool {
+			return false
+		},
+	}
+
+	if err := ctrl.NewControllerManagedBy(mgr).
+		Named("rancher-credential-translation").
+		WithOptions(options).
+		For(&corev1.Secret{}, builder.WithPredicates(secretPredicate)).
+		Watches(
+			&turtlesv1.CAPIProvider{},
+			handler.EnqueueRequestsFromMapFunc(r.enqueueAllMatchingSecrets),
+			builder.WithPredicates(capiProviderPredicate),
+		).
+		Complete(r); err != nil {
+		return fmt.Errorf("creating RancherCredential translation controller: %w", err)
+	}
+
+	return nil
+}
+
+//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;patch
+//+kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=awsclusterstaticidentities,verbs=get;list;watch;create;update;patch;delete
+
+// Reconcile watches Rancher Cloud Credentials for AWS and translates them into
+// AWSClusterStaticIdentity resources for use with CAPA.
+func (r *RancherCredentialReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	credential := &corev1.Secret{}
+	if err := r.Client.Get(ctx, req.NamespacedName, credential); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// verify that the CAPA CAPIProvider is available and ready
+	var provider turtlesv1.CAPIProvider
+
+	err := r.Client.Get(ctx, types.NamespacedName{
+		Name:      capaProviderSpecName,
+		Namespace: capaProviderNamespace,
+	}, &provider)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+
+		return ctrl.Result{}, err
+	}
+
+	if provider.Status.Phase != turtlesv1.Ready {
+		return ctrl.Result{RequeueAfter: defaultRequeueDuration}, nil
+	}
+
+	if err := r.Client.Get(ctx, req.NamespacedName, credential); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	if credential.GetAnnotations()[driverNameAnnotation] != awsDriverAnnotationValue {
+		return ctrl.Result{}, nil
+	}
+
+	if !credential.DeletionTimestamp.IsZero() {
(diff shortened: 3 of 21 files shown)


charts/rancher-turtles/templates/rancher-turtles-components.yaml
```diff
@@ -3576,6 +3576,7 @@ rules:
   - infrastructure.cluster.x-k8s.io
   resources:
   - '*'
+  - awsclusterstaticidentities
```
anmazzotti: isn't this redundant with the line above?

config/rbac/role.yaml
```diff
@@ -53,6 +53,7 @@ rules:
   - infrastructure.cluster.x-k8s.io
   resources:
   - '*'
+  - awsclusterstaticidentities
```
anmazzotti, line 56: Same as the other comment, should be covered already by '*'.

examples/clusterclasses/aws/rke2/clusterclass-ec2-rke2-example.yaml
```diff
@@ -73,7 +73,6 @@ spec:
         openAPIV3Schema:
           description: The AWSClusterStaticIdentity resource name referencing the credentials to create the Cluster.
           type: string
-          default: cluster-identity
```
anmazzotti, line 76: Don't understand the rationale behind this change. 
It's just a convention we use, so it's fine to remove it, just miss why within the context of this PR.

examples/clusterclasses/aws/rke2/clusterclass-ec2-rke2-example.yaml
```diff
@@ -167,7 +166,7 @@ spec:
     spec:
       identityRef:
         kind: AWSClusterStaticIdentity
-        name: cluster-identity
+        name: default
```
anmazzotti, line 169: Not sure about this change either, why is it now `default`?

internal/controllers/credentialtranslation_controller.go
```diff
@@ -0,0 +1,349 @@
...
+
+// RancherCredentialReconciler reconciles Rancher Cloud Credentials in the cattle-global-data
+// namespace into CAPA-specific AWSClusterStaticIdentity resources. This enables users to reuse
+// Rancher Cloud Credentials when provisioning AWS clusters with CAPI/CAPA.
+type RancherCredentialReconciler struct {
+	Client client.Client
+
+	// CAPASystemNamespace is the namespace where the CAPA controller is installed and where
+	// the credentials secret will be created. Defaults to "capa-system".
+	CAPASystemNamespace string
```
anmazzotti: Don't understand why this is not a constant.
We expect `capa-system` in most cases, so without any logic to lookup the correct namespace I think having this configurable is misleading.

internal/controllers/credentialtranslation_controller.go
```diff
@@ -0,0 +1,349 @@
...
+	rancherAWSAccessKeyField = "amazonec2credentialConfig-accessKey"
+
+	// rancherAWSSecretKeyField is the field name in Rancher Cloud Credential secrets for the AWS secret key.
+	rancherAWSSecretKeyField = "amazonec2credentialConfig-secretKey"
+)
+
+// RancherCredentialReconciler reconciles Rancher Cloud Credentials in the cattle-global-data
+// namespace into CAPA-specific AWSClusterStaticIdentity resources. This enables users to reuse
+// Rancher Cloud Credentials when provisioning AWS clusters with CAPI/CAPA.
+type RancherCredentialReconciler struct {
```
anmazzotti, line 62: The entire implementation of this reconciler is AWS specific.
I think this needs to be addressed by making it provider agnostic already, or maybe find a more specific name. 

In the current form if we want one reconciler only, adding another credential type to the logic is going to be a whole lot of refactoring, so it's not future proof.

internal/controllers/helpers.go
```diff
@@ -54,6 +54,9 @@ const (
 	fleetNamespaceMigrated    = "cluster-api.cattle.io/fleet-namespace-migrated"
 	fleetDisabledLabel        = "cluster-api.cattle.io/disable-fleet-auto-import"
 
+	rancherCredentialsNamespace = "cattle-global-data"
```
anmazzotti, line 57: We already have logic to translate credentials in `cattle-global-data`. 
That is by using CAPIProvider.spec.credentials.rancherCloudCredentials

internal/controllers/credentialtranslation_controller.go
```diff
@@ -0,0 +1,349 @@
...
+		return providerObj.Spec.Type == turtlesv1.Infrastructure && providerObj.Spec.Name == capaProviderSpecName
+	})
+
+	if err := ctrl.NewControllerManagedBy(mgr).
+		Named("rancher-credential-translation").
+		WithOptions(options).
+		For(&corev1.Secret{}, builder.WithPredicates(secretPredicate)).
+		Watches(
+			&turtlesv1.CAPIProvider{},
+			handler.EnqueueRequestsFromMapFunc(r.enqueueAllMatchingSecrets),
```
anmazzotti, line 120: How are changes to the CAPIProvider related to credentials?
Like for example is this going to `enqueueAllMatchingSecrets` when the provider is updated to a newer version and why is that?

internal/controllers/credentialtranslation_controller.go
```diff
@@ -0,0 +1,349 @@
...
+	}, &provider)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	if provider.Status.Phase != turtlesv1.Ready {
+		return ctrl.Result{}, nil
```
anmazzotti: Worth to add some log line here.

anmazzotti: Also shouldn't this requeue?

internal/controllers/credentialtranslation_controller.go
```diff
@@ -0,0 +1,349 @@
...
+// reconcileCredentialSecret creates or updates the credentials Secret in the CAPA system namespace.
+func (r *RancherCredentialReconciler) reconcileCredentialSecret(ctx context.Context, name, accessKeyID, secretAccessKey string) error {
+	credSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: r.CAPASystemNamespace,
+		},
+	}
+
+	_, err := controllerutil.CreateOrUpdate(ctx, r.Client, credSecret, func() error {
```
anmazzotti: I'm afraid about the Update part. 
What if a secret with the same name already exists in this namespace? 
How does the controller know what we are not overriding a secret already in use by some other cluster identity?

internal/controllers/credentialtranslation_controller.go
```diff
@@ -0,0 +1,349 @@
...
+}
+
+// reconcileAWSClusterStaticIdentity creates or updates the AWSClusterStaticIdentity referencing
+// the credentials secret.
+func (r *RancherCredentialReconciler) reconcileAWSClusterStaticIdentity(ctx context.Context, name string) error {
+	awsIdentity := r.awsClusterStaticIdentity(name)
+
+	identitySpec := map[string]interface{}{
+		"secretRef":         name,
+		"allowedNamespaces": map[string]interface{}{},
```
anmazzotti: This creates a global credential.
This is something we are trying to deprecate.

Global credentials are already supported and synced by the CAPIProvider.spec.credentials.rancherCloudCredentials. This logic already exists in Turtles today.

internal/controllers/credentialtranslation_controller.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	_, err := controllerutil.CreateOrUpdate(ctx, r.Client, awsIdentity, func() error {
```
anmazzotti: Same as for the other comment, I worry about the Update here.
Nothing checks that we are not overriding an identity already in use for something else which happens to have the same name.

anmazzotti (dismissed):
I skipped reviewing the test coverage because I think the logic misses the point.
In general this is a replica of a feature that already exists which syncs `cattle-global-data` credentials with providers credentials.
Unfortunately this is not documented, but you can see the `CAPIProvider.spec.credentials.rancherCloudCredential` in the reference doc: https://turtles.docs.rancher.com/turtles/stable/en/reference/capiprovider.html

See here for the implementation: https://github.com/rancher/turtles/blob/release/v0.27/internal/controllers/operator_reconciler.go#L489

The difference is that this new controller seem to create global ClusterIdentities instead of assigning global credentials to the CAPA controller (like the existing functionality does), which to me is kind of worse.

This defeats the purpose of ClusterCredentials entirely, as they are not generally meant to be shared and they should be restricted to one namespace only where the Cluster is deployed in isolation.

This new controller, being opaque to the user, introduces a security breach where it's not exactly clear how credentials are handled and propagated within the management cluster.

At least `CAPIProvider.spec.credentials.rancherCloudCredential` is explicit since has to be set by the user, it only lacks documentation.

examples/clusterclasses/aws/rke2/clusterclass-ec2-rke2-example.yaml
salasberryfin: I'd like to make sure that if the patch is not applied correctly, the test fails because it points to a non-existing `AWSClusterStaticIdentity`. If it still points to `cluster-identity` and the patch is not correctly applied, the test would pass using the identity created for other AWS provisioning tests.

internal/controllers/helpers.go
salasberryfin: `CAPIProvider.spec.credentials.rancherCloudCredentials` does not cover the functionality that we need for CAPI Plug and Play because it forces you to set the credential at the `CAPIProvider` layer, and there's no way to add user management. 

Whether this feature is actually being used and it is sensible to maintain it is a different topic, in my opinion.

internal/controllers/credentialtranslation_controller.go
salasberryfin: This is a good point. It should only trigger when there's a `Create` event, because what we're trying to do is detect that the AWS provider is installed to reconcile the credential translation.

Any other event is likely not relevant, because deleting the provider won't delete the CRDs (which is what we technically require from the provider) and updates should just be ignored (no reconciliation).

I'll take a look at it.

internal/controllers/credentialtranslation_controller.go
salasberryfin: The `Secret` name here is the same as the `Secret` name used for the Rancher Cloud Credential. For example:

- Rancher Cloud Credential `Secret`:
  - name: cc-1234
  - namespace: cattle-global-data
- AWS identity `Secret`:
  - name: cc-1234
  - namespace: capa-system
  
Effectively, to have a name collision in the translated `Secret` we would have had a prior collision in the Rancher Cloud Credential.

examples/clusterclasses/aws/rke2/clusterclass-ec2-rke2-example.yaml
salasberryfin: We have `required: true` and then a default.

config/rbac/role.yaml
salasberryfin: For some reason this is automatically added by `make generate`.

salasberryfin:
As stated in the PR description:

> [...] it cannot be enabled until the second part of this feature (validation webhook) is implemented adding a security layer [...]

this feature is not complete without the security validation and won't be available to users unless the validation webhook is implemented and enabled. However, we need to have the translation logic to unblock feature development.

The `rancherCloudCredential` feature of the CAPIProvider ties a single Rancher Cloud Credential to be used globally with the provider, which is not what [we need](https://github.com/SUSE/rancher-architecture/pull/58) for CAPI Plug and Play. The ultimate goal is to bring Rancher Cloud Credential management to CAPI provisioning. In my opinion, considering the push to have CAPI Plug and Play, I don't think it is worth maintaining `rancherCloudCredential` sync in `CAPIProvider`, but this is just my personal opinion.

internal/controllers/credentialtranslation_controller.go
anmazzotti: This assumes that:
1. There are no Secrets named `cc-1234` in `capa-system` namespace created by users from other sources.
2. The `cc-*` naming scheme will never change raising the odds of collisions with already existing secrets (see .1)

internal/controllers/credentialtranslation_controller.go
salasberryfin: I updated the predicate to trigger on `CAPIProvider` AWS creation only. All other events should not trigger a reconcile loop.

internal/controllers/credentialtranslation_controller.go
salasberryfin: Fair point. What do you suggest to prevent collisions considering the controller must update the `Secret` when the original Rancher Cloud Credential changes? i.e. if my AWS keys expire and I update them in Rancher, I need my static identity to be updated accordingly.

internal/controllers/credentialtranslation_controller.go
salasberryfin: Thanks for catching this. Updated!

internal/controllers/credentialtranslation_controller.go
salasberryfin: I will try to refactor this to make it agnostic. For now I wanted to validate the AWS logic.

salasberryfin: @anmazzotti I created a new version of the controller which provides a translation interface to make it provider-agnostic.

jiaqiluo (changes requested):
Update on 6/8:

The webhook will add typed credential validation for AWSCluster and AWSClusterStaticIdentity, so no annotation is required at this stage. 

We'll continue iterating on the annotation and configuration approach and implement it once we have a more complete design. 

<details>
<summary>Original Message</summary>

Could you please make the following changes that are needed for Rancher Webhook to work: 

When installing the CAPA's CRDs
1. patch the `awsclusterstaticidentities.infrastructure.cluster.x-k8s.io` CRD to add the following annotation  `turtles-capi.cattle.io/credential-policy: '{"credentialRef":{"kind":"Secret","name":".spec.secretRef","namespace":"capa-system"}}'`
2. patch the `awsclusters.infrastructure.cluster.x-k8s.io` CRD to add the following annotation:
`turtles-capi.cattle.io/credential-policy: '{"credentialRef":{"kind":".spec.identityRef.kind","name":".spec.identityRef.name"}}'`


The annotation name `turtles-capi.cattle.io/credential-policy` can be changed if you have a better option. 



</details>

jakefhyde (dismissed):
Approving (pending @jiaqiluo's comment)

internal/controllers/credentialtranslation_controller.go
salasberryfin: @anmazzotti the latest version of the code introduces a way to detect naming collisions on `Secret` and `AWSClusterStaticIdentity` using the original `Secret`'s UID.
- Before creating any translation object, it will verify if there are other objects with the same name.
- If there are any, it will verify that they are annotated with a reference to the UID of the credential `Secret`. If not annotated, the controller will not create any translation resources.
- When no collisions are found, the translated resources are annotated with the reference to the original `Secret`'s UID.
- On deletion of a Rancher Cloud Credential, it will once again verify the annotation to avoid removing resources that caused a collision.

salasberryfin:
The PR's now updated to prevent naming collisions when creating the translated objects. For AWS `Secret` and `AWSClusterStaticIdentity`.

jiaqiluo (dismissed):

salasberryfin:
As discussed on our most recent call, I will modify `allowedNamespaces` in the translated AWS identity to restrict users to create clusters in `fleet-default`.

jiaqiluo (dismissed):

jiaqiluo (approved):

jakefhyde (approved):

furkatgofurov7 (approved):

yiannistri (approved):

Issue #2439, opened 2026-05-22
ClusterGroup is not moved from original fleet workspace to `fleet-default` after running `migrate_caapf.sh post'


thehejik:
### What steps did you take and what happened?

Follow-up for https://github.com/rancher/turtles/pull/2315 where we noticed that after successful migration of clusters to dedicated `fleet-default` workspace its Fleet ClusterGroup (which is created automatically by [caapf addon](https://github.com/rancher/cluster-api-addon-provider-fleet/pull/11) for every clusterclass) is not moved together with the cluster(s).

It might be a problem when fleet is targeting a workload via this ClusterGroup, workaround would be to manually create the ClusterGroup in `fleet-default` workspace with the same name and deleting it from original workspace like `capi-clusters`.

<img width="1246" height="637" alt="Image" src="https://github.com/user-attachments/assets/30e07a96-d98f-4cb7-8d43-b5aa6aaefdb3" />

### What did you expect to happen?

The group will be moved together with clusters

### Rancher Turtles version

Spotted on Prime 2.14.1

### Label(s) to be applied

/kind bug

thehejik:
Fixed migration script works as expected, the `clusterGroup` originally created on 2.13.6 by caapf withing `capi-clusters` workspace has been migrated to `fleet-default` workspace together with the cluster and the cluster can be targeted by using the migrated group in fleet on rancher 2.14.2.

I just had to add `spec.helm.force: true` on cni helmop resource otherwise the `post` phase stuck.

Tested on prime-alpha/2.13.6-rc1 with capd+kubeadm cluster and upgraded to prime-alpha/2.14.2-alpha8 (used outdated pre-release builds) with latest migrate-caapf.sh script.

capi-custers ws without the group
<img width="1064" height="328" alt="Image" src="https://github.com/user-attachments/assets/db3c0f10-d473-4691-a64c-f616a565fc1c" />

fleet-default ws contains the migrated group
<img width="1064" height="328" alt="Image" src="https://github.com/user-attachments/assets/1177bc49-423d-443e-b8e9-20e9272dd169" />

PR #2464, opened 2026-06-02
fix: Migrate ClusterGroup objects

yiannistri:

This PR updates the CAAPF migration script in two ways:

1. It migrates ClusterGroup objects that CAAPF originally created, from the CAPI namespaces to `fleet-default`, after verifying that they don't cause any name collisions.
2. It sets `keepResources: true` for any Fleet resources in CAPI/ClusterClass namespaces so that they don't get garbage collected when the new fleet-agent starts running on the downstream cluster, and removes it as part of the post phase. This was noticed when the migration happened without an upgrade taking place.

Relates to #2439 

It may be easier to follow the changes by reviewing each commit separately.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


valaparthvi (approved):
Tested on same rancher version and it works as expected when modified with a few additional config to `.spec.helm.force: true` (but this part still needs to be tested further).

salasberryfin (approved):
Thanks

PR #2481, opened 2026-06-06
fix: avoid secondary envtest panic on startup failure

immanuwell:
When envtest startup blows up, `Build()` panics before returning and teardown can hit a second nil panic in `Stop()`. That hides the real failure and makes the output kinda noisy.

This patch returns the startup error from `Build()`, makes `Stop()` safe for nil or partial envs, and adds unit tests for both paths.

N/A

Repro before fix:
1. Run `go test ./util/predicates -run TestClusterPredicates -count=1` in an env where envtest startup fails, for example missing `KUBEBUILDER_ASSETS`.
2. You get the real startup error, then a second nil pointer panic in teardown. not great

After this patch, you only get the real startup failure.

I also ran `make test` after the change.

**Checklist**:
- [ ] includes documentation
- [x] adds unit tests


anmazzotti (changes requested):
Thank you for your PR.
This is technically not needed, users are expected to use the makefile (make test) that uses setup-envtest.

If not then I think a panic is appropriate.

immanuwell:
@anmazzotti got it, so I think I should close this PR

Issue #2363, opened 2026-05-06
release: ensure dependencies are updated in release branches


anmazzotti:
This issue is about implementing dependencies bump automation for **release branches**.

Acceptance Criteria:
- Automation should only target branches for releases that are maintained and supported.
- Automation should only bump patch releases, no minor/major bumps.
- Changes should be tested by all available e2e coverage (ex. run e2e tests on release branches)
- Bumps should include:
  - go version
  - `go.mod` dependencies
  - CAPI providers (including Core)
  - `rancher/kuberlr-kubectl` image

yiannistri:
Renovate is now used for main as well as our supported release branches.

PR #2492, opened 2026-06-10
ci: add renovate workflow and config

yiannistri:

This PR introduces a CI workflow that runs Renovate daily as well as a Renovate config file that constraints the versions of some of the dependencies to better align them with upstream CAPI. For now only `release/v0.26` will receive dependency bump PRs, this was intentional to avoid flooding us with PRs until we fine-tune the config further.  

Related to #2363 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,0 +1,42 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "github>rancher/renovate-config//rancher-main#release"
+  ],
+  "devbox": {
+    "enabled": false
+  },
+  "baseBranchPatterns": [
+    "release/v0.26"
+  ],
+  "packageRules": [
+    {
+      "description": "Disable rancher/turtles updates on dockerfiles. This avoids PRs that bump the version tag in Turtles Helm chart.",
+      "matchPackageNames": ["rancher/turtles"],
+      "matchDatasources": ["docker"],
+      "enabled": false
+    },
+    {
+      "description": "Align with core CAPI v1.12 dependencies",
+      "matchBaseBranches": ["release/v0.26"],
+      "matchPackageNames": [
+        "sigs.k8s.io/controller-runtime"
+      ],
+      "allowedVersions": "<0.23.0"
+    },
+    {
+      "description": "Align with core CAPI v1.12 dependencies",
+      "matchBaseBranches": ["release/v0.26"],
+      "matchPackageNames": [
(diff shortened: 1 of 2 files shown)


furkatgofurov7 (dismissed):
Thank you

.github/renovate.json
```diff
@@ -0,0 +1,42 @@
...
+      "matchPackageNames": [
+        "sigs.k8s.io/controller-runtime"
+      ],
+      "allowedVersions": "<0.23.0"
+    },
+    {
+      "description": "Align with core CAPI v1.12 dependencies",
+      "matchBaseBranches": ["release/v0.26"],
+      "matchPackageNames": [
+        "k8s.io/**"
```
anmazzotti, line 31: Wouldn't this also hit `k8s.io/klog/v2` that follows its own version? or is that ignored in this case since it's at `v2.*.*` already?

.github/renovate.json
```diff
@@ -0,0 +1,42 @@
...
+      ],
+      "allowedVersions": "<0.35.0"
+    },
+    {
+      "description": "Pin rancher/kuberlr-kubectl to version 7 (v7.x.x)",
+      "matchBaseBranches": ["release/v0.26"],
+      "matchPackageNames": ["rancher/kuberlr-kubectl"],
+      "allowedVersions": "^7.0.0"
```
anmazzotti: Not sure why there is a rule if the version is pinned.
Should this be `<8.0.0` instead?

.github/renovate.json
yiannistri: The pattern matching does make it eligible for dependency bumps, so in this case I would expect to bump it in `release/v0.26` from v2.130.1 to the latest v2.140.0. This looks a bit odd because the package name itself includes a version (`k8s.io/klog/v2`) but that should be ignored and just look at the actual version of that package to see if it requires any updates.

.github/renovate.json
yiannistri: Actually this is interpreted as <8.0.0 (example [here](https://github.com/masterminds/semver#caret-range-comparisons-major)) but I get how this can be interpreted as regex so let me change it.

yiannistri: Updated

anmazzotti (approved):

salasberryfin (approved):
Thanks @yiannistri

furkatgofurov7 (commented):
/approve

PR #2516, opened 2026-06-15
fix: ci flakes when deleting capiprovider object

salasberryfin:

Deleting the `CAPIProvider` object on `AfterEach` intermittently fails because the object is modified in between fetching and deleting. Looks like this is related to re-using a pointer for the CAPA provider. This change uses a copy of the provider template instead.

Additionally, the test that checks the requeue on cases where the `CAPIProvider` is not yet `Ready` is now embedded in other test case to reduce repetition.

Fixes #

I've executed the CI check multiple times trying to reproduce the original issue and I haven't been able to do it.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


furkatgofurov7 (approved):
Do we need to backort this change?

salasberryfin:
@furkatgofurov7 yes, we do!

/backport v2.15.0 release/v0.27

yiannistri (approved):

PR #2521, opened 2026-06-15
fix: replace cr upload with gh CLI to fix asset upload on re-run and unblock immutable releases

### LINKED ISSUE #2359: chart-release workflow does not comply with immutable releases

### LINKED ISSUE #2421: Enable fully immutable releases in Turtles
### Path to immutable releases

~https://github.com/rancher/turtles/issues/2359 was fixed in https://github.com/rancher/turtles/pull/2416 which carves a path towards achieving a fully immutable release process in Turtles (Phase 1 of a gradual migration):~

~Phase 1: Draft-first flow, manual undraft. Gives us confidence that the suggested process works reliably without the risk of being stuck with a broken immutable release. (✅ completed now). This issue is track Phase 2 and 3 of this gradual migration.~

~Phase 2 (follow-up, once Phase 1 is validated over a few releases): Uncomment the Publish release step already included in the workflow. This automates the undraft as the very last step, only after all assets are confirmed present, fully following [GitHub's recommended immutable release workflow](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases)~

Phase 3 (no code change needed, a follow-up once Phase 2 is validated): Enable immutable releases in the
repository settings under Settings -> Releases -> "Enable release immutability".

We can start tackling this issue after few release using the ~current~ upcoming worklow ~(Phase 1 state)~ and once we sure and confident we won't have any surprises and failures, this can be picked up and completed.

furkatgofurov7:
`cr upload --skip-existing` silently skips asset upload when a release already exists, leaving the GitHub release with no .tgz attached on re-runs after a partial failure. This was papered over by a recovery step using `gh release upload --clobber`, but that recovery step itself violates immutable releases once enabled (modifying an already-published release is not allowed).

This PR replaces `cr upload` and the recovery step with explicit GH CLI steps:

- Create or reuse a draft release idempotently (`gh release create`, skips gracefully if the release already exists)
- Always upload the chart asset with `--clobber `(safe on re-runs, replaces the recovery pattern with a proactive upload)
- Hard-fail if the asset is missing after upload (no silent failures)
- Publish the release (`gh release edit --draft=false`) before cr index, because `GET/repos/{owner}/{repo}/releases/tags/{tag}` only returns published releases and cr index would silently produce no index update against a draft

Fixes #2359 
Partially fixes: #2421 

Changes were tested against my personal fork repo (xref: [source code](https://github.com/rancher/turtles/compare/main...furkatgofurov7:turtles:main), with additional changes to make it fork-runnable), and a [test release](https://github.com/furkatgofurov7/turtles/releases/tag/v0.99.7-test) was successfully created and published (not marked as latest, but published with assets).

This PR needs a backport once merged.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/chart-release.yml
+++ b/.github/workflows/chart-release.yml
@@ -58,34 +58,44 @@ jobs:
           rm -rf .cr-index
           mkdir -p .cr-index
 
-      - name: Run chart-releaser upload
-        run: cr upload --skip-existing -c "$(git rev-parse HEAD)" --generate-release-notes --release-name-template "${{ inputs.tag }}" --make-release-latest=false
+      - name: Create draft release
+        run: |
+          if gh release view "${{ inputs.tag }}" &>/dev/null; then
+            echo "Release '${{ inputs.tag }}' already exists, reusing."
+          else
+            gh release create "${{ inputs.tag }}" \
+              --draft \
+              --latest=false \
+              --generate-notes \
+              --title "${{ inputs.tag }}" \
+              --target "$(git rev-parse HEAD)"
+          fi
 
-      - name: Verify chart asset present (recover if missing)
+      - name: Upload chart asset
         run: |
           CHART_FILE=$(ls ${{ inputs.release_dir }}/*.tgz 2>/dev/null | head -1)
           if [ -z "${CHART_FILE}" ]; then
             echo "ERROR: No .tgz chart package found in ${{ inputs.release_dir }}"
             exit 1
           fi
+          gh release upload "${{ inputs.tag }}" "${CHART_FILE}" --clobber
+          echo "Uploaded $(basename ${CHART_FILE})"
+
+      - name: Verify chart asset present
+        run: |
+          CHART_FILE=$(ls ${{ inputs.release_dir }}/*.tgz | head -1)
           ASSET_NAME=$(basename "${CHART_FILE}")
-
-          if ! gh release view "${{ inputs.tag }}" > /dev/null 2>&1; then
-            echo "ERROR: Release '${{ inputs.tag }}' not found."
-            exit 1
-          fi
-
           FOUND=$(gh release view "${{ inputs.tag }}" \
             --json assets \
             --jq ".assets[] | select(.name == \"${ASSET_NAME}\") | .name")
-
           if [ -z "${FOUND}" ]; then
-            echo "Chart asset '${ASSET_NAME}' missing — uploading now..."
-            gh release upload "${{ inputs.tag }}" "${CHART_FILE}" --clobber
-            echo "Upload complete."
-          else
-            echo "Chart asset '${ASSET_NAME}' confirmed present."
+            echo "ERROR: Asset '${ASSET_NAME}' still missing after upload. Failing."
+            exit 1
           fi
+          echo "Asset '${ASSET_NAME}' confirmed present."
+
+      - name: Publish release
+        run: gh release edit "${{ inputs.tag }}" --draft=false --latest=false
 
       - name: Run chart-releaser index
         run: cr index --push --release-name-template "${{ inputs.tag }}"


salasberryfin (approved):

.github/workflows/chart-release.yml
```diff
@@ -58,34 +58,44 @@ jobs:
...
+            echo "ERROR: Asset '${ASSET_NAME}' still missing after upload. Failing."
+            exit 1
           fi
+          echo "Asset '${ASSET_NAME}' confirmed present."
+
+      - name: Publish release
+        run: gh release edit "${{ inputs.tag }}" --draft=false --latest=false
 
       - name: Run chart-releaser index
         run: cr index --push --release-name-template "${{ inputs.tag }}"
```
anmazzotti, line 101: Do we still need this? If not we can also remove the installation of this tool

.github/workflows/chart-release.yml
```diff
@@ -58,34 +58,44 @@ jobs:
           rm -rf .cr-index
           mkdir -p .cr-index
 
-      - name: Run chart-releaser upload
-        run: cr upload --skip-existing -c "$(git rev-parse HEAD)" --generate-release-notes --release-name-template "${{ inputs.tag }}" --make-release-latest=false
+      - name: Create draft release
+        run: |
+          if gh release view "${{ inputs.tag }}" &>/dev/null; then
```
anmazzotti, line 63: Confusing. 
Is this workflow supposed to create the release or not? 

I don't like the idea of dynamic workflows and conditions, too much unpredictable and I don't see the purpose.

.github/workflows/chart-release.yml
```diff
@@ -58,34 +58,44 @@ jobs:
...
-            echo "Upload complete."
-          else
-            echo "Chart asset '${ASSET_NAME}' confirmed present."
+            echo "ERROR: Asset '${ASSET_NAME}' still missing after upload. Failing."
+            exit 1
           fi
+          echo "Asset '${ASSET_NAME}' confirmed present."
+
+      - name: Publish release
+        run: gh release edit "${{ inputs.tag }}" --draft=false --latest=false
```
anmazzotti, line 98: Is this workflow supposed to finalize the release or just upload the chart?
Note that with immutable releases this would be final, no more changes to the release are allowed, the following steps modifying the release would fail.

Shouldn't the caller workflow (release.yaml) finalize it instead? Or should it be a manual step maybe?

.github/workflows/chart-release.yml
furkatgofurov7: Yes we do need it, without this step the Helm index on gh-pages never gets updated, and for example `helm repo add rancher-turtles` users would never see the new chart version. The release would exist on GitHub but be invisible to Helm without this step

.github/workflows/chart-release.yml
furkatgofurov7: @anmazzotti This condition exists specifically to make the workflow re-runnable after partial failures without manual intervention. I would not say it is unpredictable behaviour, but more of a deliberate recovery strategy to me

anmazzotti: Thank you, makes sense when you put it that way.

.github/workflows/chart-release.yml
anmazzotti: I didn't know https://rancher.github.io/turtles/index.yaml exists.
Are we using this in any form or capacity? Turtles is not meant to be installed standalone anymore.

I don't have a strong opinion on this, it's fine to me to maintain this index if we need to, I was more looking into deprecating the `cr` tool entirely if possible.

furkatgofurov7: I don't know of usage, but to remove cr tool, it needs more of refactoring the workflow and removing the tool probably and we can do it once we are sure it does not have any usage for helm index

.github/workflows/chart-release.yml
furkatgofurov7: What you are suggesting is a significant restructure: we need to split chart-release.yml into two workflows, add a publish job in release.yaml between them, move cr index and Install chart-releaser out which probably has its advantages but I am not sure if it is worth the effort.

The current order is:

- Draft created
- Asset uploaded
- Asset verified (hard fail)
- Publish: release is complete and confirmed before this runs
- cr index: read-only, just reads the published release and updates gh-pages

Just to clarify: cr index does not modify the release. It reads the asset, calculates the hash, and pushes to gh-pages; that is it.

anmazzotti (approved):

Issue #2505, opened 2026-06-11
Remove CAAPF from AWS/RKE2 example


yiannistri:
Remove CAAPF usage from the e2e short test using AWS/RKE2 ClusterClass.

cpinjani:
Re-opening for validation, thanks

cpinjani:
Validation passed on v2.15.0-alpha21, Turtles v0.27.0-rc.7

Rancher:

<img width="1252" height="252" alt="Image" src="https://github.com/user-attachments/assets/264ab3b4-f1a3-4bbc-a440-08e2e516456e" />

Fleet:

<img width="1897" height="316" alt="Image" src="https://github.com/user-attachments/assets/acddab04-d0cc-45c1-958c-981b2926553e" />

CAPI cluster:

<img width="1896" height="908" alt="Image" src="https://github.com/user-attachments/assets/65562492-91fd-4f93-ad52-7b148b73d186" />

PR #2512, opened 2026-06-15
chore: deprecate CAAPF usage in Azure/RKE2 example

anmazzotti:
**What this PR does / why we need it**:

PR for https://github.com/rancher/turtles/issues/2506

Test run: https://github.com/rancher/turtles/actions/runs/27559235873

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits into logical changes
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

PR #2527, opened 2026-06-16
test: refactor AWS/EC2/RKE2 test away from CAAPF

yiannistri:

This PR updates the AWS/EC2/RKE2 test to use CRS and the RKE2's ability to define a CNI in order to no longer rely on CAAPF. 

Long test run: https://github.com/rancher/turtles/actions/runs/27680377285/job/81866498200

Related to #2505 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/test/e2e/data/cluster-templates/aws-ec2-rke2-topology.yaml
+++ b/test/e2e/data/cluster-templates/aws-ec2-rke2-topology.yaml
@@ -3,7 +3,6 @@ kind: Cluster
 metadata:
   labels:
     cloud-provider: aws
-    cni: calico
     csi: aws-ebs-csi-driver
   name: ${CLUSTER_NAME}
   namespace: ${NAMESPACE}
@@ -28,7 +27,7 @@ spec:
       - name: amiID
         value: ${AWS_AMI_ID}
       - name: cni
-        value: none
+        value: calico
       - name: awsClusterIdentityName
         value: ${AWS_CLUSTER_IDENTITY_NAME}
     version: ${RKE2_KUBERNETES_VERSION}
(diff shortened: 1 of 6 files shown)


test/e2e/data/cluster-templates/aws-ec2-rke2-topology.yaml
```diff
@@ -28,7 +27,7 @@ spec:
       - name: amiID
         value: ${AWS_AMI_ID}
       - name: cni
-        value: none
+        value: calico
```
anmazzotti, line 30: nit pick: for the other RKE2 examples I just let this empty and it will default to RKE2 Canal chart, so unless we have reasons to use Calico we can just default to what RKE2 uses out of the box

anmazzotti (dismissed):
I do miss one commit in this PR https://github.com/rancher/turtles/commit/b630ec67ecdb8f44b6d74afa61b81cb5dae6cf87

Was this omitted to not conflict with https://github.com/rancher/turtles/pull/2512?
You can still include the commit imho to make the PR complete, we can fix conflicts depending on what is merged first, we will have the same problem with the suite_test.go file anyway, so no worries.

Either way is ok to me.

test/e2e/data/cluster-templates/aws-ec2-rke2-topology.yaml
yiannistri: I just left it as calico as this was used previously but I'm happy to leave this empty. No other reasons.

yiannistri:
Yes exactly, I pushed the commit that included your change only to get the long test run but omitted it here to avoid conflicts.

test/e2e/data/cluster-templates/aws-ec2-rke2-topology.yaml
yiannistri: Actually, I've realised that the AWSClusterTemplate includes ingress rules that are specific for Calico i.e. https://github.com/rancher/turtles/blob/9a5bd8ce9741459d4c649693de9d3cbf9e77e7a1/examples/clusterclasses/aws/rke2/clusterclass-ec2-rke2-example.yaml#L212-L223
So I think we should keep this as is but maybe add an issue to update this clusterclass to deal with the different CNI rules that RKE2 supports wdyt?

anmazzotti: Good point, I didn't think about this.
It's fine to use Calico as example, may get tricky to support more.

salasberryfin (dismissed):

anmazzotti (approved):

salasberryfin (approved):

PR #2602, opened 2026-06-30
fix: ci flakes on cleanup

salasberryfin:

Trying to resolve the intermittent flakes of the CI tests because cleanup is failing on `AfterEach`.

After some investigation I suspect there may be some problem with write operations hitting the API directly and the cache not being updated accordingly.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/internal/test/cleanup.go
+++ b/internal/test/cleanup.go
@@ -91,11 +91,7 @@ func cleanup(ctx context.Context, cl client.Client, objs ...client.Object) error
 		}
 
 		if err := cl.Get(ctx, client.ObjectKeyFromObject(o), copyObj); err != nil {
-			if apierrors.IsNotFound(err) {
-				continue
-			}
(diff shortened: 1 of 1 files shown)


anmazzotti (approved):
I think that the entire logic of removing finalizers should be removed. 
I feel like this is test cheating, and it will hide real failures where controllers are expected to remove their finalizers.

This logic seems to be used by the import controller only, which historically had some problems with finalizers, so could be some legacy test workaround.

So I don't think this PR is improving the situation, rather making it more elaborate.
I think it's fine as a mitigation, but I'd rather see a deeper investigation on why we are removing finalizers in the first place, instead of just deleting objects and wait for the controllers to reconcile.

furkatgofurov7 (approved):

Issue #2695, opened 2026-07-30
CAPG v1.13.0 is now generally available


salasberryfin:
# Description

A new version of the Google provider is now available `v1.13.0` with support for CAPI v1.13.

We should:
1. Create a downstream version of this release.
2. Update pinned provider version.

PR #2706, opened 2026-08-03
chore: bump capg to v1.13.0

salasberryfin:

cluster-api-provider:v1.13.0 is the latest release of the Google provider, which uses CAPI v1.13.2.

Original issue: https://github.com/rancher/turtles/issues/2695

:white_check_mark: Link to successful E2E long: https://github.com/rancher/turtles/actions/runs/30823368376

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/test/e2e/data/applications/cloud-provider-gcp.yaml
+++ b/test/e2e/data/applications/cloud-provider-gcp.yaml
@@ -1,7 +1,8 @@
-# Source: https://raw.githubusercontent.com/kubernetes-sigs/cluster-api-provider-gcp/refs/tags/v1.11.2/test/e2e/data/ccm/gce-cloud-controller-manager.yaml
+# Source: https://raw.githubusercontent.com/kubernetes-sigs/cluster-api-provider-gcp/refs/tags/v1.13.0/test/e2e/data/ccm/gce-cloud-controller-manager.yaml
 #
 # Notes: 
 # 1. Patch the DeamonSet command with the ${CLUSTER_CIDR} variable instead of a static value.
+# 2. Patch the DeamonSet command with the ${GCP_CCM_VERSION} variable: recommend using the same as upstream (see above link).
 ---
 apiVersion: apps/v1
 kind: DaemonSet
@@ -46,7 +47,7 @@ spec:
       serviceAccountName: cloud-controller-manager
       containers:
       - name: cloud-controller-manager
-        image: gcr.io/k8s-staging-cloud-provider-gcp/cloud-controller-manager:master
+        image: gcr.io/k8s-staging-cloud-provider-gcp/cloud-controller-manager:${GCP_CCM_VERSION}
         imagePullPolicy: IfNotPresent
         command:
         - /cloud-controller-manager
--- a/charts/rancher-turtles-providers/values.yaml
+++ b/charts/rancher-turtles-providers/values.yaml
@@ -368,7 +368,7 @@ images:
       tag: v2.13.0
   infrastructureGCP:
     repository: rancher/cluster-api-gcp-controller
-    tag: v1.11.2
+    tag: v1.13.0
   infrastructureVSphere:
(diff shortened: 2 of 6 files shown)


test/e2e/data/applications/cloud-provider-gcp.yaml
```diff
@@ -1,4 +1,4 @@
-# Source: https://raw.githubusercontent.com/kubernetes-sigs/cluster-api-provider-gcp/refs/tags/v1.11.2/test/e2e/data/ccm/gce-cloud-controller-manager.yaml
+# Source: https://raw.githubusercontent.com/kubernetes-sigs/cluster-api-provider-gcp/refs/tags/v1.13.0/test/e2e/data/ccm/gce-cloud-controller-manager.yaml
```
anmazzotti, line 1: @salasberryfin did you review this manifest?
I see there is at least one change with the cloud-controller-manager tag from `master` to `v35.1.3`. Is the intention to keep using `master`? If so may you add it to the comment section?

anmazzotti (changes requested):
Unclear if the CCM manifest was updated accordingly.

Also there's no evidence that the test branch has been ran successfully on this branch, please provide a link with the run in the PR description.

test/e2e/data/applications/cloud-provider-gcp.yaml
salasberryfin: I'm waiting for CI to pass before starting a long E2E. Wouldn't want to run it before confirming that basic tests are passing (and having to run it more than once).

Upstream recently changed the CCM manifest to pin a specific version. We can opt to follow this and pin the same ourselves or keep using the latest.

salasberryfin: Note that cloud-provider-gcp versioning also aligns with Kubernetes minors, which should also be taken into account on the conversation we had recently on providers using the same version of Kubernetes modules as core CAPI, which are not aligned with the newest Kubernetes version supported in Rancher.

For example, we're now supporting Kubernetes up to v1.36, but are still on modules for 1.35 (same as upstream CAPI).

anmazzotti: @salasberryfin it's not the version in use that matters. If we want to keep using `master` that's great to me, sounds like less maintenance.

My point mainly is that you must make a decision and reflect that decision in this PR before submitting it.

I do expect the file to be updated with an additional note that says to keep the tag on `master` if that's the conscious choice we are making with this PR.

This will make things easier for the next developer updating this same manifest when a new version of CAPG is introduced.

I did something similar with the `${CLUSTER_CIDR}` note, hoping that maintainers would read it.

If you don't find this effective enough and we need to automate this somehow or implement any other solution for our manifest drifting from the source, please propose anything you'd like to try, I'm all up for improvements.

As for now we can not simply replace the source URL without additional changes, since the actual source drifted. This is only going to create problems and confusion later on.

salasberryfin:
Link to passing long E2E is now available in the PR's description.

anmazzotti (approved):

furkatgofurov7 (approved):

test/e2e/data/applications/cloud-provider-gcp.yaml
salasberryfin: You are right. I'm fixing this and re-running E2E.

salasberryfin: E2E is green again after setting CCM version as a parameter. A comment has been added recommending using upstream's default value.

anmazzotti: Thank you!

salasberryfin:
There was a flake in one of the Azure tests in the linked E2E execution. GCP test cases all passed, so this should be ready to be merged.

Issue #2086, opened 2026-02-11
[Feature] Toggable `rancher-auto-import`


anmazzotti:
### What would you like to be added (User Story)?

The `cluster-api.cattle.io/rancher-auto-import` label should be toggable so that users can manage the import seamlessly using the same CAPI Cluster API.

### Detailed Description

The auto import can have three states set via the `cluster-api.cattle.io/rancher-auto-import` label:

1. No Label --> No automatic import

```yaml
apiVersion: cluster.x-k8s.io/v1beta2
kind: Cluster
metadata:
  labels: {}
  name: my-capi-cluster
```

2. `cluster-api.cattle.io/rancher-auto-import: "true"` --> Automatic import

```yaml
apiVersion: cluster.x-k8s.io/v1beta2
kind: Cluster
metadata:
  labels:
    cluster-api.cattle.io/rancher-auto-import: "true"
  name: my-capi-cluster
```

3. `cluster-api.cattle.io/rancher-auto-import: "false"` --> No automatic import && management.cattle.io Cleanup

```yaml
apiVersion: cluster.x-k8s.io/v1beta2
kind: Cluster
metadata:
  labels:
    cluster-api.cattle.io/rancher-auto-import: "false"
  name: my-capi-cluster
```

Note that 3. should replace the current `imported: "true"` annotation, so that there should be some conversion from:

```yaml
apiVersion: cluster.x-k8s.io/v1beta2
kind: Cluster
metadata:
  labels: {}
  annotations:
    imported: "true"
  name: my-capi-cluster
```

to:

```yaml
apiVersion: cluster.x-k8s.io/v1beta2
kind: Cluster
metadata:
  labels:
    cluster-api.cattle.io/rancher-auto-import: "false"
  name: my-capi-cluster
```

to avoid reimporting existing Clusters and allow deprecation of the `imported` not fully qualified annotation.

### Label(s) to be applied

/kind feature

PR #2777, opened 2026-08-31
fix: do not disable the import when the cluster record vanishes mid-import

### LINKED ISSUE #2641: Import controller: CAPI cluster can be annotated `imported=true` before the agent manifest is applied, and the import is never retried
While running `CreateUsingGitOpsSpec` (turtles/test v0.26.2 framework, Turtles 109.0.1+up0.26.1 installed by Rancher 2.14.1's system chart) against a CAPI cluster on [cluster-api-provider-harvester](https://github.com/rancher-sandbox/cluster-api-provider-harvester), we hit a state where:

- the `management.cattle.io/v3` cluster record existed with a populated `clusterRegistrationToken` / manifest URL,
- but the downstream cluster had no `cattle-system` namespace content (agent manifest never applied),
- and the CAPI `Cluster` carried the `imported: "true"` annotation, so `import_controller` logged "cluster was imported already and has imported=true annotation set, skipping re-import" on every reconcile — the import never recovered without manual intervention.

**Timeline observed**: on first reconcile the v3 cluster record was created and the controller logged `Import manifest URL not set yet, requeue` (token not generated yet). The initially-created v3 record was then replaced (a second record with a new `c-xxxxx` id appeared); the deletion path of the first record ran `reconcileDelete`, which per the code comment patches the CAPI cluster with `imported=true` "to prevent further-reimports" (import_controller.go, v0.26.2, the `DeletionTimestamp` branch around L234). From that point the annotation short-circuits every future reconcile (L179), so the second, live v3 record never gets its agent applied.

**Repro characteristics**: fresh Rancher + Turtles install, CAPI cluster created via Fleet with namespace auto-import; timing-dependent (the replacement of the initial v3 record is what triggers it — more likely when Rancher is still settling right after install).

**Expected**: either the `imported` annotation should only be applied after the agent manifest has been successfully applied at least once, or the controller should re-verify agent presence when the current v3 record is not being deleted.

**Workaround we use in CI**: [a watcher](https://github.com/rancher-sandbox/cluster-api-provider-harvester/blob/379b34946ba92809b4b83d14a32bd63b6bd81a0f/test/certification/suites/import-gitops/suite_test.go#L168) that strips the premature annotation when the CAPI cluster is not being deleted; the next reconcile then downloads and applies the manifest and the import completes normally.

jniedergang:

`reconcileDelete` annotates the CAPI cluster with `imported=true` whenever the
`management.cattle.io/v3` cluster record is observed with a deletion timestamp,
regardless of whether the import ever completed. If a record is deleted while
the import is still settling and its finalization lasts long enough for a
reconcile to observe it, the annotation permanently disables the import: every
future reconcile is short-circuited, no new record is created and the agent is
never deployed. A deterministic reproduction with controller logs is in #2641.

With this change the annotation is only set when the record being removed had
reached the `AgentDeployed` or `Ready` condition. A record deleted mid-import
now results in a re-import on the next reconcile, while unimporting a cluster
whose import completed keeps preventing re-imports. ADR 0003 is amended
accordingly.

Fixes #2641

- Behavior note: deleting a record whose import never completed used to stop
  the import for good; with this change Turtles recreates the record.
  Preventing the import for such a cluster can still be done by setting the
  `imported` annotation on the CAPI cluster manually.
- No e2e test: reproducing the race deterministically requires holding the
  record in deleting state at the exact reconcile window, which is what the
  added envtest cases do; a timing-based e2e would be flaky by construction.

**Checklist**:
- [x] includes documentation
- [x] adds unit tests

### DIFF
--- a/docs/adr/0003-deletion-strategy.md
+++ b/docs/adr/0003-deletion-strategy.md
@@ -27,6 +27,14 @@ The deletion workflow is structured as follows:
 
 **Cluster Annotation:** When deleting a Rancher cluster, the operator will annotate the corresponding CAPI cluster with the `ClusterImportedAnnotation` (`imported=“true”`) annotation. This annotation will prevent automatic re-import of the CAPI cluster after corresponding Rancher cluster deletion. The underlying infrastructure provisioned by CAPI is left intact.
 
+Amendment (2026-08): the annotation is only applied when the import of the
+deleted Rancher cluster had completed at least once (its record reached the
+`AgentDeployed` or `Ready` condition). A record that disappears before the
+agent was ever deployed, for example when it is replaced while the import is
+still settling, does not mark the CAPI cluster as imported: the next reconcile
+re-creates the record and the import proceeds. Without this restriction such a
+cluster could never be imported again without manual intervention.
+
 From an end-user perspective:
 
 * If user manually deletes CAPI cluster. Rancher turtles uses [Kubernetes owner references](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/) to track relationships between objects. These references are used for Kubernetes garbage collection, which is the basis of Rancher cluster deletion in Rancher turtles.
--- a/internal/controllers/import_controller.go
+++ b/internal/controllers/import_controller.go
@@ -236,7 +236,7 @@ func (r *CAPIImportReconciler) reconcile(ctx context.Context, capiCluster *clust
 		// Patch CAPI Cluster with:
 		// 1. `imported=true` annotation to prevent further-reimports.
 		// 2. Removed capicluster.turtles.cattle.io finalizer to allow deletion.
-		if err := r.reconcileDelete(ctx, capiCluster); err != nil {
+		if err := r.reconcileDelete(ctx, capiCluster, rancherCluster); err != nil {
 			log.Error(err, "Removing CAPI Cluster failed, retrying")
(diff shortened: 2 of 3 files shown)


anmazzotti:
I feel we are trying to implement a solution for a case which is still not well understood.
Technically what is observed is correct already. The management.cattle.io/v3 Cluster removal is how users can unimport Clusters. Adding a dependency on the agent being actually deployed may introduce issues for other users, for example to cleanup failed imports.

I think we still need to determine the cause of management.cattle.io/v3 removal first.

jniedergang:
That is a fair concern, and it is the trade-off I tried to flag in the notes:
deletion-as-cleanup and deletion-as-accident are indistinguishable with the
annotation mechanism, whichever default one picks. Since #2086 deprecates the
annotation in favor of an explicit label, it solves the whole class rather
than picking a default, so I am fine parking or closing this PR in favor of
it. Happy to adjust or close, whichever you prefer.

anmazzotti:
@jniedergang Thank you, I'd rather close this one. #2086 is going to be more complex as there's probably a lot of refactor needed in the import controller. Also regarding priority for #2086 I would recommend the Jira route.

jniedergang:
Understood, closing in favor of #2086. Thanks for taking the time to review.
The envtest cases from this PR may still be useful as regression coverage
whenever the import controller refactor lands; feel free to cherry-pick
anything that helps.

PR #2796, opened 2026-09-03
feat: populate Fleet cluster's `.spec.templateValues`

yiannistri:
**What this PR does / why we need it**:

This PR updates the Fleet controller to make the `Cluster`, `ControlPlane` and `InfrastructureCluster` objects of the CAPI cluster, available under the  Fleet cluster's `spec.templateValues` field.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Part of https://github.com/rancher/rancher/issues/56422

**Special notes for your reviewer**:

**Checklist**:

- [x] squashed commits into logical changes
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

PR #2803, opened 2026-09-04
feat: add Fleet Controller

anmazzotti:

This PR adds an initial implementation of a Fleet controller.

This controller is meant to port CAAPF features and it's the enabler for the related issue: https://github.com/rancher/rancher/issues/56799

It is currently gated under a new alpha feature and disabled by default.

Test coverage is missing e2e tests. These will come with #2796 or immediately after.

Documentation:
- https://github.com/rancher/turtles-product-docs/pull/278

Changes:
- Adds a new alpha feature `fleet-integration`
- Adds a new Fleet Controller in Turtles that reconciles Rancher managed Fleet Clusters (associated by the `management.cattle.io/cluster-name` label).
- Propagate `cluster-api.cattle.io/clusterclass-name` and `cluster-api.cattle.io/clusterclass-namespace` labels to Fleet Cluster, if associated to CAPI Clusters consuming ClusterClasses.
- Create and manage a BundleNamespaceMapping to link the CAPI ClusterClass namespace to the Fleet Cluster namespace.

Fixes:
- Streamline the usage of `cluster-api.cattle.io/owned`, `cluster-api.cattle.io/capi-cluster-owner`, `cluster-api.cattle.io/capi-cluster-owner-ns` labels on `management.cattle.io` Clusters. All three labels are meant to be required for a management Cluster to be considered owned by Turtles. Note that the `cluster-api.cattle.io/owned` is normally left empty and it's probably legacy from the first provisioningv1 import controller. Also note that `rancher/rancher` uses this label alone for some Turtles related day2ops logic.

- Streamline predicates to filter Turtles importable CAPI Clusters so that it can be reused by other controllers.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


yiannistri (approved):
Great work 💯

feature/feature.go
```diff
@@ -56,4 +60,5 @@ var DefaultGates = map[featuregate.Feature]featuregate.FeatureSpec{
 	UIPlugin:                  {Default: false, PreRelease: featuregate.Alpha},
 	UseCAAPF:                  {Default: false, PreRelease: featuregate.Alpha},
 	UseRancherDefaultRegistry: {Default: true, PreRelease: featuregate.Beta},
+	FleetIntegration:          {Default: false, PreRelease: featuregate.Alpha},
```
furkatgofurov7, line 63: Have we not decided to enable it "by default"

internal/controllers/helpers.go
```diff
@@ -466,3 +473,36 @@ func isObjectOwnedBySecret(obj client.Object, sourceSecret *corev1.Secret) bool
...
+			sort.Slice(clusterList.Items, func(i, j int) bool {
+				return clusterList.Items[i].CreationTimestamp.Before(
+					&clusterList.Items[j].CreationTimestamp,
+				)
+			})
+
+			// If we found multiple management.cattle.io Clusters, print an error to highlight the inconsistency to the user.
+			// This is an unexpected state and it should not happen when Clusters are managed by Turtles.
+			// Since there is no way to determine a root cause for this state, the only solution is for the user to investigate
+			// and possibly remove the other Clusters manually.
```
furkatgofurov7, line 493: Can't we make sure to have only Turtles managed clusters in the list by checking it, like in `IsClusterManagedByTurtles`?

util/util.go
```diff
@@ -76,3 +79,55 @@ func ShouldAutoImport(ctx context.Context, logger logr.Logger, cl client.Client,
...
+		turtlesv1.LabelCAPIClusterOwnerName:      capiCluster.Name,
+		turtlesv1.LabelCAPIClusterOwnerNamespace: capiCluster.Namespace,
+		turtlesv1.LabelCAPIClusterOwned:          "",
+	}
+}
+
+// IsFleetClusterManagedByRancher returns true if the fleet.cattle.io Cluster is managed by Rancher.
+// Fleet Clusters created by Turtles will carry the following label:
+//
+//	management.cattle.io/cluster-name: c-bkcp2
```
furkatgofurov7, line 122: this looks like a typo?

furkatgofurov7 (commented):
Nothing blocking, thanks for the PR

furkatgofurov7 (approved):

feature/feature.go
anmazzotti: No, based on your remark it was decided to start with Alpha and disabled first.
May still be the case that it will be promoted to Beta before release.

internal/controllers/helpers.go
anmazzotti: The input in this list should already be filtered. 
Filtering is done controller-runtime client side already and streamlined.

What we do here is to just to sort the list by creation date. Previously we just used the first entry in the list, but the list is not ordered, so this code is here to sanitize this logic that previously existed.

On **how** is it possible to have multiple management.cattle.io Clusters per CAPI Cluster, it's just an API problem since management.cattle.io Clusters are not namespaced and we generate random `c-xyz` names, we may end up in a many to one case.

util/util.go
anmazzotti: I don't see the typo. 
The `management.cattle.io/cluster-name` label is set by Rancher on fleet.cattle.io Clusters that are associated with a certain `management.cattle.io` Cluster.

`c-bkcp2` is just an example name. Turtles does use the `c-`random naming convention.

